PR #1, opened 2021-11-25
Swap 2004

bolmsten:
## Description

Addingg endpoint for generating jwt token

## Motivation

needed for download of files

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?


package.json
```diff
@@ -33,6 +33,7 @@
     "@nestjs/platform-express": "^8.0.0",
     "@nestjs/swagger": "^5.0.9",
     "bcrypt": "^5.0.1",
+    "jsonwebtoken": "^8.5.1",
```
henrikjohansson712: This package is unnecessary, functionality is already part of the @nestjs/jwt package. See auth.service.ts: https://github.com/SciCatProject/scicat-backend-next/blob/0382a0ab14745a9a20154e6b96c2d90b27a64492/src/auth/auth.service.ts#L39

src/config/configuration.example.ts
```diff
@@ -0,0 +1,51 @@
+export default () => ({
```
henrikjohansson712: This file is no longer used, should be deleted.

src/users/users.service.ts
```diff
@@ -13,6 +19,9 @@ import {
 import { User, UserDocument } from "./schemas/user.schema";
 import { CreateRoleDto } from "./dto/create-role.dto";
 import { CreateUserRoleDto } from "./dto/create-user-role.dto";
+import { CreateUserJWT } from "./dto/create-user-jwt.dto";
+import * as jwt from "jsonwebtoken";
```
henrikjohansson712: Replace with:
```
import { JwtService } from "@nestjs/jwt";
```

henrikjohansson712 (changes requested):

src/users/users.service.ts
```diff
@@ -139,4 +148,37 @@ export class UsersService implements OnModuleInit {
...
   }
+
+  async createUserJWT(
+    accessToken: JWTUser | undefined,
+  ): Promise<CreateUserJWT | null> {
+    const signAndVerifyOptions = {
+      expiresIn: this.configService.get<string>("jwt.expiresIn") || "1h",
+      secret: this.configService.get<string>("jwt.secret"),
+    };
+    const jwt = new JwtService(signAndVerifyOptions);
```
henrikjohansson712: JwtService should be added through the constructor, see [auth.service.ts](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/auth/auth.service.ts).

src/users/users.service.ts
```diff
@@ -139,4 +148,37 @@ export class UsersService implements OnModuleInit {
...
+
+    if (!signAndVerifyOptions.secret) {
```
henrikjohansson712: If you add JwtModule to the imports in *users.module.ts*, like in [auth.module.ts](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/auth/auth.module.ts), there's no need for this check, as it's already built into the module.

henrikjohansson712 (changes requested):

henrikjohansson712 (approved):

PR #20, opened 2022-01-13
Feature/enable dataset findone filtering

henrikjohansson712:
## Description

Sets up the ability to include a datasets related models in a query.

## Motivation

Necessary functionality.

## Changes:

* Create Datablock module
* Create OrigDatablockModule
* Add filter logic for Datasets related models
* Add endpoints fort dataset related models

- [ ] Docs updated?


nguyenlinhlinh:
@henrikjohansson712  I am working on a PR which I am trying to separate data fetching of the dataset details. 
The goal is to fetch data on each tab on demand instead of fetching everything at once. I am trying to make it works by using these endpoints: GET /Datasets/{id}/[origdatablocks | attachments | datalifecycle ...] 
Will these endpoints  still be available? or could I somehow easily filter what fields in this findOne should be included in case they are not?

henrikjohansson712:
@nguyenlinhlinh datasetlifecycle is currently embedded in the dataset, so that is not available at the endpoint you mentioned (but maybe we could change that, I don't think that it's used outside of that tab in the frontend), but the others (attachments, datablocks, origdatablocks) will be available once this pr is merged. You will also have the option to just get the Dataset with a GET /Datasets/{id}, or adding an include filter to get the related models.

nguyenlinhlinh:
@henrikjohansson712 I think the datalifecyle could be huge later so I could be better to also separate it. If the endpoints I am using now are still avalaible then it is fine. Regarding the findOne with include filter I planned to use that endpoint at the beginning but it is inconvenient since I have to write a filter like this to exclude the fields I dont want to have in response (I only want the origdatablocks and not the dataset details ). It would be nice if the new backend have a better way to only get fields from related models and filter out field from Dataset model.
```include: [
            {
              relation: "origdatablocks",
              scope: {
                fields: [
                   "id",
                  "size",
                  "ownerGroup",
                  "accessGroup",
                  "createdBy",
                  "datasetId",
                  "dataFileList",
                  "createdAt",
                  "updatedAt"
                ]
                }
              }
          ],
          fields: {
            pid: true,
            datasetId: true,
          },```

PR #30, opened 2022-02-17
fix(uuid): fix new id uniqueness

jekabs-karklins:
## Description

Fix new id uniqueness

## Motivation

Making sure that new UUID is generated for each entry

## Changes:

* darablock.schema.ts
* origdatablock.schema.ts
* sample.schema.ts


henrikjohansson712 (approved):
Great, thanks for fixing this!

PR #119, opened 2022-07-15
 Partial matching of strings in generateScientificExpression

linupi:
**Port of https://github.com/SciCatProject/backend/pull/680 to the new backend!**

## Description
This PR adds an additional `CONTAINS_STRING` condition to the `generateScientificExpression` function to filter within strings in scientific metadata.

## Motivation

When defining our meta data schemas we realized that it would be very helpful for us to also filter on partially matching strings in the scientific metadata. However, so far there is only `EQUAL_TO_STRING` for text based entries in scientific metadata so we propose an additional `CONTAINS_STRING` which should work on simple strings as well as on lists. Here is an example:

```json
"scientificMetadata": {
        "localContact": "John Doe",
        "experimentalists": [
                "John Doe",
                "Max Mustermann"
         ]
```
The idea of the proposed `CONTAINS_STRING` is that it would match e.g. `John` in both cases (as `localContact` as well as in  `experimentalists`)

## Changes:

* additinal condition for `generateScientificExpression` in `mongo-queryable.js`
* the related frontend changes can be found in https://github.com/SciCatProject/frontend/pull/974

- [ ] Docs updated?


nitrosx:
@stephan271 can you please provide an example of a scientificMetadata without value/unit?

stephan271:
image

nitrosx:
At ESS, we always have value and unit.
Regarding this PR, we should expect both cases.
Separately, we should open a new issue to check the scientific metadata that should allow both cases

nguyenlinhlinh:
At MAXIV we also only have unit and value for physical quantities. Others will only have value.

nitrosx:
@linupi are you able to update the code to include the case when the scientific metadata entry is a string and not a object with fields value and unit?

linupi:
> @linupi are you able to update the code to include the case when the scientific metadata entry is a string and not a object with fields value and unit?

I could for sure adopt the code proposed in here, but I am wondering if it is not a more general issue since the existence of a `.value` entry in the scientific metadata is assumed at different places in the code.
e.g.
<pre>src/common/utils.ts<font color="#06989A">:</font>20<font color="#06989A">:</font>    return { valueSI: Number(quantity<font color="#CC0000"><b>.value</b></font>), unitSI: quantity.unit };
src/common/utils.ts<font color="#06989A">:</font>92<font color="#06989A">:</font>    const matchKeyMeasurement = `scientificMetadata.${lhs}<font color="#CC0000"><b>.value</b></font>SI`;
src/common/utils.ts<font color="#06989A">:</font>97<font color="#06989A">:</font>        scientificFilterQuery[`${matchKeyGeneric}<font color="#CC0000"><b>.value</b></font>`] = { $eq: rhs };
src/common/utils.ts<font color="#06989A">:</font>102<font color="#06989A">:</font>        scientificFilterQuery[`${matchKeyGeneric}<font color="#CC0000"><b>.value</b></font>`] = { $eq: escapedRhs };
src/common/utils.ts<font color="#06989A">:</font>111<font color="#06989A">:</font>          scientificFilterQuery[`${matchKeyGeneric}<font color="#CC0000"><b>.value</b></font>`] = { $eq: rhs };
src/common/utils.ts<font color="#06989A">:</font>121<font color="#06989A">:</font>          scientificFilterQuery[`${matchKeyGeneric}<font color="#CC0000"><b>.value</b></font>`] = { $gt: rhs };
src/common/utils.ts<font color="#06989A">:</font>131<font color="#06989A">:</font>          scientificFilterQuery[`${matchKeyGeneric}<font color="#CC0000"><b>.value</b></font>`] = { $lt: rhs };
src/datasets/interceptors/fullquery.interceptor.ts<font color="#06989A">:</font>34<font color="#06989A">:</font>                `${lhs}<font color="#CC0000"><b>.value</b></font>`,
src/datasets/interceptors/fullquery.interceptor.ts<font color="#06989A">:</font>46<font color="#06989A">:</font>                  `${lhs}<font color="#CC0000"><b>.value</b></font>`,
</pre>

wouldn't it make sense to provide a more generic helper to deal with the different options of having `.value`, `.value`+`.unit` or none of them? ... in that case I wouldn't be the best person to implement that I think.

nitrosx:
@linupi should I merge this branch in to master?

linupi:
> @linupi should I merge this branch in to master?

@nitrosx no not yet. First we have to solve the issue discussed above (#138) and after that I will have to adopt this code again. Help with #138 would be much appreciated.

nitrosx:
We should close this PR as I included the changes in #1293.

PR #128, opened 2022-07-26
Remove unique index on dataset.techniques

### LINKED ISSUE #121: Index on datasets collection causes issues
Techniques is a required field on Dataset. The Technique object is defined as: 

``` json
"techniques": [
    {
      "pid": "string",
      "name": "string"
    }
```
There is a unique index on `Datasets.techniques.pid`. Therefore, two datasets cannot share the same technique PID. Am I missing something here?

dylanmcreynolds:
## Description

Fixes #121. There was a PR a few months ago that removed many unique index constraints, replacing them with sparse. This particular index was missed...it was made sparse but unique was not removed. This PR fixes that.

Tests pass.

### DIFF
--- a/src/datasets/schemas/technique.schema.ts
+++ b/src/datasets/schemas/technique.schema.ts
@@ -10,7 +10,7 @@ export class Technique {
     type: String,
     description: "Persistent Identifier dervied from UUIDv4",
   })
-  @Prop({ type: String, unique: true, sparse: true })
+  @Prop({ type: String, sparse: true })
   pid: string;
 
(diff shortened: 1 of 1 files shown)


dylanmcreynolds:
Existing databases will want to manually delete the affected index:

``` javascript
use('dacat_new')

db.Dataset.dropIndex('techniques.pid_1')
```

Then, restart the server. The index will be regenerated, without the uniqueness constraint.

nguyenlinhlinh (approved):

PR #129, opened 2022-07-26
Add migrate-mongo and a migration

dylanmcreynolds:
## Description
This is experimental, would love some opinions before anyone thinks about merging.

Adds a package, migrate-mongo, to be able to manage database schema changes on the new backend.

## Motivation

Database migrations are a fairly standard practice, and there are a lot of tools out there to help manage the process of handling them during upgrades of the backend code. I hit the potential need in a recent PR, where I know that folks will want to drop and index after taking the changes.

I spent a little time playing with different tools. There are a couple of projects that use mongoose directly, but they are effectively dead. `migrate-mongo` appears to be well maintained and seemed to work for my very simple use case.

## How to use
If this gets merged, we'll want to write some official documentation. Here's at least an outline.

### For Scicat Deployers
Occasionally, upgrades of the backend code will require database migration scripts to be run. These scripts are managed by a package called  [migrate-mongo](https://github.com/seppevs/migrate-mongo). After upgrading server code, it is best practice to run the following from the command line where the server is installed.

```
migrate-mongo up
```

Any scripts that are in the `/migrations` directory will be run.

### For SciCat Developers
Did you prepare a PR that requires database schema changes? Create a migration script!

First, to create a script, in the root of the project, run:
```
migrate-mongo create <script name>
```
This will create a stub script in the `/migrations` directory. Code your changes there.

## Changes:

* Added dependency on `migrate-mongo`
* Added a migration script for a small change to Dataset indexes

- [ x] Passing? (Merge will not be approved unless this is checked) 
- [ ] Docs updated?

### DIFF
--- a/migrate-mongo-config.js
+++ b/migrate-mongo-config.js
@@ -1,0 +1,36 @@
+// In this file you can configure migrate-mongo
+
+const config = {
+  mongodb: {
+    // TODO Change (or review) the url to your MongoDB:
+    url: "mongodb://localhost:27017",
+
+    // TODO Change this to your database name:
+    databaseName: "dacat",
+
+    options: {
+      useNewUrlParser: true, // removes a deprecation warning when connecting
+      useUnifiedTopology: true, // removes a deprecating warning when connecting
+      //   connectTimeoutMS: 3600000, // increase connection timeout to 1 hour
+      //   socketTimeoutMS: 3600000, // increase socket timeout to 1 hour
+    }
+  },
+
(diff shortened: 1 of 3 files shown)


nitrosx:
@dylanmcreynolds : I think this is a great idea!!! I have seen in other platforms.

nitrosx:
I agree we need to document really well the migration process in the documentation

nitrosx (approved):

Issue #107, opened 2022-07-05
Support for Oauth2 Authenciation


dylanmcreynolds:
We need  to migrate OAuth2 authentication from the old backed into the new backend.

Some design notes:

We will need:
- A guard for OIDC - gets put in the decorator for the endpoint
- An OAuth strategy (extending the passport openid strategy)- tells you what to do with the result of Oauth, which you use to create the SciCat user profile.
- A mechanism to replace the profile loading callback from loopback3

The last point will require some investigation. How much do we want to support a generalized mechanism that replaces the login callback from the old backend and how much do we just rely on sites to implement their own code within the Auth strategy that they create (above)?

minottic:
these could help:

1. in general: https://github.com/loopbackio/loopback-next/tree/master/examples/passport-login/src 
2. auth via oidc: https://github.com/paulscherrerinstitute/scilog/blob/psdk/sci-log-db/src/authentication-strategies/oidc.ts#L30-L33
3. user creation on the DB: https://github.com/loopbackio/loopback-next/blob/master/examples/passport-login/src/services/user.service.ts#L16-L95 (passed as the verify function to 2)
4.

minottic:
@dylanmcreynolds , since we have some time to focus on this and we're thinking to use it soon, is it fine if I pick up this issue and implement?

dylanmcreynolds:
@minottic sure, thanks! I've focused on the pyscicat stuff, so I would love if you did this.

bolmsten:
Done? @dylanmcreynolds

dylanmcreynolds:
> Done? @dylanmcreynolds

Yes, I have to figure out my strategy with multiple authenticators, but that could potentially be another PR.

PR #144, opened 2022-08-19
OIDC

minottic:
## Description

First version of the OIDC implementation. Current limits: 

1. it supports one single OIDC provider
2. it does not yet support custom callback logic
3. tests are missing

## Motivation
issue: #107 

- [ ] Docs updated?

### DIFF
--- a/src/auth/strategies/oidc.strategy.ts
+++ b/src/auth/strategies/oidc.strategy.ts
@@ -6,31 +6,36 @@ import { CreateUserIdentityDto } from "src/users/dto/create-user-identity.dto";
 import { CreateUserDto } from "src/users/dto/create-user.dto";
 import { User, UserDocument } from "src/users/schemas/user.schema";
 import { UsersService } from "src/users/users.service";
-import { Strategy, Client, UserinfoResponse, TokenSet, Issuer } from 'openid-client';
+import {
+  Strategy,
+  Client,
+  UserinfoResponse,
+  TokenSet,
+  Issuer,
+} from "openid-client";
 import { AuthService } from "../auth.service";
 import { Profile } from "passport";
 import { UserProfile } from "src/users/schemas/user-profile.schema";
 
-
-export class buildOpenIdClient {
-  constructor(
-    private configService: ConfigService
-  ) { }
+export class BuildOpenIdClient {
+  constructor(private configService: ConfigService) {}
   async build() {
-    const oidcConfig = this.configService.get<Record<string, unknown>>("oidc")
-    const TrustIssuer = await Issuer.discover(`${oidcConfig?.issuer}/.well-known/openid-configuration`);
-    const client = new TrustIssuer.Client({
+    const oidcConfig = this.configService.get<Record<string, unknown>>("oidc");
+    const trustIssuer = await Issuer.discover(
+      `${oidcConfig?.issuer}/.well-known/openid-configuration`,
+    );
+    const client = new trustIssuer.Client({
       client_id: oidcConfig?.clientID as string,
-      client_secret: oidcConfig?.clientSecret as string
+      client_secret: oidcConfig?.clientSecret as string,
     });
     return client;
   }
-};
+}
 
 @Injectable()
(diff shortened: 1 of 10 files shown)


minottic:
@dylanmcreynolds , @nitrosx I've pushed what I've done before I leave for holiday, even though it still has quite some limitations. I think this first version could in principle be merged after adding the tests and when happy. 

The improvements (1,2) IMHO can be added in later PRs

src/auth/strategies/oidc.strategy.ts
```diff
@@ -0,0 +1,141 @@
...
+  Client,
+  UserinfoResponse,
+  TokenSet,
+  Issuer,
+} from "openid-client";
+import { AuthService } from "../auth.service";
+import { Profile } from "passport";
+import { UserProfile } from "src/users/schemas/user-profile.schema";
+
+export class buildOpenIdClient {
```
dylanmcreynolds, line 20: Super picky point, classes should be UpperCamelCase

src/auth/strategies/oidc.strategy.ts
```diff
@@ -0,0 +1,141 @@
...
+  constructor(private configService: ConfigService) {}
+  async build() {
+    const oidcConfig = this.configService.get<Record<string, unknown>>("oidc");
+    const TrustIssuer = await Issuer.discover(
```
dylanmcreynolds, line 24: lowerCamelCase

src/auth/strategies/oidc.strategy.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (userinfo.preferred_username) {
+      profile.username = userinfo.preferred_username;
+    }
+    if (userinfo.email) {
+      profile.emails = [{ value: userinfo.email }];
+      profile.email = userinfo.email;
+    }
+
+    profile.accessGroups = userinfo[
+      this.configService.get("oidc").accessGroups
```
dylanmcreynolds, line 136: Does this mean that access groups are always hard-coded in configuration?

dylanmcreynolds:
I may be missing somethbing, but It seems to me that to be useful, we also need some mechanism of asking an external system for a user's group memberships (as the old backend got with the login-callback.js). In your case, are you getting that right out of keycloak in the profile? In this PR, it looks to me that it's hard-coded from the configuration. So, again, if I'm reading this right, the PR solves authentication with OIDC but not authorization, no?

src/auth/strategies/oidc.strategy.ts
minottic: only the key to retrieve the accessGroups values from the userInfo object (which is returned by the oidc when calling the userInfo endpoint):
Schematically, these are the steps: 

1. the code gets the oidc `.well-know` config
2. authentication is triggered through the oidc provider
3. when successful, oidc returns a token (jwt) 
4. token from 3 is used to call the oidc userInfo endpoint, which is contained in 1
5. 4 returns general user information, e.g. name, username etc and here this PR makes the assumption that one property of the userInfo is the accessGroups
6. line 136 extracts the accessGroups having specified the name of the userinfo accessGroups property in the config (e.g. at PSI we call them p-groups)

minottic:
to some extent, yes, and that is what I tried to explain in comments (1, 2) opening the PR. 

For this PR to be useful, there is the assumption that the accessGroups are in the userInfo which are returned after the oidc authentication (see comment above).

I agree that this PR does not address the case where accessGroups come from different sources neither it covers cases where there are multiple oidc providers, but I think, after having resolved your other comments (which I'll do asap), that these 2 limitations can be tackled in later PRs

dylanmcreynolds (approved):
For some sites, a follow up PR will be required to obtain user group from an external source. But this will work for sites for whom that information is coming from the OIDC provider. Great stuff. Thanks!

minottic:
This is still open as I was hoping to add some acceptance tests, but I have struggled to find time recently. My idea was to add to the test docker-compose an instance of keycloak with some predefined accounts that can be used to test

PR #153, opened 2022-09-26
Add docker-compose file to ease local deployment

minottic:
## Description

This PR adds a docker-compose.dev.yaml file which spins up a dev environment 

## Motivation

Eases local deployment 

## Changes:

* docker-compose.dev.yaml added
* Dockerfile split in 3 parts, vs 2 before, to have one suited for local development

- [x] Docs updated?

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,16 @@
 6. Go to http://localhost:3000/explorer-next to get an overview of available endpoints and database schemas.
 7. To be able to run the e2e tests with the same setup as in the Github actions you will need to run `npm run  prepare:local` and after that run `npm run start:dev`. This will start all needed containers and copy some configuration to the right place.
 
+## Develop in a container using the docker-compose.dev file
+
+1. `git clone https://github.com/SciCatProject/scicat-backend-next.git`
+2. docker-compose -f docker-compose.dev.yaml up -d
+3. *Optional* Mount *functionalAccounts.json* file to a volume in the container. If not set up, the functional accounts in [functionalAccounts.json.example](/functionalAccounts.json.example) will be created automatically.
+4. *Optional* change the container env variables
+5. Attach to the container
+6. `npm run start:dev`
+7. Go to http://localhost:3000/explorer-next to get an overview of available endpoints and database schemas.
+
 ## Test the app
 
 1. **Running the unit tests:** `npm run test`
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -1,0 +1,34 @@
+services:
+  mongodb:
+    image: bitnami/mongodb:4.2
+    volumes:
+      - mongodb_data:/bitnami
+  backend:
+    build: 
+      context: .
+      target: dev
+    depends_on:
+      - mongodb
+    ports:
+      - 3000:3000
+    volumes:
+      - .:/home/node/app
+      - /home/node/app/node_modules
+    environment:
+      MONGODB_URI: mongodb://mongodb:27017/scicat
+      EXPRESS_SESSION_SECRET: a_scicat_secret
+      JWT_SECRET: a_scicat_secret
+      PORT: 3000
+      HTTP_MAX_REDIRECTS: 5
+      HTTP_TIMEOUT: 5000
+      JWT_EXPIRES_IN: 3600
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -18,6 +18,16 @@
...
 7. To be able to run the e2e tests with the same setup as in the Github actions you will need to run `npm run  prepare:local` and after that run `npm run start:dev`. This will start all needed containers and copy some configuration to the right place.
 
+## Develop in a container using the docker-compose.dev file
+
+1. `git clone https://github.com/SciCatProject/scicat-backend-next.git`
+2. docker-compose -f docker-compose.dev.yaml up -d
+3. *Optional* Mount *functionalAccounts.json* file to a volume in the container. If not set up, the functional accounts in [functionalAccounts.json.example](/functionalAccounts.json.example) will be created automatically.
+4. *Optional* change the container env variables
+5. Attach to the container
+6. `npm run start:dev`
```
LAShemilt, line 28: One thing that I don't like about the process is having to run the "npm run start:dev" after the docker containers have been set up. This manual step won't deploy nicely to production. I have tried putting this command in the docker file and in the docker compose file under the command part. However neither work. It would be nice to have a set up that we can spin up in docker-compose without any other intervention.

README.md
```diff
@@ -18,6 +18,16 @@
 6. Go to http://localhost:3000/explorer-next to get an overview of available endpoints and database schemas.
 7. To be able to run the e2e tests with the same setup as in the Github actions you will need to run `npm run  prepare:local` and after that run `npm run start:dev`. This will start all needed containers and copy some configuration to the right place.
 
+## Develop in a container using the docker-compose.dev file
+
+1. `git clone https://github.com/SciCatProject/scicat-backend-next.git`
+2. docker-compose -f docker-compose.dev.yaml up -d
+3. *Optional* Mount *functionalAccounts.json* file to a volume in the container. If not set up, the functional accounts in [functionalAccounts.json.example](/functionalAccounts.json.example) will be created automatically.
```
LAShemilt, line 25: This is a very nice feature for new users to scicat, thanks for including it.

docker-compose.dev.yaml
```diff
@@ -0,0 +1,34 @@
...
+      PORT: 3000
+      HTTP_MAX_REDIRECTS: 5
+      HTTP_TIMEOUT: 5000
+      JWT_EXPIRES_IN: 3600
+      SITE: SAMPLE-SITE
+      PID_PREFIX: PID.SAMPLE.PREFIX
+      DOI_PREFIX: DOI.SAMPLE.PREFIX
+      METADATA_KEYS_RETURN_LIMIT: 100
+      METADATA_PARENT_INSTANCES_RETURN_LIMIT: 100
+    command: /bin/sh -c "while true; do sleep 600; done"
```
LAShemilt, line 30: why is it necessary to run this sleep command?

LAShemilt (commented):
Thanks for writing this it is a very useful starting place for users wanting to set up for the first time.

README.md
LAShemilt: So in my above comment I have got it running with the backend from `main` which is a different dockerfile. I can't get it to work with the docker file in this PR.

docker-compose.dev.yaml
minottic: the idea of this docker-compose is to have an env which you can attach to do your local dev (that's also why the code is mounted). So the sleep is to keep the container busy so that you can attach to it

README.md
minottic: ok, that could be. I have the feeling the breaking bit is [this](https://github.com/SciCatProject/scicat-backend-next/pull/153/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R1). I had to add it because on MAC that is needed, but it could be that it complains on other architectures (I couldn't try it), and I forgot to add such a comment. Could you please try without it and let me know if it works? or maybe you could post here the error logs. Thanks!

nitrosx:
@LAShemilt @minottic Should I merge this PR in to master?

nitrosx:
Never mind, I just read the other PR. I will merge it and than fix it from master if needed

PR #209, opened 2022-12-01
Proposal model and documentation

nitrosx:
## Description

This PR add better documentation for the Proposal model and endpoint. It includes improvements to the swagger explorer UI for this endpoint.
A complete review of the model has been also done, together with renaming the class model to ProposalClass

## Motivation

Improving documentation, naming, and swagger integration.
Review model, eliminated inconsistencies, and ported field constrains from old backend

## Changes:

* Proposal model class and dto
* Swagger decorators for proposals mode, dto and controller

- [ ] Docs updated?


src/common/interfaces/common.interface.ts
```diff
@@ -40,6 +41,12 @@ export interface ILimitsFilter {
   order: string;
 }
 
+export class ILimitsFilterDto implements ILimitsFilter {
```
martin-trajanovski, line 44: As I see this ILimitsFilterDto is completly the same as ILimitsFilter. Do we really need two different representations or we can have just one?

src/common/interfaces/common.interface.ts
```diff
@@ -1,3 +1,4 @@
+//import { ApiProperty } from "@nestjs/swagger";
```
martin-trajanovski, line 1: This comment can be removed.

src/common/utils.ts
```diff
@@ -623,3 +623,40 @@ export const addUpdatedFields = <T>(
...
+export const fullQueryDescription = '<pre>\n \
+{\n \
+  "fields?": [ "field1", "field2", ...],\n \
+  "limits?": {\n \
+    "limit": number,\n \
+    "skip": number,\n \
+    "order": [ascending, descending]\n \
+  }\n \
+}\n \
+</pre>';;
```
martin-trajanovski, line 662: All these descriptions are a bit confusing but I understand the point it is good to have better documentations for the filters and queries.

src/proposals/proposals.controller.ts
```diff
@@ -97,26 +114,51 @@ export class ProposalsController {
...
   @Get()
+  //@ApiExtraModels(IProposalFilterDto)
   @ApiQuery({
     name: "filters",
-    description: "Database filters to apply when retrieve all proposals",
+    description: "Database filters to apply when retrieve proposals\n" + filterDescription,
     required: false,
+    //schema: {
+    //  $ref: getSchemaPath(IProposalFilterDto),
+    //}
```
martin-trajanovski, line 126: Comments can be removed here

src/proposals/proposals.controller.ts
```diff
@@ -215,49 +344,86 @@ export class ProposalsController {
...
     return this.attachmentsService.findOneAndRemove({
       _id: attachmentId,
       proposalId: proposalId,
     });
   }
 
-  // POST /proposals/:id/datasets
-  @UseGuards(PoliciesGuard)
+  // POST /proposals/:pid/datasets
+/*   @UseGuards(PoliciesGuard)
```
martin-trajanovski, line 378: I see you have commented this part and if it is not needed anymore maybe removing it is better.

src/proposals/proposals.controller.ts
```diff
@@ -215,49 +344,86 @@ export class ProposalsController {
...
+  })
   async findAllDatasets(
-    @Param("id") proposalId: string,
+    @Param("pid") proposalId: string,
   ): Promise<DatasetClass[] | null> {
     return this.datasetsService.findAll({ where: { proposalId } });
   }
 
-  // PATCH /proposals/:id/datasets/:fk
+/*   // PATCH /proposals/:id/datasets/:fk
```
martin-trajanovski, line 426: I see this part is also commented. Are we using it or it should be removed.

martin-trajanovski (changes requested):

src/common/interfaces/common.interface.ts
nitrosx: You are correct!!!
I'm going to remove it.

src/common/utils.ts
nitrosx: I agree with you!!!
I want to get these info in, although I'm absolutely convinced that this is not the best and most efficient way.

src/proposals/proposals.controller.ts
nitrosx: Done. We want to provide only the functionality to return the linked datasets, but not to create/update datasets from proposals

src/proposals/proposals.controller.ts
nitrosx: Done!!!

PR #246, opened 2022-12-15
[WIP] Fix datasets filters and use mongosh for migrations

martin-trajanovski:
## Description

Short description of the pull request

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?

### DIFF
--- a/scripts/migrateIdFields.sh
+++ b/scripts/migrateIdFields.sh
@@ -9,5 +9,5 @@ echo
 read -p "Are you sure you want to run the script? [y/N]" -n 1 -r
 echo
 if [[ $REPLY =~ ^[Yy]$ ]]; then
-    mongo -u "dacatDBAdmin" -p "...."  --authenticationDatabase "dacat" HOSTNAME/dacat  migrateIdFields.js
+    mongosh -u "dacatDBAdmin" -p "...."  --authenticationDatabase "dacat" HOSTNAME/dacat  migrateIdFields.js
 fi
--- a/scripts/replaceObjectIds.sh
+++ b/scripts/replaceObjectIds.sh
@@ -13,5 +13,5 @@ read -p "Are you sure you want to run the script ? This action can not be undone
 echo    
 if [[ $REPLY =~ ^[Yy]$ ]]
 then
-    mongo -u "dacatDBAdmin" -p "...."  --authenticationDatabase "dacat" HOSTNAME/dacat  replaceObjectIds.js
+    mongosh -u "dacatDBAdmin" -p "...."  --authenticationDatabase "dacat" HOSTNAME/dacat  replaceObjectIds.js
 fi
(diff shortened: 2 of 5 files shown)


scripts/migrateIdFields.sh
```diff
@@ -1,13 +1,13 @@
...
+# This command migrates the id fields of the relevant docuemtns in the database
+# It uses the mongo client for this, so you must adjust the host and password parameters
+echo "You need to adjust the script to contain the proper username/password information"
+echo
+echo "You should make a backup of your MongoDB before running this command!"
+echo
+read -p "Are you sure you want to run the script? [y/N]" -n 1 -r
+echo
+if [[ $REPLY =~ ^[Yy]$ ]]; then
+    mongosh -u "dacatDBAdmin" -p "...."  --authenticationDatabase "dacat" HOSTNAME/dacat  migrateIdFields.js
```
nitrosx, line 12: Can we make the database name configurable?
Right now is hard coded to "dacat" but we do not know if the name is consistent across instances. Also "dacat" is not very meaningful.

scripts/replaceObjectIds.sh
```diff
@@ -13,5 +13,5 @@ read -p "Are you sure you want to run the script ? This action can not be undone
 echo    
 if [[ $REPLY =~ ^[Yy]$ ]]
 then
-    mongo -u "dacatDBAdmin" -p "...."  --authenticationDatabase "dacat" HOSTNAME/dacat  replaceObjectIds.js
+    mongosh -u "dacatDBAdmin" -p "...."  --authenticationDatabase "dacat" HOSTNAME/dacat  replaceObjectIds.js
```
nitrosx, line 16: Same as previous comment?

nitrosx (commented):
Please check my comments on migrateIdFields and ReplaceObjectIds scripts and let me know

scripts/migrateIdFields.sh
martin-trajanovski: I agree with you. Would be nicer to have it as an output parameter to the script. I think for now we can leave it as it is (it was hardcoded even before) cause I have some notes about the migration and what needs to be done initially. If you read the instructions on the top of this script I don't find it that useful because it says that you need to adjust the db username and password inside it anyway. So if you do that you can specify the correct database name as well. Actually I wasn't able to run this script inside the container because of some wrong path issues and what I did is just running the command `mongosh -u "dacatDBAdmin" -p "...."  --authenticationDatabase "dacat" HOSTNAME/dacat  migrateIdFields.js` directly after copying the `migrateIdFields.js` inside the container. So this way you can specify whatever database name you want.

nitrosx: OK. Let's merge and we open a ticket later

nitrosx (approved):

PR #247, opened 2022-12-15
Swap 2846 scicat fix user groups OIDC 

jekabs-karklins:
## Motivation

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?


minottic:
this looks cool thanks! 

following up on my Tuesday comment, we currently do something similar for a project called scilog.
The idea is to embed the accessGroups logic in a function which is imported from a dedicated file and that can be overridden with a bind mounting after the container build (very similar to what we do in the old backend). 
Even if we rely on TS, the file to mount after compilation is a trivial JS file, with no special syntax coming from the TS->JS compilation. 

This said I like this approach too, so up to you what to do with this info! The advantage of it would be more flexible to new changes, while the disadvantage is to test the mounted code.

You can find a usage example here: 

https://github.com/paulscherrerinstitute/scilog/blob/main/docker-compose.dev.yaml#L20 

which, when uncommented, mounts this file after build:
https://github.com/paulscherrerinstitute/scilog/blob/main/config/be/roles.js

If unclear, I will be happy to follow up!

dylanmcreynolds:
Thanks for the PR!

Am I reading this code correctly?

With this PR, accessGroups can be configured to come from several places:

1. a static list from somewhere (AccessGroupFromStaticValuesService)
2. from the payload of the IdP (AccessGroupFromStaticValuesService). Does the come out of the id_token that comes back from the IdP?
3. multiple configured providers (AccessGroupFromMultipleProvidersService). This code looks like you could have duplicates in the accessGroups array, no?
4. A GraphQL service. (AccessGroupFromGraphQLApiService). This query is defined in `access-group-service-factory.ts`

This is an interesting approach, but is written to be specific to ESS. (The methods in `access-group-service-factory.ts`) even have `Ess` in them, and the graphql service uses an ESS-specific query.

When I wrote the oidc code in the old backend, I anticipated that different faculties would have wildly different ways of getting accessGroups, assuming there was a common signature to call, and letting each facility provide their own implementation if desired.

@minottic , regarding your suggestion about overriding *.js files (similar to how we did it in the old backend) with this PR, could I not write an access-group-service-factory.js and my own AccessGroupService.js and map them in the right place?

minottic:
> @minottic , regarding your suggestion about overriding *.js files (similar to how we did it in the old backend) with this PR, could I not write an access-group-service-factory.js and my own AccessGroupService.js and map them in the right place?

Indeed! as long as we make the mounting approach work

PR #275, opened 2023-01-03
Swap 2846 generalization

nitrosx:
## Description

We want to generalize the solution, providing a default simple solution that returns the groups listed in configuration.
Together with the default, we want to provide a set of examples highlighting how to customize the access group file to suit the individual facility needs

## Motivation

The solution provided in the original PR is specific for ESS. To facilitate adoption and customization, the default behaviour will return the list of access groups listed in configuration, although the code will include a set of examples on how such file has been customized at ESS and other facilities.
The customized solution must be mounted on the default release image. 

## Fixes:
* src/auth/access-group-provider/access-group-service-factory.ts_ESS_implementation

## Changes:
*  src/auth/access-group-provider/access-group-service-factory.ts

- [ ] Docs updated?


src/auth/access-group-provider/access-group-service-factory.ts
```diff
@@ -1,72 +1,15 @@
...
 import { AccessGroupFromStaticValuesService } from "./access-group-from-static-values.service";
 import { AccessGroupService } from "./access-group.service";
 
+/*
+ * this is the default function which provides an empty array as groups
+ */
 export const accessGroupServiceFactory = {
   provide: AccessGroupService,
   useFactory: (configService: ConfigService) => {
     const site = configService.get<string>("site");
```
jekabs-karklins, line 11: This line can probably be removed

jekabs-karklins (approved):

PR #285, opened 2023-01-15
Swap 2957 2958

nitrosx:
## Description

Testing user authentication for all type of users

## Motivation

SciCat supports three types of accounts: local, ldap and oidc. We are currently testing that such integration works correctly within the ESS infrastructure. Local account do not need any external service.

## Fixes:

* Updates user profile correctly at each login, including access groups

## Changes:

* removed provider from user name
* added provider to user info
* added user profile for local user

- [x ] Included for each change/fix?
- [x] Docs updated?


src/auth/access-group-provider/access-group-from-graphql-api-call.service.ts
```diff
@@ -0,0 +1,46 @@
+import { AccessGroupService as AccessGroupService } from "./access-group.service";
+import { Injectable } from "@nestjs/common";
+import fetch from "node-fetch";
```
martin-trajanovski, line 3: @nitrosx I think instead of introducing new library like `node-fetch` we can use `@nestjs/axios`. Here is an example: https://docs.nestjs.com/techniques/http-module

nitrosx: @Martin: I agree with you.
Let's leave it like that and I will create a new ticket.

martin-trajanovski (approved):
Looks good to me

PR #293, opened 2023-01-19
Swap 2975 add headers filtering on datasets

martin-trajanovski:
## Description

This PR aims to support dataset filtering by headers as well. Previously we only supported filters by query params but some facilities need to use headers cause frontend can also send headers if it is configured like this.


src/datasets/datasets.controller.ts
```diff
@@ -94,6 +95,40 @@ export class DatasetsController {
     private origDatablocksService: OrigDatablocksService,
   ) {}
 
+  getFilters(
+    headers: Record<string, string>,
+    queryFilter: { filter?: string },
+  ) {
+    // NOTE: If both headers and query filters are present return error because we don't want to support this scenario.
+    if (queryFilter?.filter && (headers?.filter || headers?.where)) {
```
nitrosx, line 103: Why are we checking for the where clause only in the header?

nitrosx (approved):
The PR looks good, we might need to brainstorm and provide a better documentation on how the filtering works which will clarify the code dealing with filters.

PR #306, opened 2023-01-25
feat: add all the migration scripts in one place to be able to run with migrate-mongo

martin-trajanovski:
## Description

Adding all database migration scripts in one place.

## Motivation

Previously we had shell scripts and migrate-mongo scripts for initial database migration and needed two different ways of running them.

## Fixes:

* https://jira.esss.lu.se/browse/SWAP-2982

### DIFF
--- a/migrations/20220726224709-technique_index.js
+++ b/migrations/20220726224709-technique_index.js
@@ -1,13 +1,23 @@
 module.exports = {
   async up(db, client) {
-    await db.collection("Dataset").dropIndex("techniques.pid_1");
+    try {
+      await db.collection("Dataset").dropIndex("techniques.pid_1");
+    } catch (error) {
+      console.info(`Index techniques.pid_1 not found`, error);
+    }
     await db
       .collection("Dataset")
       .createIndex({ "techniques.pid": 1 }, { sparse: true });
   },
 
   async down(db, client) {
-    await db.collection("Dataset").dropIndex("techniques.pid_1");
+    try {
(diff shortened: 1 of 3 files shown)


migrations/20220726224709-technique_index.js
```diff
@@ -1,13 +1,23 @@
...
 
   async down(db, client) {
-    await db.collection("Dataset").dropIndex("techniques.pid_1");
+    try {
+      await db.collection("Dataset").dropIndex("techniques.pid_1");
+    } catch (error) {
+      console.info(`Index techniques.pid_1 not found`, error);
+    }
+    // NOTE: This unique flag here is problematic because after importing production data and trying it throws an error:
+    // E11000 duplicate key error collection: dacat.Dataset index: techniques.pid_1 dup key
```
nitrosx, line 20: If this down grade is intended just for db that have just been upgraded, the error should disappear and not be a problem.
If we intend to have a downgrade path for db that have been already used with the new backend, we should drop the unique feature

nitrosx (changes requested):

nitrosx:
@martin-trajanovski : I think we can change this PR to "ready"

migrations/20220726224709-technique_index.js
nitrosx: Following the discussion from the last Collaborators meeting, this PR meets the acceptance criteria.
At the moment, we do not want to address the issue of downgrading a database used with the new backend.

nitrosx (approved):

PR #307, opened 2023-01-26
Added logout function

nitrosx:
## Description

Added logout functionality in auth module

## Motivation

A logout functionality is needed to complete the migration to the new version of SciCat backend. 
auth/logout function invalidates the user session in nest app and also will call the oidc service endpoint to invalidate the oidc session if the account has been authenticated through oidc and the oidc service has a logout functionality.

## Fixes:
* n/a

## Changes:
* added logout endpoint in auth controller
* added auth tests

- [ ] Docs updated?


martin-trajanovski (approved):
This looks great!

bolmsten (approved):

src/auth/auth.controller.ts
```diff
@@ -84,8 +85,44 @@ export class AuthController {
...
+        // we should provide a message
+        res.status(HttpStatus.BAD_REQUEST);
+      }
+      if (user.authStrategy == 'oidc') {
+        const oidcConfig = this.configService.get<OidcConfig>("oidc");
+        req.session.destroy(async (error: any) => {
+          const trustIssuer = await Issuer.discover(
+            `${oidcConfig?.issuer}/.well-known/openid-configuration`,
+          );
+          const end_session_endpoint = trustIssuer.metadata.end_session_endpoint;
```
dylanmcreynolds, line 113: Is this intended to log the user out of the OIDC provider? If so, I don't think we want to do this. I think it's sufficient to log the user out of SciCat and remain logged into the OIDC provider. Thinking of applications that use google, microsoft or ORCID as an OIDC authenticator, do we see any examples that log out out of, say, google directly from the application?

bolmsten: Hi Dylan,

the JWT expires after an hour, if we need the logout to invalidate the JWT immediately we would need to have a database storing tokens with a state. For us this is not necessary but if this is a requirement  I would suggest moving over to ordinary sessions.

/Fredrik

bolmsten: @dylanmcreynolds

dylanmcreynolds: @bolmsten Do I understand correctly?  At the point that the user clicks the logout button it taken here, the session contains a SciCat-generated JWT.  You're saying that we can't invalidate the SciJWT without maintaining it in a database (makes sense), so this code both deletes the session (which it would to on line 122 in in the non-OIDC case) and l redirects the user to their IdP to log them out? In both cases we delete the SciJWT from the session, so I still don't see why the redirection to logout of the IdP is necessary. In my current understanding, if was using, say, google as my OIDC IdP, then logging out of SciCat would log me out of gmail too, because of this redirect on line 155.

Apologies if I'm missing something basic here.

bolmsten: Yes, that is correct. 

We can make the logout from the Idp optional as we would want users to be logged out from our idp but I see you use case. We could however not invalidate the scicat-jwt token without the need for a database table and all that comes with it.

dylanmcreynolds: I think would be happy with the option to not log out of the IdP. However, I would not want to hold up this PR for it.

PR #336, opened 2023-02-13
add prestart script that runs the migrations before every start

martin-trajanovski:
## Description

This PR aims to make database migrations run before every start of the application if you use the `npm run start` script.


nitrosx (commented):
Will this run automatically the upgrade?

nitrosx (approved):

PR #391, opened 2023-03-17
Added user endpoint to return if user is authorized to create datasets

nitrosx:
## Description
This PR adds a user endpoint to check if the user specified is allowed to create datasets or not.
Users with admin privileges are allowed to check different users.

## Motivation
The frontend currently show the "Create Dataset" button in the dataset list page, no matter if the user logged in has the authorization to create the datasets or not.
We want the frontend to show/hide the button according to the user authorization. This endpoint should provide the info needed.

## Changes:
* user controller
* user service

- [ ] Docs updated?

### DIFF
--- a/src/casl/guards/policies.guard.ts
+++ b/src/casl/guards/policies.guard.ts
@@ -18,7 +18,8 @@ export class PoliciesGuard implements CanActivate {
         context.getHandler(),
       ) || [];
 
-    const { user } = context.switchToHttp().getRequest();
+    const req = context.switchToHttp().getRequest();
+    const user = req.user;
     const ability = this.caslAbilityFactory.createForUser(user);
 
     return policyHandlers.every((handler) =>
--- a/src/users/users.controller.ts
+++ b/src/users/users.controller.ts
@@ -9,10 +9,12 @@ import {
   Delete,
   UseInterceptors,
   Put,
+  ForbiddenException,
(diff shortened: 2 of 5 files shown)


src/casl/guards/policies.guard.ts
```diff
@@ -18,7 +18,8 @@ export class PoliciesGuard implements CanActivate {
         context.getHandler(),
       ) || [];
 
-    const { user } = context.switchToHttp().getRequest();
+    const req = context.switchToHttp().getRequest();
```
martin-trajanovski, line 21: This change might not be needed because it was one-liner previously and now it's two for exactly same thing.

src/users/users.controller.ts
```diff
@@ -133,4 +136,49 @@ export class UsersController {
...
+    viewedUserSchema.realm = viewedUser.realm;
+    viewedUserSchema.username = viewedUser.username;
+    viewedUserSchema.email = viewedUser.email;
+
+    if (!ability.can(Action.Read, viewedUserSchema)) {
+      throw new UnauthorizedException("Unauthorized access");
+    }
+  }
+
+  //@UseGuards(PoliciesGuard)
```
martin-trajanovski, line 162: Comments not needed here.

martin-trajanovski (commented):
Small comments but otherwise I find it good

PR #468, opened 2023-04-25
fix: fixed sharing dataset 500 error

Junjiequan:
## Description

Fixed some problematic types for /:pid/appendToArrayField endpoint.

## Motivation

Share datasets to others causing 500 error

## Fixes:

https://user-images.githubusercontent.com/78078898/234321107-be5a732d-7f8c-40cc-957c-039b128a25fc.mp4

- [ ] Docs updated?


src/datasets/datasets.controller.ts
```diff
@@ -924,12 +924,13 @@ export class DatasetsController {
...
     type: String,
   })
-  @ApiParam({
+  @ApiQuery({
     name: "fieldName",
     description: "Name of the field to be updated",
     type: String,
   })
-  @ApiBody({
+  @ApiQuery({
```
nitrosx, line 932: Why have we changed from body to query param?
Are we sure that this endpoint is not used anywhere else?

nitrosx (commented):
I'm afraid that the changed from body to query param is a drastic change that have unforeseen consequences.

src/datasets/datasets.controller.ts
Junjiequan:
```
  public appendToArrayField(id: any, fieldName: any, data: any, customHeaders?: Function): Observable<any> {
    let _method: string = "POST";
    let _url: string = LoopBackConfig.getPath() + "/" + LoopBackConfig.getApiVersion() +
    "/Datasets/:id/appendToArrayField";
    let _routeParams: any = {
      id: id
    };
    let _postBody: any = {};
    let _urlParams: any = {};
    if (typeof fieldName !== 'undefined' && fieldName !== null) _urlParams.fieldName = fieldName;
    if (typeof data !== 'undefined' && data !== null) _urlParams.data = data;
    let result = this.request(_method, _url, _routeParams, _urlParams, _postBody, null, customHeaders);
    return result;
  }
```
I don't see this endpoint is used somewhere else yet
But even this endpoint is used somewhere else the body will always be empty object. You can see the frontend code: `let _postBody: any = {};`, its sending empty {} to backend by default and is not attached to any values.

nitrosx (approved):

PR #481, opened 2023-05-02
Swap 3245 2 - still hunting missing groups

nitrosx:
## Description
Still hunting why access groups static values are not read from .env

## Motivation
We are not getting the full group lists


CI/E2E/.env.backend
```diff
@@ -0,0 +1,33 @@
+DOI_PREFIX="10.17199/"
```
martin-trajanovski, line 1: Are these really needed because how it worked until now?

CI/E2E/config-frontend.backend-e2e.json
```diff
@@ -0,0 +1,122 @@
+{
```
martin-trajanovski, line 1: Also not sure about this file as well if it is needed

martin-trajanovski (changes requested):

PR #485, opened 2023-05-08
feat: enable dataset access for shared emails

martin-trajanovski:
## Description

This PR aims to make shared datasets available for users. Also prevents all users for being able to update datasets and leave that ability only for owners.

## Motivation

Anyone that has access to dataset could update or share it and people in sharedWith field were not able to access the datasets.

## Fixes:

* https://jira.esss.lu.se/browse/SWAP-3274

## Changes:

* changes made

- [ ] Docs updated?

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -948,18 +948,28 @@ export class DatasetsController {
     @Query("data") data: string,
   ): Promise<DatasetClass | null> {
     const loggedInUser: JWTUser = request.user as JWTUser;
-
+    const ability = this.caslAbilityFactory.createForUser(loggedInUser);
     const datasetToUpdate = await this.datasetsService.findOne({
       where: { pid: id },
     });
 
-    // NOTE: Not sure if email is the best way to do this but since we don't have userId or something like that email is acceptable.
-    const couldUpdateDataset =
-      datasetToUpdate?.ownerEmail === loggedInUser.email;
-    // $addToSet is necessary to append to the field and not overwrite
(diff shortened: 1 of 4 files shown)


src/datasets/datasets.controller.ts
```diff
@@ -941,13 +942,27 @@ export class DatasetsController {
...
+    // NOTE: Not sure if email is the best way to do this but since we don't have userId or something like that email is acceptable.
+    const couldUpdateDataset =
+      datasetToUpdate?.ownerEmail === loggedInUser.email;
     // $addToSet is necessary to append to the field and not overwrite
     // $each is necessary as data is an array of values
 
+    if (!couldUpdateDataset) {
+      throw new ForbiddenException();
+    }
+
```
nitrosx, line 965: In order to allow the update on the dataset, we should use the casl factory.
The rules is the following:
- if the user logged in belongs to the group that owns the dataset, he/she is allowed to update the dataset.
- if the user belongs to any of the admin groups, he/she can update the dataset.

nitrosx (changes requested):
Please use the casl facility to check if the user can update the dataset.

src/datasets/datasets.controller.ts
martin-trajanovski: Ahaaa... I see now but do we have these rules set up already or I need to fix that?

martin-trajanovski: @nitrosx I fixed it and now it uses casl to decide if user is able to update something or not.

nitrosx (approved):

PR #493, opened 2023-05-12
fix: logout success with error message

Junjiequan:
## Description

fixed local logout error:

- 404 err: backend is expecting GET but FE is sending POST
- 401 err: logout function on BE expects auth token

fixed OICD logout error:

- Same errors above.
- On OICD logout, fetch dataset thumbnail fail error

## Motivation
Logout successfully should not show error message

- [ ] Docs updated?


martin-trajanovski (approved):
Great job! 💯

PR #507, opened 2023-05-21
feat: orgDatablock fetch retuns with datasetExist field

Junjiequan:
## Description

Included a customized pipelineStages to origDatablocks' fullquery service. 
So that when we trigger `GET /origdatablocks/fullquery/files`, the pipeline will add and return with `datasetExist `field that contains boolean value. This `datasetExist` value can be used on frontend to highlight origDatablocks with invalid dataset.

## Motivation

On Scicat files page, if you click an origDatablock that doesn't have valid dataset PID, it will redirect you to a empty dataset detail page. 

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2103,9 +2103,9 @@
       "dev": true
     },
     "node_modules/@eslint/js": {
-      "version": "8.40.0",
-      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-8.40.0.tgz",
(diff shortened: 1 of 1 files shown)


martin-trajanovski (approved):
Looks good to me but would be nice to have some tests if they are needed to cover the fixes.

PR #518, opened 2023-05-24
Swap 3250 post dataset with explicit pid

martin-trajanovski:
## Description

This PR aims to add the feature of creating datasets with explicit PID. There are few environmental variables introduced here and their purpose is:
* CREATE_DATASET_WITH_PID_GROUPS, create_dataset_with_pid_groups
* DATASET_CREATION_VALIDATION_ENABLED, dataset_creation_validation_enabled
* DATASET_CREATION_VALIDATION_REGEX, dataset_creation_validation_regex

As a user:

If I belong to a group that is listed as admin group (ADMIN_GROUPS), I can create a dataset with explicit pid
if I belong to a group that is listed in CREATE_DATASET_WITH_PID_GROUPS, I can create a dataset with explicit pid
if I belong to a group that is listed in CREATE_DATASET_GROUPS, I can create a dataset without pid. PID is assigned by the system

In usecase 1 and 2, if pid is not provided, the system automatically assigned a uuid v4 as dataset pid.
In use case 3, pid is automatically assigned as a uuid v4.

When pid is explicitly provided, the system should check that it is unique in the database.
If DATASET_CREATION_VALIDATION_ENABLED is set to TRUE, the system should validate the pid provided using the regex provided in DATASET_CREATION_VALIDATION_REGEX.

These configuration options should have the following default values:

```
CREATE_DATASET_GROUPS = "all" 
CREATE_DATASET_WITH_PID_GROUPS = "" 
DATASET_CREATION_VALIDATION_ENABLED = true 
DATASET_CREATION_VALIDATION_REGEX = "^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$"  
```

## Motivation

There was no option to create dataset with some explicit PID that is not generated by the system.

## Fixes:

* https://jira.esss.lu.se/browse/SWAP-3250

## Changes:

* changes made

- [ ] Docs updated?

### DIFF
--- a/src/origdatablocks/origdatablocks.controller.ts
+++ b/src/origdatablocks/origdatablocks.controller.ts
@@ -11,6 +11,7 @@ import {
   Query,
   HttpCode,
   HttpStatus,
+  BadRequestException,
 } from "@nestjs/common";
 import { OrigDatablocksService } from "./origdatablocks.service";
 import { CreateOrigDatablockDto } from "./dto/create-origdatablock.dto";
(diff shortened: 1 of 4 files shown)


src/datasets/datasets.controller.ts
```diff
@@ -210,18 +240,57 @@ export class DatasetsController {
...
+        ) {
+          throw new ForbiddenException(
+            "Unauthorized to create datasets with explicit PID",
+          );
+        } else {
+          // NOTE: If it can create dataset but not part of ADMIN_GROUPS and not part of CREATE_DATASET_WITH_PID_GROUPS,
+          // then we make sure that pid is not provided by user but it is generated by the system.
+          if (!isPartOfAdminGroups && !userCanCreateDatasetWithPid) {
+            delete dataset.pid;
+          }
```
nitrosx, line 266: Should we return an error if we are allowed to create a dataset but without pid?
I wonder if that is a better behavior than allow it and change the pid.

nitrosx (approved):

PR #523, opened 2023-05-26
fix: policies page pagination not working

Junjiequan:
## Description

Fixed page pagination for policies page.

Did not include API test for the fix. As in the DatasetLifecycle already included filter test for policies query and it works fine. Considering we are sending filter values in a mixed way of query & header, I think we should keep the test as it is now. 
The testing for filter was passing with pagination bug, because in the test it expects to receive filter from query, but the actual filter was sent through header, whereas query filter by default is set to `{}`. 

- [ ] Docs updated?


martin-trajanovski (approved):
Looks good to me

PR #603, opened 2023-07-19
Maxiv OIDC login patch

ddijken:
## Description

The login has a env variable for the redirect URL back to the frontend after OIDC login (env var successURL). We want a variable URL, so we intercept the request in the oidc guard to catch the referer and assign this to the session ID, and when it returns from keycloak we add back the referer so we can link back to the original request URL

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?


bpedersen2:
looks fine from the MLZ  side

nitrosx (approved):

PR #660, opened 2023-08-09
FEAT: elastic search integration

Junjiequan:
## Description

ElasticSearch integration for SciCat as optional feature. 
Disable elasticSearch will fall back to the regular mongoDB query.

## Motivation

MongoDB's aggregation method can yield low search accuracy. Longer queries often result in less relevant outcomes, frustrating users.

In contrast, Elasticsearch boasts a robust search engine. It indexes and stores data flexibly, allowing adaptable search methods based on user preferences. It promptly presents the most relevant information using document relevancy scores, significantly enhancing user experiences.

- [ ] Docs updated?
- [x] New packages used/requires npm install? 


enthusiastio:
Let me know if I should look into it while it is WIP

Junjiequan:
> Let me know if I should look into it while it is WIP

@enthusiastio This PR is ready for review now, I will change description by tomorrow : )

package.json
```diff
@@ -27,14 +27,16 @@
     "test:api": "npm run test:api:jest --maxWorkers=50% && concurrently -k -s first \"wait-on http://localhost:3000/explorer/ && npm run test:api:mocha\" \"npm run start\"",
     "test:api:jest": "jest --config ./test/config/jest-e2e.json --maxWorkers=50%",
     "test:api:mocha": "mocha --config ./test/config/.mocharc.json",
-    "prepare:local": "docker-compose -f CI/E2E/docker-compose-local.yaml up -d && cp functionalAccounts.json.example functionalAccounts.json"
+    "prepare:local": "docker-compose -f CI/E2E/docker-compose-local.yaml --env-file CI/E2E/.env.elastic-search up  -d && cp functionalAccounts.json.example functionalAccounts.json"
```
enthusiastio, line 30: I wonder if we didn t want to store the environmental variables in HELM charts, if there is any difference between environments

enthusiastio (approved):
To be honest not everything I comprehend, as I don; t have that much experience with scat, but to me overall this looks good, just a small comment

package.json
Junjiequan: @enthusiastio
We should store the ENV variables in the helm charts for dev environment. This prepare:local script is only for local usage, to spin up containers for local development.

nitrosx:
@Junjiequan I'm going to review it by tomorrow.

CI/E2E/.env.elastic-search
```diff
@@ -0,0 +1,18 @@
...
+MONGODB_COLLECTION=Dataset
+ES_MAX_RESULT=210000
+ES_FIELDS_LIMIT=400000
+ES_INDEX="dataset"
+ES_PORT=9200
+ES_HOST=https://localhost:9200
+ES_USERNAME=elastic
+ES_PASSWORD=password
+KIBANA_PORT=5601
+KIBANA_PASSWORD=password
```
nitrosx: Do we need all of them?

CI/E2E/backend.env
```diff
@@ -34,3 +34,21 @@ CREATE_DATASET_WITH_PID_GROUPS=""
...
+MONGODB_COLLECTION=Dataset
+ES_MAX_RESULT=210000
+ES_FIELDS_LIMIT=400000
+ES_INDEX="dataset"
+ES_PORT=9200
+ES_HOST=https://localhost:9200
+ES_USERNAME=elastic
+ES_PASSWORD=password
+KIBANA_PORT=5601
+KIBANA_PASSWORD=password
```
nitrosx: Do we need all of them here?

src/common/utils.ts
```diff
@@ -480,14 +481,17 @@ export const createFullfacetPipeline = <T, Y extends object>(
   fields: Y,
   facets: string[],
   subField = "",
+  esPids?: string[],
 ): PipelineStage[] => {
   const pipeline = [];
-  const facetMatch: Record<string, unknown> = {};
+  const facetMatch: Record<string, unknown> = esPids
```
nitrosx, line 487: I am wondering if we are going to see some performance issues if the list of pids in esPIDS becomes really long

src/config/configuration.ts
```diff
@@ -158,6 +158,29 @@ const configuration = () => {
...
+      encryptionKeys: process.env.ENCRYPTION_KEYS,
+      nodeExtraCaCerts: process.env.NODE_EXTRA_CA_CERTS,
+      mongoDBCollection: process.env.MONGODB_COLLECTION,
+      defaultIndex: process.env.ES_INDEX ?? "dataset",
+    },
+    kibana: {
+      enabled: process.env.KIBANA_ENABLED ?? "no",
+      port: parseInt(process.env.KIBANA_PORT || "5601", 10),
+      password: process.env.KIBANA_PASSWORD,
+    },
```
nitrosx: Do we need all this entries for elastic search?
Why do we need kibana configuration in scicat backend?

src/datasets/datasets.service.ts
```diff
@@ -41,22 +43,33 @@ import { DatasetClass, DatasetDocument } from "./schemas/dataset.schema";
...
     const createdDataset = new this.datasetModel(
       // insert created and updated fields
       addCreatedByFields(createDatasetDto, username),
     );
-
+    if (this.ESClient) {
+      this.ESClient.insertDocument(
+        createdDataset.toObject() as DatasetDocument,
+      );
+    }
```
nitrosx, line 72: How do we change which dataset fields we are inserting in elastic search?

src/datasets/datasets.service.ts
```diff
@@ -88,16 +100,33 @@ export class DatasetsService {
...
+      modifiers.skip,
+    );
 
     const datasets = await this.datasetModel
-      .find(whereClause, null, modifiers)
+      .find(
+        { _id: { $in: esResult.data }, ...remainingClauses },
+        null,
+        modifiers,
+      )
```
nitrosx, line 128: It seems to me that here the elastic search is always used in the search.
What happens if we disable elastic search in configuration?

src/datasets/datasets.service.ts
```diff
@@ -355,6 +419,15 @@ export class DatasetsService {
...
 
+  async getDatasetsWithoutId() {
+    try {
+      const datasets = this.datasetModel.find({}, { _id: 0 }).exec();
+      return datasets;
+    } catch (error) {
+      throw new NotFoundException();
+    }
+  }
+
```
nitrosx: What does this function do?

src/elastic-search/elastic-search.controller.ts
```diff
@@ -0,0 +1,138 @@
+import {
```
nitrosx, line 1: We need to implement permissions?
Do we need access to the underlying elastic search from scicat?

src/elastic-search/mappings/datasetFieldMapping.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const datasetMappings: MappingObject = {
+  description: {
+    type: "text",
+    analyzer: "autocomplete",
+    search_analyzer: "autocomplete_search",
+    fields: {
+      keyword: {
+        type: "keyword",
+        ignore_above: 256,
```
nitrosx, line 11: Do we really limit the search to 256 characters?

src/elastic-search/query-builder.service.ts
```diff
@@ -0,0 +1,98 @@
...
+      const searchTermArray = text.toLowerCase().split(" ");
+      const wildcardQueries = searchTermArray.flatMap((term) =>
+        queryFields.map((fieldName) => ({
+          wildcard: {
+            [fieldName]: {
+              value: `*${term}*`,
+            },
+          },
+        })),
+      );
```
nitrosx: Can we submit the query as it is to elastic search without splitting it in terms?

nitrosx (changes requested):

src/datasets/datasets.service.ts
Junjiequan:
```
    if (!this.ESClient || !$text) {
      const datasets = await this.datasetModel
        .find(whereClause, null, modifiers)
        .exec();
      return datasets;
    }
```
if we disable elastic search `!this.ESClient` will be true and it will execute only within the block

src/elastic-search/query-builder.service.ts
Junjiequan: We can do that, but then we lose the benefit of partial match with multiple terms to list results based on relevancy score.
One solution would be we provide a toggle button on the filter UI and let the user to decide whether they want to splitting terms or not.

src/elastic-search/elastic-search.controller.ts
Junjiequan: Yes we login to elastic search from scicat, but I don't think we need to implement permissions. Once elastic search is enabled anyone should be able to search through elastic search. And let the Datasource controller or service layer to decide what data to show which group of people - that being said, there's still improvement needed to achieve that

src/datasets/datasets.service.ts
Junjiequan:
```
  async insertDocument(data: DatasetDocument) {
    delete data._id;
    try {
      await this.esService.index({
        index: this.defaultIndex,
        id: data.pid,
        document: data,
      });
```
I think we need to specify here, let the `document: ` to only update given field.
Should be feasible to insert or update only specific field

src/datasets/datasets.service.ts
Junjiequan: elastic search itself generates `_id` for each document, so we need to remove `_id` that we get from mongo in order to do the data sync.
But since we already have PID field, we can simply specify { _id:0 } to retrive data from mongoDB without id

src/config/configuration.ts
Junjiequan: I will cleanup this

CI/E2E/backend.env
Junjiequan: I'll cleanup these, only keeping necessary ones

src/common/utils.ts
Junjiequan: Fetching 20K records seems no performance issue yet.
But I have included a condition for where we need to fetch all data. if there's no text in the search query, elastic search will not be executed and we use mongo query to retrieve all data.
```
   if (!this.ESClient || !$text) {
      const datasets = await this.datasetModel
        .find(whereClause, null, modifiers)
        .exec();
      return datasets;
    }
```

src/elastic-search/mappings/datasetFieldMapping.ts
Junjiequan: Type keyword means that the text will be stored as a whole which can only be searched by the entire text. So, unless an user want to search by writing an article, character limits under 256 should be more than enough

CI/E2E/docker-compose-local.yaml
bpedersen2, line 1: I would keep a non-es local compose as well

.github/workflows/test.yml
bpedersen2, line 1: Probably needs an e2e test with elasticsearch enabled as well

CI/E2E/docker-compose-local.yaml
```diff
@@ -16,12 +16,139 @@ services:
...
+        "."
       ]
     volumes:
       - ".env.scichat-loopback:/home/node/app/.env"
     depends_on:
       - mongodb
+
+  esSetup:
+    image: docker.elastic.co/elasticsearch/elasticsearch:${STACK_VERSION}
+    volumes:
```
bpedersen2: I would setup dev/test without TLS

src/datasets/datasets.service.ts
nitrosx: WE should specify the document fields that will be inserted in ES in scicat BE configuration.
Let's open a new issue regarding this topic

src/datasets/datasets.service.ts
nitrosx: Could would you please insert a comment that this function is used only when a manual sync is triggered.

src/elastic-search/elastic-search.controller.ts
nitrosx: Check how dataset controller implements the authorization.
When I check the code the ES endpoint did not specify any authorization on any of them!!!

src/elastic-search/query-builder.service.ts
nitrosx: Either way is fine by me, but if we split the query, we have to do something more elaborate according to NLP methodologies!!!
Have you checked if ES can perform the query splitting internally?

.github/workflows/test.yml
nitrosx: I agree. Let's open an issue to address this

bpedersen2:
I would prefer if we squash (or manually rebase and squash in smaller groups) to get a cleaner history before merging.

PR #719, opened 2023-09-08
E2e fix

bpedersen2:
Adjust e2e tests to current frontend master usage

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -159,7 +159,7 @@ jobs:
         uses: cypress-io/github-action@v6
         with:
           working-directory: frontend
-          config-file: CI/ESS/e2e/cypress.nestjs.json
+          config-file: CI/ESS/e2e/cypress.github.ts
           install-command: npm install --omit peer
           browser: chrome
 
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -159,7 +159,7 @@ jobs:
         uses: cypress-io/github-action@v6
         with:
           working-directory: frontend
-          config-file: CI/ESS/e2e/cypress.nestjs.json
+          config-file: CI/ESS/e2e/cypress.github.ts
```
Junjiequan, line 162: I have a question.
Can we remove the cypress.nest.json from FE after this change?
It looks like currently we have 3 cypress.xxx.ts file with almost exact configuration values

Junjiequan (approved):

.github/workflows/test.yml
bpedersen2: No, this uses the one from frontend (check the working directory)

But I agree, we should create a single e2e setup.

bpedersen2: I created #1187

PR #723, opened 2023-09-08
Use separate eslint/prettier step

bpedersen2:
and remove posttest in package json.

That way lint errors are much more visible in the test stages output.


bpedersen2:
needs #722 to be merged first

nitrosx (approved):

PR #765, opened 2023-09-25
fix: implement authorization on origDatablock

Junjiequan:
## Description

A user can create an orig datablock only if he/she is allowed to update the dataset specified.

- [ ] Docs updated?

### DIFF
--- a/src/origdatablocks/origdatablocks.controller.ts
+++ b/src/origdatablocks/origdatablocks.controller.ts
@@ -91,12 +91,15 @@ export class OrigDatablocksController {
     });
 
     if (dataset) {
-      const datasetInstance = new DatasetClass();
-      datasetInstance.ownerGroup = dataset.ownerGroup;
+      // NOTE: We need DatasetClass instance because casl module
+      // can not recognize the type from dataset mongo database model.
+      // If other fields are needed can be added later.
+      const newDatasetClass = new DatasetClass();
+      newDatasetClass.ownerGroup = dataset.ownerGroup;
 
       if (user) {
         const ability = this.caslAbilityFactory.createForUser(user);
-        const canUpdate = ability.can(Action.Update, datasetInstance);
+        const canUpdate = ability.can(Action.Update, newDatasetClass);
         if (!canUpdate) {
           throw new ForbiddenException("Unauthorized access");
         }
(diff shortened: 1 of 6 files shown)


src/origdatablocks/origdatablocks.controller.ts
nitrosx, line 1: Do we have the same tests on the other endpoints?
Also are we testing for all the different use cases?

Junjiequan: There's another ticket for creating test cases, I haven't start to working on the ticket yet https://jira.esss.lu.se/browse/SWAP-3334

nitrosx:
@Junjiequan unfortunately I need to review completely the datasets authorization model, than I can complete this PR review.
I think there are inconsistencies between Datasets and OrigDatablocks permissions that needs to be addressed.

src/origdatablocks/origdatablocks.controller.ts
```diff
@@ -263,12 +347,26 @@ export class OrigDatablocksController {
   )
   @Get("/fullfacet")
   async fullfacet(
+    @Req() request: Request,
     @Query() filters: { fields?: string; facets?: string },
   ): Promise<Record<string, unknown>[]> {
+    const user: JWTUser = request.user as JWTUser;
+    const fields: IOrigDatablockFields = JSON.parse(filters.fields ?? "{}");
```
bpedersen2, line 354: there are a lot of similar blocks like this,
I guess we should try to refactor it to use a common function. 
something like (pseudo-code):
```
function accessGroupsFilterFields( fields_in, user, OrigDataBlock) {
  if (user) {
   ability = ...
  canViewAll = ...
 if ( !canViewAll ){
   ...
     fields_in.userGroups.push(...)
   }
return fields_in
}
```

bpedersen2 (changes requested):
Generally looks fine, but a bit too much code seems duplicated which makes later maintenance harder.

src/origdatablocks/origdatablocks.controller.ts
nitrosx: @bpedersen2 I will postpone code refactoring to a new issue.
I'm going through a review of all the authorization for datasets and connected entities, which will require some changes.
We chat tomorrow during the meeting if you are present.

Junjiequan: I've created new issue to address this issue #821

nitrosx (approved):

Issue #762, opened 2023-09-21
Automate merge for dependabot PR


nitrosx:
### Summary
Dependabot is a great tool, but at times, the number of PR that needs to be reviewed and merged becomes not manageable.
I wonder if we can auto-merge dependabot PR that successfully pass the tests.

Here some material which says how to set-up the automerge, pros and cons:
- https://dev.to/slashgear_/how-to-automatically-merge-dependabot-pull-requests-with-github-actions--30pe

At the SciCat meeting help on 2023/09/20, collaborators expressed in favor to auto merge.

Please comment and provide feedback

PR #778, opened 2023-09-27
chore: auto-merge dependabot patch versions

Junjiequan:
## Description
Added new workflow to auto merge dependabot package version bump PR when all checks pass.
Solves https://github.com/SciCatProject/scicat-backend-next/issues/762

The logic of `Dependabot auto-merge` workflow are:
1. checks if **pull request user** is _**dependabot**_
2. approve the pull request made by dependabot, if the bump is either **_minor or patch level_**. ( TOKEN with PR approval permission is required )
3. enabling auto merge, so the PR is **_only get merged after all checks passes_** (auto-merge prerequisite can be configured)

auto-merge prerequisite configuration can be done in **Settings > Branches > Edit or Add rule**

image

- [ ] Docs updated?

### DIFF
--- a/.github/workflows/auto-merge-dependabot.yaml
+++ b/.github/workflows/auto-merge-dependabot.yaml
@@ -1,17 +1,50 @@
-name: Auto merge dependabot
+name: Dependabot auto-merge
 
-on:
-  pull_request:
-    branches:
-      - master
+on: pull_request_target
+
+permissions:
+  pull-requests: write
+  contents: write
+  packages: read
 
 jobs:
-  auto-merge-dependabot:
+  dependabot:
     runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.user.login == 'dependabot[bot]' }}
     steps:
-      - uses: actions/checkout@v4
-        ## https://github.com/ahmadnassri/action-dependabot-auto-merge
-      - uses: ahmadnassri/action-dependabot-auto-merge@v2
+      ## Extract information about the dependencies being updated by a Dependabot-generated PR
+      - name: Dependabot metadata
+        id: dependabot-metadata
+        uses: dependabot/fetch-metadata@v1.6.0
         with:
-          target: patch
-          github-token: ${{ secrets.DEPENDABOT_TOKEN  }}
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/auto-merge-dependabot.yaml
```diff
@@ -0,0 +1,17 @@
...
+    branches:
+      - master
+
+jobs:
+  auto-merge-dependabot:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        ## https://github.com/ahmadnassri/action-dependabot-auto-merge
+      - uses: ahmadnassri/action-dependabot-auto-merge@v2
```
bpedersen2: instead of using an intransparent third-party action, check https://github.com/SciCatProject/frontend/pull/1216/files#diff-2fe48d1021006468ac63263583b5db615dd5e63a39ebff862cc4e2f06b59caa1 

its just  3 or 4 lines more without, and we have much more control.

nitrosx: I agree with @bpedersen2. Con we try the other solution?

nitrosx (changes requested):

.github/workflows/auto-merge-dependabot.yaml
Junjiequan: With that solution it seems like we need to enable auto-merge, I'm not sure if that's a good idea.
Most importantly, it doesn't check whether all the tests are passing or not..
I need to test the script further

Junjiequan:
https://github.com/ahmadnassri/action-dependabot-auto-merge

.github/workflows/auto-merge-dependabot.yaml
bpedersen2: My version just sets  automerge per pull-request if it is coming from dependabot.  So changes get merge if tests pass.

Junjiequan: I see, enable auto merge will check if all the requirements passed. Then we just need to include an PR approved flow it should work, because with current merge rule at least 1 approval is required

bpedersen2 (approved):
Looks f ine.

nitrosx (approved):

PR #818, opened 2023-10-17
Add "create dataset privileged" groups and review dataset authorization

nitrosx:
This PR adds a configuration group that allows users to be able to create datasets for other groups (ingestion) and implement the correct dataset authorization.

- [ ] Docs updated?


Junjiequan:
@bpedersen2, @nitrosx 
Backend e2e test is removed for now, because of the inconsistency caused by different backend and frontend images, as well as changes in authorization, requiring either the frontend or backend to be forcibly merged with failing e2e tests.
In the future, we could apply monorepo strategy to execute e2e, instead of relying on docker images.

bpedersen2 (approved):
Looks essential fine  to me.

bpedersen2:
I would prefer for better reviewability that this branch get rebase on the current dev head and force-pushed (possbily also squashing it a bit down to useful commits ( I like this approach: https://github.blog/2022-06-30-write-better-commits-build-better-projects/).

src/users/users.controller.ts
```diff
@@ -54,6 +53,18 @@ export class UsersController {
...
+    pid?: string,
+  ): Promise<DatasetClass> {
+    const datasetInstance = new DatasetClass();
+    datasetInstance._id = "";
+    datasetInstance.pid = pid || "";
+    datasetInstance.accessGroups = user.currentGroups || [];
+    datasetInstance.ownerGroup = user.currentGroups[0] || "";
+
+    return datasetInstance;
+  }
```
nitrosx: Why is this function here? Should it belong to the dataset sub-system?

test/CheckDifferentDatasetTypes.js
nitrosx, line 1: Why did we remove this tests?

nitrosx (commented):
It looks really good. Nice work.

test/CheckDifferentDatasetTypes.js
Junjiequan: We shouldn't remove this, I think I've removed it accidentally

src/users/users.controller.ts
Junjiequan: I forgot to remove this. Previously it was intended to check what kind of dataset users can create - withpid, or without pid for `/:id/authorization/dataset/create` endpoint.  But after discussion we decided to return only true or false of whether user can create dataset, since the boolean value will be used on Frontend  to display `CreateDataSet` button

nitrosx: Great. Thank for clarifying

test/CheckDifferentDatasetTypes.js
nitrosx: Great. As long as it passes

nitrosx (commented):
I think we are ready to merge

PR #820, opened 2023-10-17
feat: add better check for logout urls

martin-trajanovski:
## Description

Improve the resolution of the logout_url by checking for end_session endpoint.

### DIFF
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -58,46 +58,80 @@ export class AuthService {
       "expressSessionSecret",
     );
 
+    const logoutResult = await this.additionalLogoutTasks(req, logoutURL);
+
     if (expressSessionSecret) {
       req.logout(async (err) => {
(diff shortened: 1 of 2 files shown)


src/auth/auth.service.ts
```diff
@@ -58,46 +58,80 @@ export class AuthService {
...
+        for (const j in temp) {
+          result[i + "." + j] = temp[j];
+        }
+      } else {
+        result[i] = obj[i];
+      }
+    }
+    return result;
+  };
+
```
Junjiequan: I'd suggest move this function to the common/utility file

Junjiequan (approved):
just a small comment, otherwise looks good to me

src/auth/auth.service.ts
martin-trajanovski: @Junjiequan Thanks for the review and the good comment. It is ready now.

PR #823, opened 2023-10-18
build(tsconfig): include only src when building the project

martin-trajanovski:
## Description

When building the backend we should include only the `src` folder and everything else needed is added either when building the docker image or mounted when running it in a container.

## Motivation

We had this setup before and after upgrading some packages and changing configuration the structure of the `dist` folder changed where some unnecessary files and folders were included.

## Changes:

* changes are only in the build configuration where instead of excluding lot of things we just include the `src` and then `src` folder is not added inside the `dist` but only it's content.

**Before**

before

**After**

after

- [ ] Docs updated?


tsconfig.build.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "./tsconfig.json",
-  "exclude": ["node_modules", "test", "dist", "**/*spec.ts"]
+  "include": ["src"]
```
bpedersen2: We still want to exclude `**/*spec.ts` ( or actually `**/*.spec.ts`) for the build.

bpedersen2 (changes requested):
And we need to find the reason for the failing e2e-tests

tsconfig.build.json
martin-trajanovski: Yes you are right about this. Thanks for the feedback.

bpedersen2:
Something seems still missing, as the e2e tests all fail when trying to authenticate. Suspects are:
 - functionalAccounts.json
 - migrations and related stuff (they are in the current dist dir

bpedersen2:
Needs change in CI/E2E/Dockerfile, as `main.js` moves with this change.

Will also affect all custom builds

bpedersen2 (approved):
Looks good to go, but don't forget to do the Dockerfile change also on frontend /CI/ESS/e2e/Dockerfile, otherwise  we get trouble there.

martin-trajanovski:
> Needs change in CI/E2E/Dockerfile, as `main.js` moves with this change.
> 
> Will also affect all custom builds

@bpedersen2 this seems that was the issue and now it should be resolved.

bpedersen2:
@nitrosx: Should  the migration dir and migration.ts get included in the production build?

martin-trajanovski:
> @nitrosx: Should the migration dir and migration.ts get included in the production build?

As far as I know they are explicitly included in our `Dockerfile` after we have built the app. https://github.com/SciCatProject/scicat-backend-next/blob/master/Dockerfile#L46

PR #864, opened 2023-11-02
Add a security policy

jl-wynen:
This is good practice so that users know what to do when they find a problem.

I basically copied this from Scitacean. So it may not be exactly what you want here. In particular, please 
- Let me know what email address I can insert.
- Check if the supported versions info is correct.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,7 +6,7 @@ Only the latest version is supported.
 
 ## Reporting a Vulnerability
 
-If you believe you have found a security vulnerability in Scitacean, please
+If you believe you have found a security vulnerability in SciCat, please
 - ✅ report it to us by creating a [security advisory](https://github.com/SciCatProject/scicat-backend-next/security/advisories/new).
 - ❌ do not report security vulnerabilities through public GitHub issues, discussions, or pull requests, etc.
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,19 @@
+# Security Policy
+
+## Supported Versions
+
+Only the latest version is supported.
+
+## Reporting a Vulnerability
+
+If you believe you have found a security vulnerability in SciCat, please report it to us through coordinated disclosure.
+
+**Please do not report security vulnerabilities through public GitHub issues, discussions, or pull requests.**
+
+Instead, please email <ADDRESS>
+
+Please include as much information as you can to help us better understand and resolve the issue.
+
+## Disclosure
+
+We use GitHub [security advisories](https://github.com/SciCatProject/scitacean/security/advisories) to disclose fixed vulnerabilities.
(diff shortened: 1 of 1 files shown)


bpedersen2 (commented):
Good idea. 

We should discuss in the meeting whether  we want support at least two minr -version branches for security updates (or a LTS version and one faster moving branch with regular releases (e.g. the jenkins model: one release from master every 2 weeks and a LTS release every 3 month) to cater for facilties that cannot support too frequent updates?

nitrosx:
@jl-wynen based on the file that you added, we should have a log of all the vulnerabilities that we have fixed.
How do you propose to do that in a way that minimize the maintenance effort.
Thanks

On a side note, what else do we need to do on this PR to approve and merge?

jl-wynen:
Just use GH security advisories. They allow you to track what has been fixed and for what version and disclose those when appropriate.

Related, there is actually no need for an email address. You can enable security reporting on the repo which is like issues but can be opened by anyone but seen only by maintainers.

SECURITY.md
```diff
@@ -0,0 +1,19 @@
...
+
+**Please do not report security vulnerabilities through public GitHub issues, discussions, or pull requests.**
+
+Instead, please email <ADDRESS>
+
+Please include as much information as you can to help us better understand and resolve the issue.
+
+## Disclosure
+
+We use GitHub [security advisories](https://github.com/SciCatProject/scitacean/security/advisories) to disclose fixed vulnerabilities.
```
Junjiequan: After change the URL to https://github.com/SciCatProject/scicat-backend-next/security/advisories, we should get it merged

Junjiequan (changes requested):

SECURITY.md
jl-wynen: Updated the link

SECURITY.md
```diff
@@ -0,0 +1,19 @@
...
+
+Only the latest version is supported.
+
+## Reporting a Vulnerability
+
+If you believe you have found a security vulnerability in SciCat, please report it to us through coordinated disclosure.
+
+**Please do not report security vulnerabilities through public GitHub issues, discussions, or pull requests.**
+
+Instead, please email <ADDRESS>
```
jl-wynen: We still need to put an email address here.

Alternatively, or in addition, you can also enable private security reporting for the repo: https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository

Junjiequan: @jl-wynen  Thanks for the info. I have enabled the security report feature for this repository. 
@nitrosx What email should we put in the template?

Junjiequan (approved):

PR #865, opened 2023-11-06
add: added heatlh check endpoint

Junjiequan:
## Description

Added a health endpoint to allow external services to confirm backend readiness before initiating dependent services.

Utilizing this endpoint ensures frontend e2e tests via Cypress run only after backend is fully operational, instead of relying on `sleep 5 `command.

- [ ] Docs updated?


.github/workflows/test.yml
```diff
@@ -204,6 +204,18 @@ jobs:
           path: frontend
           fetch-depth: 1
 
+      - name: Wait for Backend
+        run: |
+          npm install -g wait-on && \
+          wait-on http://localhost:3000/api/v3/health --timeout 200000
+        # run: |
```
martin-trajanovski: Maybe the commented part should be removed completely.

Junjiequan: Yes I will remove them, I need to re-run this test more times to see if it is stable

Junjiequan: Yes I will remove them, I need to re-run this test more times to see if it is stable

martin-trajanovski (approved):
This looks good to me now

nitrosx (approved):

PR #894, opened 2023-11-16
Patch out vulnerable vm2 with an override

bpedersen2:
fixes critical issues in vm2, see https://github.com/patriksimek/vm2/security by using degenerator 5+, which does not require vm2.

- [X] New packages used/requires npm install?


bpedersen2:
Should fix https://github.com/SciCatProject/scicat-backend-next/security/dependabot/39 and https://github.com/SciCatProject/scicat-backend-next/security/dependabot/38

nitrosx (approved):

PR #923, opened 2023-11-28
updated jobs module

nitrosx:
## Description
Reviewed and updated job module file

- [ ] Docs updated?


HayenNico (approved):
Looks good. We may be able to throw out the import for PoliciesModule at a later point, since its components were only used in the custom email notification code in jobs.service. For now we should leave it in.

PR #925, opened 2023-11-30
Issue 924: add support for arrays in scientificMetadata

### LINKED ISSUE #924: add support for arrays in scientificMetadata
### Summary

At DESY we often store some parts of scientificMetadata in arrays/lists e.g. vectors
```
"vector": {
    "value": [1,1,0],
    "unit": "mm"
}
```
However, I such case the current algorithm to convert value and units to SI crashes.

### Expected Behaviour

I would like to support also  arrays/lists in  scientificMetadata as well.

jkotan:
## Description

It resolves #924  by adding support for arrays in scientificMetadata

## Motivation

At DESY we often use a list to store vectors

## Changes:

* I add `convertArrayToSI` function which is called by `appendSIUnitToPhysicalQuantity`

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -19,12 +19,12 @@ export const convertArrayToSI = (
 ): { valueSI: number[]; unitSI: string } => {
   try {
     if(inputValue && inputValue.length){
-       const quantity = unit(inputValue[0], inputUnit).toSI().toJSON();
+       const newUnit = unit(inputUnit).toSI().toJSON().unit;
        const value = Array.from(
 	       inputValue,
-	       (iValue) => unit(iValue, inputUnit).toSI().toJSON().value
+	       (iValue) => unit(iValue, inputUnit).to(newUnit).toJSON().value
        );
-      return { valueSI: value, unitSI: quantity.unit };
+      return { valueSI: value, unitSI: newUnit };
    }
     else {
     return { valueSI: inputValue, unitSI: inputUnit };
@@ -41,7 +41,7 @@ export const convertToSI = (
   inputUnit: string,
 ): { valueSI: number; unitSI: string } => {
   try {
-    const quantity = unit(inputValue, inputUnit).toSI().toJSON();
+    const quantity = unit(inputValue, inputUnit).to(unit(inputUnit).toSI().toJSON().unit).toJSON();
     return { valueSI: Number(quantity.value), unitSI: quantity.unit };
   } catch (error) {
     console.error(error);
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -12,6 +12,30 @@ import {
 } from "./interfaces/common.interface";
 import { ScientificRelation } from "./scientific-relation.enum";
 
+
+export const convertArrayToSI = (
+  inputValue: number[],
+  inputUnit: string,
+): { valueSI: number[]; unitSI: string } => {
+  try {
+    if(inputValue && inputValue.length){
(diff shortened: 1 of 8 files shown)


jkotan:
Hi,
on the other hand if you don't want to store additional SI values for array because it takes too much space, maybe it would be good to skip arrays conversion without throwing the error as in the current code 
(or/and create env VARIABLE to define if you convert or not the arrays to SI or the size of maximal array to be converted).

jkotan:
I've just added a workaround for #926

src/common/utils.ts
```diff
@@ -12,12 +12,34 @@ import {
 } from "./interfaces/common.interface";
 import { ScientificRelation } from "./scientific-relation.enum";
 
+export const convertArrayToSI = (
+  inputValue: number[],
+  inputUnit: string,
+): { valueSI: number[]; unitSI: string } => {
+  try {
+    const newUnit = unit(inputUnit).toSI().toJSON().unit;
```
bpedersen2, line 20: can add a comment for this workaround, so that we can easier find to roll back these changes once upstream mathjs has fixed that behaviour ( and if possible  reference a mathjs bug report as well)

bpedersen2 (commented):
Except  for some additional  comment requests, this looks fine

bpedersen2 (changes requested):
see my previous comments

nitrosx (commented):
This looks fine to me, although it does not address the following:
- API tests to test specifically this feature
- fullquery logic and syntax to deal with this new type of metadata.
- UI to view and edit this type of metadata (this should be a FE PR)

Potentially we could merge this PR, as long as we start working right away on the other two tasks right.
@Junjiequan and myself just discussed about this and we think that it should be discussed at the next meeting.

src/common/utils.ts
```diff
@@ -12,12 +12,36 @@ import {
 } from "./interfaces/common.interface";
 import { ScientificRelation } from "./scientific-relation.enum";
 
+export const convertArrayToSI = (
```
sbliven, line 15: Can you add a docstring?

jkotan: Hello @sbliven, 
I'm not sure what should be in the docstring. In the case of `convertArrayToSI` the code does not contain the workaround. I think it is better to convert the unit only once and then apply it with use of `.to()` as it is in the code now. If you would like to execute `.toSI()` for every list item you convert the unit for each item so it does not seem to be optimal.

bpedersen2:
Please check the frontend edit function as well ( it currently seems to truncate  the scientific metadata if it encounters an array in the values)!

src/common/utils.ts
bpedersen2: #1005 bumped to version with fixed mathjs (see https://github.com/josdejong/mathjs/blob/develop/HISTORY.md#2024-01-12-1230) so we could drop the workaround now.

jkotan:
> Please check the frontend edit function as well ( it currently seems to truncate the scientific metadata if it encounters an array in the values)!

I think the frontend  has a separate repository so one need to open a ticket in the dedicated repository

sbliven:
What's the status of this PR?

nitrosx:
We should review this request given that now metadata should contain the range type.
Also, we need to define better the user case and make sure that related work is also done in the frontend

Junjiequan:
Hi, please confirm if this PR is still relevant within the next two weeks.
Otherwise, we’ll plan to close it to keep the repo tidy.

Issue #941, opened 2023-12-08
OpenAPI Schema


ChuckMoe:
### Summary
Currenlty the OpenAPI schemas are not working as intended. The required parameters in the schema do not match the required parameters for the API and some are even opposite to each other.

### Steps to Reproduce
I created a python client from the OpenAPI schema json file from my local SciCat instance. Generating, installing and using it works just fine. When creating the `CreateRawDatasetDto` object in the client, I have to add the parameter `owner` because it is set as `required` in the OpenAPI schema. Doing this results in an `Error: Bad Request - whitelistValidation` response from the API though.

### Suggestion
Invert the class inheritance:
UpdatingClassDto as smallest subset of parameters
CreatingClassDto inherits from UpdatingClassDto
GettingClassDto (or only ClassDto) inherits from CreatingClassDto

This way you build up your parameter space and only show the parameters one is allowed to use in an API call. You can find a small practical example of this in Python [here](https://github.com/awtkns/fastapi-crudrouter/issues/109#issuecomment-1107664425)

### Notes
I can work on this issue.

nitrosx:
@ChuckMoe thank you for your offer to work on the issue. We do appreciate.
Please go ahead and submit a PR when you are ready for us to review.
If you could include API tests (both positives and negative) for the issues that you reported, it would be awesome.

Please let me know if you have any question or need help, we will do our best to support your effort

ChuckMoe:
Do you mean to test the API or the API specification (OpenAPI schema)? The API looks fine to me and I don't want to change that. It's basically it's documentation I want to change.

At the moment, there are two ways to test the API specifications that I can think of. One is to generate the client, then install and test if it works against the SciCat API. The other one is to save the OpenAPI specification as json file, parse it and check for inconsistencies manually. Both do have their pros and cons.

nitrosx:
Apologies for not being clear, I meant add tests to test for the issue that you just found under the folder `tests` that you find in code.
That's where what we call the "API test" are stored and can be run by `npm run test:api:mocha`

ChuckMoe:
I just opened a PR that shows how it can be done. For the example I choose the Instrument and Sample classes. This can be adapted for all classes.

## Instrument
### Create
The body for creating an Instrument is still the same
```
{
  "uniqueName": "string",
  "name": "string",
  "customMetadata": {}
}
```

### Update
But the allowed bodies for updating now look different.

#### Old
```
{
  "uniqueName": "string",
  "name": "string",
[...]

PR #956, opened 2023-12-14
Dependencies inverted for DTO's

ChuckMoe:
# Draft
## Description

This is an example for issue #941.
Schemas are built up from the smallest subset, in order to only allow the user to input the allowed variables.

## Motivation

Take a look at #941

## Fixes:

* Restricts API to only use (create, update, read) allowed values

## Changes:

* Instead of extending from the base class and saying only part of the data can be passed in, I reversed the dependencies and built up the variable space

- [ ] Docs updated?


nitrosx (commented):
@ChuckMoe am I correct if if I say that you define the update dto for a specific entity and from there you inherit the class for the create dto adding the missing fields?
I was wondering if you can explain why you decide to take this approach?
Thanks

ChuckMoe:
@nitrosx
Creating the UpdateDto class and using it as a base for the CreateDto class has the advantage that you can explicitly "exclude" certain parameters you do not want to be updatable, like an id.

Here an example.
```
class UpdateDto {
  readonly color: string;
}

class CreateDto extends UpdateDto {
  readonly id: int;
}
```

Now the following would be valid requests.
```
POST /create
{
  "id": 1,
  "color": "yellow"
}

POST /update/1
{
  "color": "green"
}
```

These would be invalid requests now though.
```
POST /update/1
{
  "id": 2,
  "color": "green"
}
```

Before this PR, the UpdateDto extended the CreateDto and made it partial, which theoretically allowed for updating parameters that were not supposed to be updated as well. This is also reflected in the generated OpenAPI schema in /explorer and helps for understanding and automatic generation of code as well.

ChuckMoe:
I would very much like to go through all of the classes with someone to identify if all of the variables are correctly placed, or if I need to move them between the update and create class.

nitrosx:
@ChuckMoe I will be available to go over with you as a first pass. 
I will try to open start a discussion about this PR in today's meeting.

ChuckMoe:
> I will try to open start a discussion about this PR in today's meeting.

Good morning. Any news on this @nitrosx ?

nitrosx:
Unfortunately, I was not able to initiate the discussion at the last meeting.
Should we raise the priority?

nitrosx:
@bpedersen2 Can you provide your opinion on this?

bpedersen2:
Sounds like a valid idea. Lets try to put this on the agenda for tuesday (This time I will be there again).

@ChuckMoe: could you promote this from draft to an active pull request?

nitrosx:
@ChuckMoe and @bpedersen2 I started to review once more this PR, but I realized that I quickly got lost.
I have a feeling that I/we need a well defined method to verify if we have the correct fields in the correct model.

Would a table with the fields and a column for each DTO work? Thoughts and feedback?

That said I embrace and support the idea to invert the dependencies.

ChuckMoe:
Good morning @nitrosx. A table with a column `updatable` should be enough. That would be really awesome to have. Going through all of the DTOs will be very quick then.

nitrosx:
@ChuckMoe can you provide an example please?
I will try to merge your PR as soon as I am able to

ChuckMoe:
Here is an example table that would help going through all the DTO's. The _Optional_ column is not necessarily needed for this PR but good for overview and to check if everything is as it is supposed to be.

### Sample

| Variable | Optional | Can Update |
| --- | --- | --- |
| ownerGroup | False | True |
| accessGroups | True | True |
| instrumentGroups | True | True |
| sampleId | True | False |
| owner | True | True |
| description | True | True |
| sampleCharacteristics | True | True |
| isPublished | True | True |

src/proposals/dto/update-proposal.dto.ts
```diff
@@ -1,10 +1,119 @@
...
+    type: Date,
+    required: false,
+    description: "The date when data collection finishes.",
+  })
+  @IsOptional()
+  @IsDateString()
+  readonly endTime?: Date;
+
+  @ApiProperty({
+    type: MeasurementPeriodClass,
```
fpotier: I think this should be a `CreateMeasurementPeriodDto` rather than a `MeasurementPeriodClass`

ChuckMoe: You are correct, thank you very much!

bpedersen2 (approved):

PR #957, opened 2023-12-14
Use package version for OpenAPI

ChuckMoe:
## Description

Uses environment variable set by node for the OpenAPI version number.

## Motivation

Keeps both version numbers in sync

## Changes:

* changed from hard coded string to environment variable

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -20,7 +20,7 @@ async function bootstrap() {
   const config = new DocumentBuilder()
     .setTitle("SciCat backend API")
     .setDescription("This is the API for the SciCat Backend")
-    .setVersion("4.0.0")
+    .setVersion("" + process.env.npm_package_version)
     .addBearerAuth()
     .build();
   const document = SwaggerModule.createDocument(app, config);


bpedersen2 (commented):
Looks Ok to me

nitrosx (commented):
@ChuckMoe just to confirm, you are setting the version of swagger OpenAPI to the version of the OpenAPI swagger npm package version? Is that correct?

Originally this was supposed to be version of the SciCat API.
This PR would change the meaning of it.
If that's the case, I'm not opposed, but I think it should be discussed at the next meeting.

Currently the swagger interface shows the SciCat API version and the OpenAPI version (see image)
image

Is that what you intent to do or you want the npm library version?

ChuckMoe:
@nitrosx This change does not target the OpenAPI version number but the SciCat Version number, aka. the grey box in your picture. It just automatically keeps it in sync.

nitrosx:
Thank you for clarifying.

nitrosx (approved):

PR #965, opened 2023-12-19
feat: integrate gray logger to improve scicat error logging

Junjiequan:
## Description

@user-office-software/duo-logger package is installed and gray logger is configured.

## Motivation

Enabling grayLogger allows SIMS team to get error messages in the first place.

- [ ] Docs updated?


src/elastic-search/elastic-search.service.ts
```diff
@@ -85,9 +90,18 @@ export class ElasticSearchService implements OnModuleInit {
         await this.createIndex(this.defaultIndex);
       }
       this.connected = true;
-      Logger.log("Elasticsearch Connected");
+      Logger.log("Elasticsearch Connected", "SerchService");
+      logger.logInfo("ElasticSearch Connected.", {
+        host: this.host,
+        username: this.username,
+        password: this.password,
```
bpedersen2: Please never log passwords!

src/elastic-search/elastic-search.service.ts
```diff
@@ -69,6 +70,10 @@ export class ElasticSearchService implements OnModuleInit {
       Logger.error(
         "Missing ENVIRONMENT variables for elastic search connection",
       );
+      logger.logError(
```
martin-trajanovski: I am a bit confused here about logging the same message twice with two different loggers. Is this really needed?

bpedersen2 (changes requested):

src/logger/logger.service.ts
```diff
@@ -0,0 +1,42 @@
...
+          "GrayLogger is enabled but server/port environment variable is not configured",
+          "GrayLogger Initilazation failed",
+        );
+      } else {
+        setLogger(
+          new GrayLogLogger(
+            this.server,
+            parseInt(this.port),
+            {
+              facility: "DMSC",
```
bpedersen2: harcoded facility... needs to be configurable

src/elastic-search/elastic-search.service.ts
bpedersen2: https://docs.nestjs.com/techniques/logger#using-the-logger-for-application-logging 

gives a way on how to use a single logger instance, soit seems you would never need to directly instantiate a custom class here.

Junjiequan: The idea was to use nestJS logger for development and gray logger to log critical errors on graylogs. I'm not sure if we want to send every errors to grayLog

bpedersen2: Then maybe check https://docs.nestjs.com/techniques/logger#use-external-logger as well, it recommends to use e.g. winston for production setups ( and winston allows to configure different/multiple loggers and filtering by level etc.)

Junjiequan: winston looks like a good option for configuring external loggers. I will talk about this after Max comes back from the holiday

nitrosx (commented):
I feel that this is an ESS specific change, aka graylog.
I think we should have a wider discussion and see if we can generalize it.
Also, I agree with @bpedersen2 that we should have a single logger in SciCat which can be configured to send the log to multiple logging facilities according to the IT infrastructure where the instance is running

bpedersen2 (commented):
Question: does it fallback to use the default logger with out a loggers.json present,  or is minimal loggers.json required?

src/loggers/logger.service.ts
```diff
@@ -0,0 +1,127 @@
...
+  constructor(private configService: ConfigService) {}
+
+  async onModuleInit() {
+    let loggerConfigs = this.configService.get<LoggerConfig[]>("loggerConfigs");
+
+    const filePath = "loggers.json";
+    if (fs.existsSync(filePath)) {
+      const data = fs.readFileSync(filePath, "utf8");
+      loggerConfigs = JSON.parse(data);
+    }
```
nitrosx: I would move this section of code in the configuration controller. I would like to follow the concept that if you instantiate the configuration class, you have the full configuration and avoid loading portions of the configuration in other classes.
@bpedersen2 thoughts?

src/elastic-search/elastic-search.service.ts
```diff
@@ -161,10 +162,13 @@ export class ElasticSearchService implements OnModuleInit {
...
       });
-      Logger.log(`Elasticsearch Index Created-> Index: ${index}`);
+      Logger.log(
+        `Elasticsearch Index Created-> Index: ${index}`,
+        "Elasticsearch",
+      );
       return HttpStatus.CREATED;
     } catch (error) {
-      Logger.error("createIndex failed-> ElasticSearchService", error);
+      console.error("createIndex failed-> ElasticSearchService", error);
```
nitrosx: Why not using Logger here too?

src/elastic-search/elastic-search.service.ts
nitrosx, line 1: I was wondering why you are using `console.log` and not Logger?

src/loggers/logger.service.ts
```diff
@@ -0,0 +1,127 @@
...
+            });
+          } catch (err) {
+            console.error(err);
+          }
+        }),
+      );
+    }
+  }
+
+  private handleLogForwarding(method: string, ...args: unknown[]): void {
```
nitrosx: I was wondering why you are doing this dynamic mapping of the methods instead of declaring an abstract class and make sure that all the loggers implement such class?

src/config/configuration.ts
```diff
@@ -37,6 +35,7 @@ const configuration = () => {
   // Logger.log("- Update job groups : " + updateJobGroups);
 
   return {
+    loggerConfigs: [],
```
nitrosx: Why don't we load the logger configuration in this class instead doing it in the logger class below?
Please see my comment below on logger.service.ts

nitrosx (changes requested):

src/loggers/logger.service.ts
bpedersen2: Yes, sounds better. Then we can ensure useful defaults.

src/elastic-search/elastic-search.service.ts
Junjiequan: This is for our use case specific change, if you see the catch block it contains log errors and throw errors. Since

Junjiequan:
> Question: does it fallback to use the default logger with out a loggers.json present, or is minimal loggers.json required?

@bpedersen2  minimal loggers.json is required to enable default logger. and the configuration can be found in loggers.json.example

src/elastic-search/elastic-search.service.ts
Junjiequan: if we set both logger and `throw` here, two identical errors will be sent to graylogger, one from logError and one from exception. 
Previously I've included console because `throw` was not showing any logs. It is now fixed, and Loggers from all catch blocks are removed.

src/config/configuration.ts
```diff
@@ -22,10 +22,26 @@ const configuration = () => {
...
+  const jsonConfigMap: { [key: string]: object[] } = {};
+  const jsonConfigFileList = ["loggers"];
+  for (const fileName of jsonConfigFileList) {
+    const filePath = `${fileName}.json`;
+    if (fs.existsSync(filePath)) {
+      const data = fs.readFileSync(filePath, "utf8");
+      jsonConfigMap[fileName] = JSON.parse(data);
+      break;
+    }
+  }
```
nitrosx: I would make it more generic, so we can easily use it for the functional accounts and future development.
How about something like this:
```typescript
  const jsonConfigFileList = {
    "loggers" : process.env.LOGGERS_CONFIG_FILE || "loggers.json",
  };
 Object.keys(jsonConfigFileList).forEach( key => {
    const filePath = jsonConfigFileList[key];
    if (fs.existsSync(filePath)) {
      const data = fs.readFileSync(filePath, "utf8");
      jsonConfigMap[fileName] = JSON.parse(data);
      break;
    }
  });
```

nitrosx (changes requested):

src/config/configuration.ts
```diff
@@ -53,7 +61,7 @@ const configuration = () => {
   // Logger.log("- Update job groups : " + updateJobGroups);
 
   return {
-    loggerConfigs: jsonConfigMap.loggers ?? [defaulgLogger],
+    loggerConfigs: jsonConfigMap.loggers || [defaulgLogger],
```
bpedersen2: s/defaulgLogger/defaultLogger/ (and find other instances of that typo)?

bpedersen2 (changes requested):

bpedersen2 (approved):

nitrosx (approved):

Issue #958, opened 2023-12-15
[Idea] OpenAPI Generation


ChuckMoe:
The OpenAPI definition allows to automatically generate clients for different languages. We could publish them with every release. On top of that, we can use these generated clients to test if the OpenAPI definition matches the internal implementation.

## Steps 
1. [x] Save the OpenAPI definition as a json file
2. [x] Generate clients with the just generated json file
3. [x] Write tests for at least one generated client and run them against the scicat-backend
4. [x] Publish the generated clients

## Questions
- Save json file with every startup or only when, for example, an environment variable is set?
- Which clients should we generate?
- Which language to choose for the tests? Typescript, Python, ...?

## References
- [OpenAPI Generator Action](https://github.com/openapi-generators/openapitools-generator-action)

nitrosx:
SDK are generate automatically. The OpenAPI json file is retrieved directly from an instance of the related version of the BE.
SDK(s) are provided as is and tests is out of the scope of autogeneration. Testing is left to users.

Closing the issue as SDK are generated and published at every release.

Issue #584, opened 2023-07-06
Incorrect schema for history


jl-wynen:
The schema in the json-explorer claims that `Dataset.history` is of type `#/components/schemas/HistoryClass`. In the Swagger explorer, it is shown as
```
description: | List of objects containing old and new values.

_id* | string default: 3638926a-0f07-4128-99c8-dda3138eadc0Persistent  Identifier for datasets derived from UUIDv4 and prepended automatically  by site specific PID prefix like 20.500.12345/
updatedAt* | string($date-time) default: 1686225387712Time when the update was performed.
updatedBy* | string($date-time) default: 1686225387712Username of the user that performed the update.
```

When downloading a dataset, the actual field contains a (potentially empty) list. If it is nonempty, the list contains old versions of the dataset instead of the special `HistoryClass`.

Is this an error in the exported schema or in the API endpoint?

dylanmcreynolds:
Is this a migration issue where there is old backend data that needs to get translated to a newer schema?

PR #969, opened 2023-12-22
Save OpenAPI definition as json, generate client and run tests

ChuckMoe:
See #958


.github/workflows/test.yml
```diff
@@ -174,3 +174,52 @@ jobs:
...
+
+      - name: Restore node_modules
+        id: cached-node-modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ env.NODE_VERSION }}
+
+      # Todo: May need reworking in order to not block
+      - name: Start Backend for later tests and generate openapi.json
```
bpedersen2: Please use the  same flow as in the API tests (l.132ff)  ( you need to start a mongodb instance and use the correct configs)

ChuckMoe:
To further work on this PR, I have some questions:

1. Which languages do you want to generate code for now? For all of them, please create repositories and add the credentials as environment variables in the project settings.
2. At the moment I am using [schemathesis](https://github.com/schemathesis/schemathesis) to run automatically generated tests. This is not the best solution, but it would us get started immediately. In theory. It seems like the backend and the definition are still not aligned. To try it our yourself, execute schemathesis locally.
3. If I start the node server in one part of a step and want to call its API in another part, is it still running or simply exited?
4. Should I create some rudimentary tests for one client in order to test it now?

### Run Schemathesis
#### Install
```python
pip3 install --user schemathesis
```
#### Run
Prepare the group variables in your env in a way, that your user is allowed everything.
```bash
st run \
  -H "Authorization: Bearer TOKEN" \
  --checks all \
  --validate-schema true \
  http://localhost:3000/explorer-json
```

.github/workflows/publish_openapi_client.yaml
```diff
@@ -0,0 +1,102 @@
...
+        with:
+          generator: ${{ matrix.generator }}
+          openapi-file: openapi.json
+          config-file: .github/openapi/${{ matrix.generator }}-config.yml
+
+      # Checkout client and force upload changes
+      - name: Git Push to its client library repository
+        run: |
+          cd ${{ matrix.generator }}
+          /bin/sh ./git_push.sh
```
bpedersen2, line 102: file seems to be missing, but we have to discuss if we go more in the direction of a monorepo ( then the autogenerated clients would be part of the backend repo). 
@nitrosx :  topic for meeting

bpedersen2 (changes requested):
Some minor changes, but overall this looks good.

.github/workflows/publish_openapi_client.yaml
ChuckMoe: The file `git_push.sh` is also generated by the OpenAPI generator and then exists in the generated directory.

ChuckMoe:
The list of generated clients can also be moved into an environment variable of the repository. That way, there is no need to change the code in the case you want to add or remove one.

.github/workflows/test.yml
```diff
@@ -174,3 +175,77 @@ jobs:
...
+          MEM_LIMIT: 4G
+
+        # Start mongo container and app before starting
+        run: |
+          cp CI/ESS/docker-compose.api.yaml docker-compose.yaml
+          cp functionalAccounts.json.test functionalAccounts.json
+          docker-compose up --build -d
+          npm run start
+
+      - name: Fetch Bearer Token
```
bpedersen2: probably needs a wait-for-it to have a running backend

.github/workflows/test.yml
```diff
@@ -174,3 +175,77 @@ jobs:
...
+
+      - name: Restore node_modules
+        id: cached-node-modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ env.NODE_VERSION }}
+
+      # Generate OpenAPI definition
+      - name: Start Backend for later tests and generate openapi.json
```
bpedersen2: should run the  version without elastic-search

bpedersen2 (changes requested):

.github/workflows/test.yml
ChuckMoe: I did disable it with the environment variable. What are you referencing / how do I use the version without elastic-search?

bpedersen2: That just tells scicat not to use elastic search , but still fires up es from the docker-compose file. 
From the docker-compose yuo should only need mongodb ( so a docker-compose up -d mongodb should be sufficient

.github/workflows/test.yml
```diff
@@ -174,3 +175,76 @@ jobs:
...
+          ACCESS_TOKEN=$(curl --retry 30 --retry-connrefused --retry-delay 20 -d 'username=ingestor' -d 'password=aman' http://localhost:3000/api/v3/auth/login | jq -r ".access_token")
+
+      # Test the generated OpenAPI definition
+      # https://github.com/marketplace/actions/schemathesis
+      - name: Run schemathesis against the backend
+        uses: schemathesis/action@v1
+        with:
+          schema: 'http://127.0.0.1:3000/explorer-json'
+          wait-for-schema: '60'  # max wait time in seconds
+          args: '-H "Authorization: Bearer ${{ env.ACCESS_TOKEN }}"'
```
bpedersen2: some problems:

1.  authentication is not right
    
2. please disable the upload to schemathesis:
    https://github.com/SciCatProject/scicat-backend-next/actions/runs/7623826706/job/20764677373#step:7:200

3. Only run tests for configured endpoints (e.g. auth should only use auth/login, while  ldap, msad, oidc are not available in the test instance)
4. empty strings are not good input


My suggestion: exclude the schema test from this PR, it could come later once you have a non-failing setup.

bpedersen2 (changes requested):

.github/workflows/test.yml
ChuckMoe: Using Schemathesis was mostly to try out, if we can get tests running quickly.

To answer some of your points @bpedersen2:

1. What is the problem with the authentication?
3. Sadly one can not exclude endpoints but I can explicitly define which endpoints (or endpoint patterns) to test.
4. Property testing and that is what Schemathesis is doing, does not test _good_ but _valid_ input, that's the point. An API should be able to handle valid input.

How should I proceed from here on @nitrosx? If I drop the test, this PR still includes the generation of the clients. Generating clients without testing is not the wisest choice.

docker-compose.mongo.yaml
```diff
@@ -0,0 +1,11 @@
+services:
+  mongodb:
+    image: bitnami/mongodb:4.2
+    volumes:
+      - mongodb_data:/bitnami
+    ports:
```
dylanmcreynolds: Leaving the port open to a database can be problematic (our IT department scans for this and will shut off network access).

I generally do one of two things:


If I really need access to mongo from other processes on my machine:
```
ports:
   - 127.0.0.1:27017:27017

```

If I just need to let services from other docker compose files see this, define a network for all docker-compose (this is something required even within a single docker-compose for things like podman on Mac, which we are starting to use.)

dylanmcreynolds (approved):
This looks great to me. I had one quick comment on the new mongo docker-compose. While reading, I hadn't realized that that was probably there for the github action and not local development, so maybe my comment does not really apply.

dylanmcreynolds (approved):
This looks great to me. I had one quick comment on the new mongo docker-compose. While reading, I hadn't realized that that was probably there for the github action and not local development, so maybe my comment does not really apply.

docker-compose.mongo.yaml
ChuckMoe: Yeah, this is only used for the Github action. I can also move this file into the `.github` folder to make it more clear?

ChuckMoe:
I tried writing the tests for the generated `javascript`-client and `typescript`-client but both had some problems. Is it alright, if I write the tests for the python client?

ChuckMoe:
One big open question for me is still where to save/upload the client code to. Is it supposed to be a mono-repo or should they be uploaded into different repositories?

.github/workflows/test.yml
```diff
@@ -170,7 +171,31 @@ jobs:
 
         # Start mongo container and app before running api tests
         run: |
-          cp CI/ESS/docker-compose.api.yaml docker-compose.yaml
           cp functionalAccounts.json.test functionalAccounts.json
-          docker-compose up --build -d
-          npm run test:api
+          docker-compose up --build -d -f CI/ESS/docker-compose.api.yaml
```
bpedersen2: caution, this is inside a matrix run, so it will do concurrent api generation on both a elastic-search enabled BE and a plain backend (see line 105ff. )

I would suggest to move the api generation before that block ( and add a depends condition, so that the ordering is correct.

openapi/python-nextgen/test/test_partial_update_derived_dataset_dto.py
```diff
@@ -0,0 +1,112 @@
...
+                'ingestor'
+            ],
+            instrument_group='instrument',
+            pid=str(random.randint(0, 99999)),
+            owner='ingestor',
+            owner_email='pi@email.com',
+            orcid_of_owner='',
+            contact_email='pi@email.com',
+            source_folder='/',
+            source_folder_host='http://folder.host',
```
bpedersen2: please always use an officially invalid domain name (e.g. .invalid ) ,see   https://en.wikipedia.org/wiki/.invalid

bpedersen2: RFC: https://datatracker.ietf.org/doc/html/rfc2606

bpedersen2 (changes requested):

openapi/python-nextgen/test/test_partial_update_derived_dataset_dto.py
ChuckMoe: I didn't know these, thanks!

.github/workflows/test.yml
```diff
@@ -125,7 +125,7 @@ jobs:
       - name: Start backend
         env:
           EXIT_AFTER_OPENAPI_GENERATION: true
-          ELASTICSEARCH_ENABLED: ${{ matrix.elasticsearch_enabled }}
+          ELASTICSEARCH_ENABLED: yes
```
bpedersen2: I would use no here,  no need to fire up elastic search just  for api generation

bpedersen2:
Test still fail, as you are not waiting  for mongodb (and Es,if you really want it running) long enough (start-up takes up to 2 min.) See `CI/E2E/docker-compose.yml` for the health checks used there.

.github/workflows/test.yml
ChuckMoe: I confused this variable a little and thought that without it, the API endpoints will not be generated in the definition as well. I now know this is not correct and will disable it.

ChuckMoe:
I do not need Elastic Search for the generation but do you have an idea on how to wait for MongoDB to be ready?

.github/workflows/test.yml
ChuckMoe: It was decided to write tests for one language.

.github/workflows/test.yml
```diff
@@ -170,7 +171,170 @@ jobs:
 
         # Start mongo container and app before running api tests
         run: |
-          cp CI/ESS/docker-compose.api.yaml docker-compose.yaml
           cp functionalAccounts.json.test functionalAccounts.json
-          docker-compose up --build -d
-          npm run test:api
+          docker-compose up --build -d -f CI/ESS/docker-compose.api.yaml
```
bpedersen2: wrong  docker-compose syntax

` docker-compose  -f CI/ESS/docker-compose.api.yaml up --build -d`

(`-f` is a global flag, `--build` and `-d` are flags for up)

bpedersen2 (changes requested):

ChuckMoe:
This PR is now blocked by #1055

.github/workflows/test.yml
ChuckMoe: Thanks! That was it.

nitrosx:
I just checked and the test fails for in relation to this bug: https://github.com/OpenAPITools/openapi-generator/issues/11090
I wonder if we could try the option --skip-validate-spec or if there something invalid in the openapi file

.github/workflows/publish_openapi_client.yaml
```diff
@@ -0,0 +1,102 @@
...
+          ACCESS_GROUPS_STATIC_VALUES: "ess"
+          PROPOSAL_GROUPS: ingestor
+          SAMPLE_GROUPS: ""
+          DATASET_CREATION_VALIDATION_ENABLED: false
+          MONGODB_COLLECTION: Dataset
+          MEM_LIMIT: 4G
+
+        # Start mongo container and app before starting
+        run: |
+          cp CI/ESS/docker-compose.api.yaml docker-compose.yaml
```
nitrosx, line 68: We should create a new docker compose file that starts on mongodb.

.github/workflows/test.yml
```diff
@@ -171,7 +172,162 @@ jobs:
...
+          ES_FIELDS_LIMIT: 400000
+          ES_INDEX: "dataset"
+          ES_REFRESH: "wait_for"
+          STACK_VERSION: 8.8.2
+          CLUSTER_NAME: es-cluster
+          MEM_LIMIT: 4G
+
+        # Start mongo container and app before running api tests
+        run: |
+          docker-compose -f CI/ESS/docker-compose.api.yaml up --build -d
```
nitrosx, line 244: We should cretae a new docker compose file that starts only mongodb. We do not use ES for the openapi generation

.github/workflows/test.yml
```diff
@@ -171,7 +172,162 @@ jobs:
...
+          until curl -s -f -o /dev/null "localhost:27017"; do sleep 5; done
+          npm run start
+
+      - name: Generate openapi client
+        uses: openapi-generators/openapitools-generator-action@v1
+        with:
+          generator: python-pydantic-v1
+          openapi-file: openapi.json
+          config-file: .github/openapi/python-config.yml
+          command-args: "--ignore-file-override openapi/python/test**"
```
nitrosx, line 254: Could we add the option `--skip-validate-spec` as mentioned in issue [11090](https://github.com/OpenAPITools/openapi-generator/issues/11090)

.github/workflows/test.yml
```diff
@@ -171,7 +172,162 @@ jobs:
...
+      - name: Restore node_modules
+        id: cached-node-modules
+        uses: actions/cache@v4
+        with:
+          path: node_modules
+          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ env.NODE_VERSION }}
+
+      - name: Start backend
+        env:
+          ELASTICSEARCH_ENABLED: yes
```
nitrosx, line 278: Turn ES off. No need to have it on to test openapi generate libraries

openapi/python-nextgen/test/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+    'user1': soc.CredentialsDto(
+        username='user2',
+        password='a609316768619f154ef58db4d847b75e'),
+    'user2': soc.CredentialsDto(
+        username='user2',
+        password='f522d1d715970073a6413474ca0e0f63'),
+    #
+    'nouser': soc.CredentialsDto(
+        username='nouser',
+        password='f522d1d715970073a6413474ca0e0f63'),
```
nitrosx, line 26: Can we pull the passwords in from configuration or functionalAccounts.json file?

openapi/python-nextgen/test/test_elastic_search_api.py
nitrosx, line 1: Can we leave ES out from this tests at least for the first pass?

src/main.ts
```diff
@@ -31,6 +33,11 @@ async function bootstrap() {
     .addBearerAuth()
     .build();
   const document = SwaggerModule.createDocument(app, config);
+  fs.writeFileSync("./openapi.json", JSON.stringify(document, null, 2));
+  if (process.env.EXIT_AFTER_OPENAPI_GENERATION) {
+    process.exit();
+  }
+
```
nitrosx, line 40: I am not too convince that this is the best solution!!!
The openapi json file can be downloaded once the BE is running.
Instead of making such a drastic changes to the core, could we download the open api file from the URL?

nitrosx: I need to verify and test it, but apparently you retrieve the openapi.json file by hitting the url `http://localhost:3000/explorer-json` once the scicat BE container is up and running and is accessible on port 3000.
Feature to be documented

nitrosx (changes requested):
@ChuckMoe please let me know if you are still able to work on this PR or somebody else should take over.
thank you so much for all your work and help.

nitrosx:
@bpedersen2 can check my comments and provide your opinion? Thanks

bpedersen2:
`

> I just checked and the test fails for in relation to this bug: [OpenAPITools/openapi-generator#11090](https://github.com/OpenAPITools/openapi-generator/issues/11090) I wonder if we could try the option --skip-validate-spec or if there something invalid in the openapi file

The error seems genuine ( The default for the time in HistoryClass is a timestamp, not a date-time. ) See #584

bpedersen2:
> @ChuckMoe please let me know if you are still able to work on this PR or somebody else should take over. thank you so much for all your work and help.

Currently I will try and take it over

nitrosx:
Thank you @bpedersen2.
Do you want to sync sometimes prior our meeting in Munich regarding this?

ChuckMoe:
It is still connected to issue #1055. The commit to this issue did not fix the problem stated in it.

bpedersen2:
I guess the  mongoose-managed  fields ( createdAt )etc. should  have a default in the schema definition for the HistoryClass to match  the other schemas.

bpedersen2:
#1132 should fix the HistoryClass errors

nitrosx:
Regarding the test failing: are we able to download the openapi.json file and run the api creation successfully locally?
Do we have instructions on how to do that?

ChuckMoe:
@nitrosx You can download or create it yourself but it won't change the outcome. The same error will appear until the example value of `updatedAt` in issue #1055 is fixed. You could use `--skip_validation` but that would also skip every other potential error in the schema, so I advocate against it.
If you want to create the schema locally take a look at [this tutorial on how to run OpenAPI on your machine](https://openapi-generator.tech/docs/installation/).

nitrosx:
@ChuckMoe thank you for clarifying.
I will look into it and make sure that somebody from the community will work on it.

sbliven:
Is this PR superseded by the [Core Release SDK autogeneration and pySciCat](https://github.com/orgs/SciCatProject/projects/16) project? The plan is to generate the SDK in a different repo, right? Should this PR be closed?

nitrosx:
As mentioned by @sbliven, this PR as been superseeded by the completion of [Core Release SDK autogeneration and pySciCat](https://github.com/orgs/SciCatProject/projects/16) project.

Closing it

PR #986, opened 2024-01-03
Issue 926: workaround for wrong degC to K conversion

### LINKED ISSUE #926: "300 degC" wrongly converted to "300 K"
### Summary

"300 degC" is wrongly converted to "300 K". Probably it is a bug in mathjs. I tested with (10.2.4).

### Steps to Reproduce

to reproduce it one can store a scalar value with 'degC' unit or  simpler in nodejs
```js
 unit("300 degC").toSI().toJSON()


### Current Behaviour

```json
{ mathjs: 'Unit', value: 300, unit: 'K', fixPrefix: true }
```

### Expected Behaviour

```js
{ mathjs: 'Unit', value: 573.15, unit: 'K', fixPrefix: true }
```

### Extra Details

Workaround for the problem could be to change the above line of the code in convertToSI to something like
```js
 unit(inputValue, inputUnit).to(unit(inputUnit).toSI()).toJSON()
```

which gives the correct result
```js
{ mathjs: 'Unit', value: 573.15, unit: 'K', fixPrefix: true }
```

jkotan:
It resolves #926  as it is mentioned in the ticket by changing `.to()` instead of `.toSi()`

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -17,7 +17,10 @@ export const convertToSI = (
   inputUnit: string,
 ): { valueSI: number; unitSI: string } => {
   try {
-    const quantity = unit(inputValue, inputUnit).toSI().toJSON();
+    // Workaround related to a bug reported at https://github.com/josdejong/mathjs/issues/3097 and https://github.com/josdejong/mathjs/issues/2499
+    const quantity = unit(inputValue, inputUnit)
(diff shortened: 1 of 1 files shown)


nitrosx (approved):

nitrosx:
Thank you @jkotan

PR #997, opened 2024-01-10
info for user logged in

nitrosx:
## Description
This PR adds 3 endpoints that return the info regarding the user that is currently logged in without the need to specify the user id

## Motivation
In same cases, I need to perform some action where I need to retrieve information about the user I am currently logged in. We should be able to do so without providing my id.
The system knows which user I am based on the JWT provided.

## Changes:
* Added 3 endpoints in user_controller
* Added output dto for login function and users endpoint

- [ ] Docs updated?


Junjiequan (approved):
Looks fine to me

PR #1017, opened 2024-01-19
Update mongoose to 8.0.4

bpedersen2:
Upgrade mongoose to 8.0.4 and fix  resulting problems

 - Model.updateOne/Many does not have a new option  (this option is only available on findAndReplace)
 - CountOptions is not available as a type  and not compatible with QueryOptions ( may be an upstream error)


bpedersen2:
Superseeds: #1001

Junjiequan (approved):

PR #1021, opened 2024-01-19
Improve JobConfig

sbliven:
- Fix static/instance issues associated with `JobType.actionType` by adding `getActionType()` accessor
- Add `auth` property to create/update. These are now called `JobOperation` in the data model. The job configuration schema changed slitely
- Fix some eslint issues. If this causes too many conflicts it can be reverted.

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -911,3 +911,14 @@ const replaceLikeOperatorRecursive = (
 export const sleep = (ms: number) => {
   return new Promise((resolve) => setTimeout(resolve, ms));
 };
+
+/**
+ * Accepts either an object or array of objects.
+ * 
+ * Note that T should not extend Array (eg no nested arrays)
+ * @param x object or array of objects
+ * @returns array containing x
+ */
+export function oneOrMore<T>(x: T[]|T): T[]  {
+  return Array.isArray(x) ? x : [x];
+}
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -914,11 +914,11 @@ export const sleep = (ms: number) => {
 
 /**
  * Accepts either an object or array of objects.
- * 
+ *
  * Note that T should not extend Array (eg no nested arrays)
  * @param x object or array of objects
  * @returns array containing x
  */
-export function oneOrMore<T>(x: T[]|T): T[]  {
+export function oneOrMore<T>(x: T[] | T): T[] {
   return Array.isArray(x) ? x : [x];
 }
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -161,9 +161,9 @@ const configuration = () => {
  * Registers built-in JobActions. Should be called exactly once.
  */
 export function registerDefaultActions() {
-  registerCreateAction(LogJobAction.actionType, (data) => new LogJobAction(data));
-  registerUpdateAction(LogJobAction.actionType, (data) => new LogJobAction(data));
-  registerCreateAction(EmailJobAction.actionType, (data) => new EmailJobAction(data));
+  registerCreateAction(LogJobAction);
+  registerUpdateAction(LogJobAction);
+  registerCreateAction(EmailJobAction);
 }
 
 export type OidcConfig = ReturnType<typeof configuration>["oidc"];
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -1,5 +1,10 @@
 import { Logger } from "@nestjs/common";
-import { JobConfig, loadJobConfig, registerCreateAction, registerUpdateAction } from "./jobconfig";
+import {
+  JobConfig,
+  loadJobConfig,
+  registerCreateAction,
+  registerUpdateAction,
+} from "./jobconfig";
 import { LogJobAction } from "./actions/logaction";
 import { EmailJobAction } from "./actions/emailaction";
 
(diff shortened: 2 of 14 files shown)


src/common/utils.ts
```diff
@@ -911,3 +911,14 @@ const replaceLikeOperatorRecursive = (
...
 };
+
+/**
+ * Accepts either an object or array of objects.
+ *
+ * Note that T should not extend Array (eg no nested arrays)
+ * @param x object or array of objects
+ * @returns array containing x
+ */
+export function oneOrMore<T>(x: T[] | T): T[] {
```
sbliven, line 922: Is it a bad idea to allow the configuration to have either a single object or an array of objects? I'm not sure whether this is convenient or confusing.

test/config/jobconfig.json
```diff
@@ -1,52 +1,53 @@
...
-            "user": "scicatuser",
-            "pass": ""
-          }
-        },
-        "to": "{{ join job.recipients ' '}}",
-        "from": "scicat@example.com",
-        "subject": "Archive job {{job.pid}} has completed.",
-        "body": "Dear {{job.owner}},\n\nArchive job {{job.pid}} has completed. Results: {{job.status}}\n"
+    "update": {
+      "auth": "#all",
```
sbliven, line 10: Here's an example of configuring authorization at the JobOperation level.

src/config/jobconfig.ts
```diff
@@ -19,172 +19,236 @@ import { ApiProperty } from "@nestjs/swagger";
...
+  }
+
+  /**
+   * Parse job configuration json by dispatching to currently registered JobActions
+   * @param data JSON
+   * @returns
+   */
+  static parse(data: Record<string, any>) {
+    const type = data["jobType"];
+    const create = JobOperation.parse<CreateJobDto>(
```
sbliven, line 53: Is this a good spelling for this concept? We were struggling to find a good term for jobType ('archive'), operator ('update') and actionType ('email').

src/config/jobconfig.ts
```diff
@@ -19,172 +19,236 @@ import { ApiProperty } from "@nestjs/swagger";
...
+  /**
+   * Validate the DTO, throwing an HttpException for problems
+   */
+  validate: (dto: DtoType) => Promise<void>;
+  /**
+   * Respond to the action
+   */
+  performJob: (job: JobClass) => Promise<JobClass>;
+  /**
+   * Return the actionType for this action. This should match the class's
```
sbliven, line 128: The static `JobAction.actionType` is used in parsing the config file. Method `jobAction.getActionType()` is used when executing the action (eg in error logs).
The type system doesn't enforce that they match, but hopefully this comment is enough.

test/config/jobconfig.json
```diff
@@ -1,52 +1,53 @@
...
+          },
+          "to": "{{ join job.recipients ' '}}",
+          "from": "scicat@example.com",
+          "subject": "Archive job {{job.pid}} has completed.",
+          "body": "Dear {{job.owner}},\n\nArchive job {{job.pid}} has completed. Results: {{job.status}}\n"
         },
-        "body": "{ structure to create archive files }"
-      }
+        {
+          "actionType": "filter",
```
sbliven, line 34: Not yet implemented, but I think this is how we will implement action-level filters.

src/common/utils.ts
nitrosx: Generally speaking, it might be confusing for the user and a simpler solution might be available. 
That said, it might be that it is needed in some cases.

We should check if we use it anywhere in the code

src/config/jobconfig.ts
nitrosx: How about this:
- Job type: archive
- Job phase: create, update, post
- Action type: email, log or else
A job phase can have one or more actions.

src/common/utils.ts
nitrosx: On a second thought, I would define oneOrMore as an array of one or more items of the same type, but an array nevertheless.

src/config/configuration.ts
```diff
@@ -161,9 +166,9 @@ const configuration = () => {
  * Registers built-in JobActions. Should be called exactly once.
  */
 export function registerDefaultActions() {
-  registerCreateAction(LogJobAction.actionType, (data) => new LogJobAction(data));
-  registerUpdateAction(LogJobAction.actionType, (data) => new LogJobAction(data));
-  registerCreateAction(EmailJobAction.actionType, (data) => new EmailJobAction(data));
+  registerCreateAction(LogJobAction);
+  registerUpdateAction(LogJobAction);
+  registerCreateAction(EmailJobAction);
```
nitrosx, line 171: @sbliven I am still confused why we want to register the actions? and if this is the best practice in typescript.

src/config/jobconfig.ts
nitrosx: I'm confused on this comment.
What do you mean by `the system does not enforce that they match`? Do you mean JobAction.actionType and the value returned by JobAction.GetActionType()?

test/config/jobconfig.json
nitrosx: Nice idea!!!

test/config/jobconfig.json
nitrosx: We should use this opportunity to provide a plain description of the job and workflow, and create the matching configuration.
This would be a good exercise for us to find strengths and shortcomings and to be able to create good quality documentation

nitrosx (approved):
@sbliven @despadam I left some additional comments, but I think we should go ahead, merge this PR in the release job branch and review the code for the release as a whole

PR #726, opened 2023-09-11
RFC: Configurable access groups

bpedersen2:
Introduce easier overridabel compile time configuration and use it for a configurable accessgroups service.

## Motivation

Configuring the access groups service needed a custom service provider factory. This needed to get copied during the build, making it hard to validate for all facilities. 

This change introduces a new override config file, that can be used  for injecting advanced configuration objects that are not environment vars( but as this is merged, allows to keep most defaults easily).

## Tests included/Docs Updated?

- [ ] Included for each change/fix?
- [ needs manual test for graphql ] Passing? (Merge will not be approved unless this is checked) 
- [ no ] Docs updated?
- [ ] New packages used/requires npm install? 
- [ ] Toggle added for new features?

PR #1041, opened 2024-01-30
Feat: enable dynamic profile field mapping via env variables for oidc response

Junjiequan:
## Description

This PR introduces configurable userInfo, enabling teams to customize returned values for the frontend. Default behavior remains unchanged if no configuration is provided. Notably, `displayName` in `oidc.strategy` now concatenates `givenName` and `familyName`.

## Motivation

Different teams like to show different things in their user profiles. For example, at ESS, we use oidc.sub as the username. This change lets each team pick what works best for them, making sure our system suits everyone's needs.

## Fixes:
added userInfoMapping field in the configuration file, which can be customized using the following environment variable name.
```
      userInfoMapping: {
        id: process.env.OIDC_USERINFO_MAPPING_FIELD_ID,
        username: process.env.OIDC_USERINFO_MAPPING_FIELD_USERNAME,
        displayName: process.env.OIDC_USERINFO_MAPPING_FIELD_DISPLAYNAME,
        familyName: process.env.OIDC_USERINFO_MAPPING_FIELD_FAMILYNAME,
        emails: process.env.OIDC_USERINFO_MAPPING_FIELD_EMAILS,
        email: process.env.OIDC_USERINFO_MAPPING_FIELD_EMAIL,
        thumbnailPhoto: process.env.OIDC_USERINFO_MAPPING_FIELD_THUMBNAIL_PHOTO,
        groups: process.env.OIDC_USERINFO_MAPPING_FIELD_GROUPS,
      },
```

## Changes:

* changes made

- [ ] Docs updated?


src/auth/strategies/oidc.strategy.ts
```diff
@@ -23,6 +23,11 @@ import { UserProfile } from "src/users/schemas/user-profile.schema";
 import { OidcConfig } from "src/config/configuration";
 import { AccessGroupService } from "../access-group-provider/access-group.service";
 import { UserPayload } from "../interfaces/userPayload.interface";
+import { IUserInfoMapping } from "src/common/interfaces/common.interface";
+
+type UserInfoResPonseWithGroups = UserinfoResponse & {
```
nitrosx: Change `UserInfoResPonseWithGroups` to `UserInfoResponseWithGroups`

src/auth/strategies/oidc.strategy.ts
```diff
@@ -77,12 +83,19 @@ export class OidcStrategy extends PassportStrategy(Strategy, "oidc") {
...
     };
-    userProfile.accessGroups =
+    const accessGroups =
       await this.accessGroupService.getAccessGroups(userPayload);
 
+    if (userProfile.groups && userProfile.groups.length > 0) {
+      userProfile.accessGroups = [...accessGroups, ...userProfile.groups];
+    } else {
+      userProfile.accessGroups = accessGroups;
+    }
```
nitrosx: To allow more flexibility and to update groups only in one place, this should be done within getAccessGroups.

src/users/schemas/user-profile.schema.ts
```diff
@@ -25,6 +25,9 @@ export class UserProfile {
 
   @Prop({ type: [String] })
   accessGroups: string[];
+
+  @Prop()
+  groups?: string[];
```
nitrosx: Do we need groups?

src/auth/strategies/oidc.strategy.ts
```diff
@@ -131,34 +145,57 @@ export class OidcStrategy extends PassportStrategy(Strategy, "oidc") {
...
+    // set mappings like OIDC_USERINFO_MAPPING_FIELD_USERNAME=family_name.
+    // This assigns userinfo.family_name to oidcUser.username.
+
+    const oidcUser: IUserInfoMapping = {
+      id: userinfo["sub"] || (userinfo["user_id"] as string) || "",
+      username: userinfo["sub"] || "",
+      displayName: userinfo["name"] || "",
+      familyName: userinfo["family_name"] || "",
+      email: userinfo["email"] || "",
+      thumbnailPhoto: (userinfo["thumbnailPhoto"] as string) || "",
```
nitrosx: Do we get the photo from oidc?

nitrosx (changes requested):

src/auth/strategies/oidc.strategy.ts
Junjiequan: if we want the merging process to be done within getAccessGroups then we need to get this PR merged https://github.com/SciCatProject/scicat-backend-next/pull/726/files

src/auth/strategies/oidc.strategy.ts
Junjiequan: We can get photo from oidc, but I'm not sure if we are using it or not, this field was defined long ago

src/auth/strategies/oidc.strategy.ts
```diff
@@ -77,12 +83,19 @@ export class OidcStrategy extends PassportStrategy(Strategy, "oidc") {
...
+    if (userProfile.groups && userProfile.groups.length > 0) {
+      userProfile.accessGroups = [...accessGroups, ...userProfile.groups];
+    } else {
+      userProfile.accessGroups = accessGroups;
+    }
+    delete userProfile.groups;
+
     const userFilter: FilterQuery<UserDocument> = {
       $or: [
-        { username: `oidc.${userProfile.username}` },
```
bpedersen2, line 85: this change requires data migrations for existing installations or careful adoption of the username configuration as the username gets prefixed with `oidc.` and you remove that prefix.  -> @PSI

We would probably need to augment this even further as `5.7.  Claim Stability and Uniqueness` in [1] states that only the combination of `iss` and `sub` is guaranteed to be unique, so these fields should be part of the userprofile. 

[1] https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability

bpedersen2 (changes requested):

src/auth/strategies/oidc.strategy.ts
nitrosx: Can we merge #726 in this one and refactor the code?

src/auth/strategies/oidc.strategy.ts
nitrosx: If oidc has the option to provide the thumbnail, I would implement the functionality like you have it now.
The only request that I have is to make it backward compatible: no configuration, no thumbnail.

src/auth/strategies/oidc.strategy.ts
Junjiequan: If you check `migrations/20230309125203-update-user-entries.js` we have migration script to replace the prefixes `oidc.` and `ldap.` to empty string long ago. But have never adapted source code to the migration.
The Uniqueness is a good point, thanks for sharing, maybe we can create another issue to address it? Currently, for OIDC users we are using OR condition with `username` and `email` to get the user which seems like a critical issue needs to be fixed

nitrosx: @bpedersen2 how about we keep username configurable, and we provide the configuration for backward compatibility.
Regarding the line that you highlighted above, we should again go the configuration route, as the filtering might change from institution to institution. Also here we can provide the configuration for backward compatibility.

nitrosx: Upon reading the material, should we set the default configuration for username as `iss` + `sub` ?
Although through configuration and environmental variable it can be changed

src/auth/strategies/oidc.strategy.ts
Junjiequan: The backward compatibility is done through the getUserPhoto function
```

getUserPhoto(thumbnailPhoto: string) {
  return thumbnailPhoto
    ? "data:image/jpeg;base64," +
        Buffer.from(thumbnailPhoto, "binary").toString("base64")
    : "no photo";
}
```

Junjiequan:
https://github.com/SciCatProject/scicat-backend-next/pull/726 raised by @bpedersen2 is merged into this PR to test access-group-provider implementation

src/auth/access-group-provider/access-group-from-graphql-api-call.service.ts
```diff
@@ -27,14 +26,19 @@ export class AccessGroupFromGraphQLApiService extends AccessGroupService {
...
-        accessGroups.join(","),
-    );
-    return accessGroups;
+      Logger.log(
+        "Access groups from graphql api call getAccessGroups: " +
+          accessGroups.join(","),
+      );
+      return accessGroups;
+    } catch (e) {
+      Logger.error("GraphQL", e);
```
nitrosx: I would switch `GraphQL` with `Access Groups from GraphQL`

src/auth/access-group-provider/access-group-from-multiple-providers.service.ts
```diff
@@ -16,7 +16,7 @@ export class AccessGroupFromMultipleProvidersService extends AccessGroupService
     userPayload: UserPayload,
   ): Promise<string[]> {
     const accessGroups: string[] = [];
-
+    Logger.log("get groups from multiple providers");
```
nitrosx: Can we use the SciCat Logger?

src/auth/access-group-provider/access-group-from-payload.service.ts
```diff
@@ -17,8 +17,9 @@ export class AccessGroupFromPayloadService extends AccessGroupService {
     //const defaultAccessGroups: string[] = [];
     let accessGroups: string[] = [];
 
-    const accessGroupsProperty = userPayload?.accessGroupProperty;
-
+    const accessGroupsProperty = this.configService.get(
+      "oidc.accessGroupProperty",
+    );
```
nitrosx: In my understanding, this function is not specific for retrieving groups from oidc.
If that is the case, I would not hard code `accessGroupsProperty` to `oidc.accessGroupProperty`, but I would leave it as input  parameter to the function

src/auth/strategies/oidc.strategy.ts
```diff
@@ -131,34 +148,111 @@ export class OidcStrategy extends PassportStrategy(Strategy, "oidc") {
...
+              { email: userProfile.email },
+            ],
+          }
+        : {
+            [operator]: filter,
+          };
 
-    return profile;
+    return userFilter;
   }
```
nitrosx: I'm not sure I understand what this function does

src/config/configuration.ts
```diff
@@ -118,6 +136,23 @@ const configuration = () => {
...
+      userInfoMapping: {
+        id: process.env.OIDC_USERINFO_MAPPING_FIELD_ID,
+        username:
+          oidcUsernameFieldMapping.split(",").map((v) => v.trim()) ?? [],
+        displayName: process.env.OIDC_USERINFO_MAPPING_FIELD_DISPLAYNAME,
+        familyName: process.env.OIDC_USERINFO_MAPPING_FIELD_FAMILYNAME,
+        emails: process.env.OIDC_USERINFO_MAPPING_FIELD_EMAILS,
+        email: process.env.OIDC_USERINFO_MAPPING_FIELD_EMAIL,
+        thumbnailPhoto: process.env.OIDC_USERINFO_MAPPING_FIELD_THUMBNAIL_PHOTO,
+        groups: process.env.OIDC_USERINFO_MAPPING_FIELD_GROUPS,
```
nitrosx: Do we need mapping for groups?
If so, does it duplicate the functionality that accessGroupProperty has?

nitrosx (changes requested):
@Junjiequan I'm not sure that this PR is ready for production. I think we need to do another round of brainstorming and check that the code is correct.
Thoughts?

src/auth/access-group-provider/access-group-from-multiple-providers.service.ts
Junjiequan: We are using SciCat Logger.
If you check main.ts you can see `app.useLogger(scicatLogger)`, a nestJS way of handling logger instances.
To use scicatLogger you just need to import { Logger } from '@nestjs/common

src/auth/access-group-provider/access-group-from-payload.service.ts
Junjiequan: Yes, you are right, instead of getting the value again from configService we can directly use userPayload.accessGroupProperty.

src/config/configuration.ts
Junjiequan: 1. Yes, we need mapping for groups, because the property name of groups can be changed from IAM.
2. Yes, if we only consider ESS use case, it's duplicated.  `accessGroupProperty` is used for `access-group-from-payload`, where different users may have different preferences for property name of the groups. Especially in the scenario where a user have different types of groups for different purposes.

Junjiequan: 1. Yes, we need mapping for groups because the property name of groups can be changed in IAM.
2. Yes, if we consider only the ESS use case, it is duplicated.`accessGroupProperty` is used for access-group-from-payload to set accessGroups from oidc response, wheras `OIDC_USERINFO_MAPPING_FIELD_GROUPS` is used in userProfile transformation function. In scenarios where a user has different types of groups for different purposes, it might be helpful to separate the value from each other .

src/auth/access-group-provider/access-group-from-graphql-api-call.service.ts
```diff
@@ -27,14 +26,19 @@ export class AccessGroupFromGraphQLApiService extends AccessGroupService {
...
+    try {
+      const response = await this.callGraphQLApi(query);
+      const accessGroups = this.responseProcessor(response);
 
-    Logger.log(
-      "Access groups from graphql api call getAccessGroups: " +
-        accessGroups.join(","),
-    );
-    return accessGroups;
+      Logger.log(
```
nitrosx, line 33: Which logger is used here?

src/auth/access-group-provider/access-group-service-factory.ts
```diff
@@ -1,17 +1,79 @@
...
+      "accessGroupsGraphQlConfig",
     );
-    return new AccessGroupFromStaticValuesService(accessGroupsStaticValues);
+    const accessGroupsOIDCPayloadConfig = configService.get(
+      "accessGroupsOIDCPayloadConfig",
+    );
+
+    const accessGroupServices: AccessGroupService[] = [];
+    if (accessGroupsStaticConfig?.enabled == true) {
+      Logger.log(
```
nitrosx, line 30: Change it to SciCat Logger

nitrosx (changes requested):
Make sure to use SciCat Logger and not directly nestjs one

src/auth/access-group-provider/access-group-from-graphql-api-call.service.ts
nitrosx: The logger is automatically the one that we set when the application start.
Jay clarify it in an in-person Q&A

nitrosx (approved):

src/auth/strategies/oidc.strategy.ts
Junjiequan: It's refactored and should be more readable now.
This is a function that transforms env variables into user query filter
`let user = await this.usersService.findOne(userFilter);`

bpedersen2 (commented):
Looks good to me,  just docs probably need updates

nitrosx:
I confirm that @Junjiequan updated the documentation. Ready to be merged

README.md
```diff
@@ -110,6 +116,20 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
 - `OIDC_SCOPE` [string] _Optional_ Space separated list of the info returned by the oidc service. Example: "openid profile email"
 - `OIDC_SUCCESS_URL` [string] _Optional_ SciCat Frontend auth-callback URL. Required in order to pass user credentials to SciCat Frontend after OIDC login. Example: https://myscicatfrontend/auth-callback
 - `OIDC_ACCESS_GROUPS` [string] _Optional_ Functionality is still unclear.
+- `OIDC_ACCESS_GROUPS_PROPERTY` [string] _Optional_ Target field to get the access groups value from OIDC response.
+- `OIDC_USERINFO_MAPPING_FIELD_USERNAME` [string] _Optional_ comma-separated list. Specifies the fields from the OIDC response to concatenate and use as the user's profile username. For example, setting `OIDC_USERINFO_MAPPING_FIELD_USERNAME="iss, sub"` combines the iss (issuer) and sub (subject) values from the OIDC response, resulting in a username like `myIssuer_myUserName`. This allows for customizable username definitions based on OIDC response attributes.
```
bpedersen2: This field is mentioned 2 times here  (see line 124)

README.md
```diff
@@ -110,6 +116,20 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
...
+- `OIDC_USERINFO_MAPPING_FIELD_DISPLAYNAME` [string] _Optional_ Specifies the fields from the OIDC response and use as the user's profile displayname. For example, setting `OIDC_USERINFO_MAPPING_FIELD_DISPLAYNAME="preferred_username"` use displayName value from the OIDC response, resulting in a displayname like `myPreferredName`. This allows for customizable displayname definitions based on OIDC response attributes.
+- `OIDC_USERINFO_MAPPING_FIELD_EMAIL` [string] _Optional_ Same as `OIDC_USERINFO_MAPPING_FIELD_DISPLAYNAME`. Defaults to "email"
+- `OIDC_USERINFO_MAPPING_FIELD_DISPLAYNAME` [string] _Optional_ Same as `OIDC_USERINFO_MAPPING_FIELD_DISPLAYNAME`. Defaults to "name"
+- `OIDC_USERINFO_MAPPING_FIELD_USERNAME` [string] _Optional_ Same as `OIDC_USERINFO_MAPPING_FIELD_DISPLAYNAME`. Defaults to "preferred_username" || "name"
```
bpedersen2: duplicated, and looking at the code the first one is correct.

bpedersen2 (approved):
Except for a minor doc glitch looks fine

PR #1080, opened 2024-02-26
839 create job

sbliven:
## Description

Implementation of the `POST /job` endpoint.

This should now compile, but testing is ongoing.


src/casl/authop.enum.ts
```diff
@@ -76,15 +76,14 @@ export enum Action {
...
   // -------------
   // Origdatablock
-  // endpoint authorization actions
+  // endpoint authorization
   OrigdatablockCreate = "origdatablock_create",
   OrigdatablockRead = "origdatablock_read",
   OrigdatablockUpdate = "origdatablock_update",
   OrigdatablockDelete = "origdatablock_delete",
-  // individual actions
+  // individual
```
nitrosx: can we change it to the default naming `data instance authorization`?

src/casl/authop.enum.ts
```diff
@@ -110,8 +109,7 @@ export enum Action {
   ProposalsAttachmentUpdate = "proposals_attachment_update",
   ProposalsAttachmentDelete = "proposals_attachment_delete",
   ProposalsDatasetRead = "proposals_dataset_read",
-
-  // individual actions
+  // individual
```
nitrosx: Same comment as on line 86

src/casl/authop.enum.ts
```diff
@@ -147,8 +145,7 @@ export enum Action {
   SamplesAttachmentUpdate = "samples_attachment_update",
   SamplesAttachmentDelete = "samples_attachment_delete",
   SamplesDatasetRead = "samples_dataset_read",
-
-  // individual actions
+  // individual
```
nitrosx: Same comment as on line 86

src/jobs/config/jobConfig.createExample.json
```diff
@@ -0,0 +1,28 @@
...
+        },
+        {
+          "actionType": "email",
+          "mailer": {
+            "host": "smtp.forwardemail.net",
+            "port": 465,
+            "secure": true,
+            "auth": {
+              "user": "REPLACE-WITH-YOUR-ALIAS@forwardemail.net",
+              "pass": "REPLACE-WITH-YOUR-GENERATED-PASSWORD"
```
nitrosx, line 17: I wonder if this is the right thing to do?
Do we want to encourage admins to have a clear text password here?

src/jobs/types/jobs-config-schema.enum.ts
nitrosx, line 1: What is the purpose of this definition?

nitrosx (changes requested):
I left some comments, although we could go ahead and merge in to release-jobs and make the changes there.

despadam:
Changes included in this PR:
- Renaming `Action enum` to `AuthOp enum` for a more representative name, as the file does not contain the names of what we now think of as actions and was easily creating confusion. As a result every import and occurance of the enum has been updated.
- For better organization, all files related to jobs have been moved under the `jobs` folder, which now has the following structure:
    - jobs
        - actions
        -  config
        - dto
        - schemas
        - types
- 2 dummy actions for jobs have been created, namely `emailaction` and `logaction`
    - `email` for now validates some basic fields and fills in a temp template, but is probably not complete and the mailer has not been tested yet
    - `log` requires no validation - is only used for logging 


!!! All following changes refer to Job Creation for now. Update has not yet been developed.
!!! Across the code you will find some `TODO` comments. These indicate parts that are still unclear / to be discussed.

- `jobs.controller.ts` has been updated to correspond to the new `CreateJobDto`
    `create` does the following:
    - validates the job and its actions
    - performs `instanceAuthorization` to check if the user is allowed to create a job
        - see `jobs-auth.enum.ts` for the valid options
    - creates a new job in the database, after adding the `status` and `createdBy` fields
    - performs the actions of the created job

- `configuration.ts` validates `jobConfig`
    - see `jobConfig.createExample.json` for a dummy config
    - The json schema is validated against `jobConfig.schema.ts`
        - see `jobs-config-schema.enum.ts` for the valid schema fields
        - The way validation is currently being done translates to: A job should have a `jobType` and optionally an `auth` field. If `auth` is not provided, then the default is `authenticated`. If actions are provided for this job, then each of them should at least have an `actionType`.

src/jobs/types/jobs-config-schema.enum.ts
despadam: This one is used to avoid hardcoded strings wherever possible when performing the config schema validation. For now it only contains some basic fields.

src/jobs/config/jobConfig.createExample.json
despadam: This is up for discussion. It is currently just an example of what the config of an email action could look like, but we can very easily change it.

sbliven: What's the correct way to do this? Is it any different than having cleartext in `.env`?

nitrosx: it depends from the target system. 
I would strongly discourage to have this syntax in an example in core, as mindless users might mistakenly leave their password in files

nitrosx: Let's table the issue for now and review it later

src/jobs/types/jobs-config-schema.enum.ts
sbliven: We discussed whether the schema could be generated directly from the classes (or vice versa). That would probably make this enum unneeded. For now it's just a way to collect magic strings.

PR #1093, opened 2024-03-01
URL action

sbliven:
Initial URLAction implementation.

The test configuration simply calls /health and logs the result.


src/config/configuration.ts
```diff
@@ -46,7 +47,9 @@ const configuration = () => {
 
   // Register built-in job actions
   registerDefaultActions();
-  const job_configs: Promise<JobConfig[]> = loadJobConfig("src/jobs/config/jobConfig.createExample.json");
+  const job_configs: Promise<JobConfig[]> = loadJobConfig(
+    "src/jobs/config/jobConfig.createExample.json",
```
nitrosx: We should brainstorm about how to remove the hardcoding and make it more site configurable.

nitrosx (approved):
Just left a comment as reminder to discuss the issue when the functionalities are stable

src/config/configuration.ts
sbliven: Yes, there's a lot of little things like that that will be quick to implement but need to be handled before release.

src/jobs/actions/urlaction.ts
```diff
@@ -0,0 +1,91 @@
...
+  allowedProtoProperties: {
+    id: true,
+    type: true,
+    statusCode: true,
+    statusMessage: true,
+    messageSent: true,
+    createdBy: true,
+    jobParams: false,
+  },
+  allowProtoPropertiesByDefault: false, // limit accessible fields for security
```
sbliven, line 23: I went with a more restrictive setting for now while trying to understand the security. See #1113 for details.

This approach is harder to maintain since the templatable properties need to be kept up to date with the JobClass schema.

src/jobs/actions/urlaction.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Respond to Job events by making an HTTP call.
+ */
+export class URLAction<T> implements JobAction<T> {
+  public static readonly actionType = "url";
+
+  private urlTemplate: Handlebars.TemplateDelegate<JobClass>;
+  private method = "GET";
+  private headers: Record<string, string> = {};
+  private body: Record<string, any> | null = null;
```
sbliven, line 35: Body probably also needs to accept templates, but how to deal with it in JSON input? Maybe only apply templating to string values?

src/jobs/actions/urlaction.ts
```diff
@@ -0,0 +1,91 @@
...
+      throw new HttpException(
+        {
+          status: response.status,
+          message: `Got response: ${await response.text()}`,
+        },
+        response.status,
+      );
+    }
+
+    // TODO do something with the response?
```
sbliven, line 64: Should the response body be used to change the jobstatus?

src/config/configuration.ts
sbliven: #1114

sbliven:
This is done for now. I'm going to merge into release-jobs and make any changes there.

Issue #838, opened 2023-10-25
Jobs migration: Complete statusUpdate


ddijken:
Complete the statusUpdate function in the jobs.controller

despadam:
- [x] https://github.com/SciCatProject/scicat-backend-next/pull/1110
- [x] https://github.com/SciCatProject/scicat-backend-next/pull/1133#pullrequestreview-1957815317

Issue #626, opened 2023-07-24
Jobs migration: tests


nitrosx:
## Brainstorming
Add in the comments ideas for tests

rkweehinzmann:
Brainstorming what could and should be tested to make sure jobs functionality runs smoothly.

A) if the user's input is not what it should be, what does scicay do? are potential errors caught?
B) if the user does everything correct, does scicat do it's job correctly?

Concretely, I would test these two categories (authentication/authorization and system communication):

Test
- if user put in a valid email address

- if there is a timeout if the user takes very long time to submit the job, imagine a user creates the job, breaks for a day or two..  
(longer than default token of 1hr if not anonymous). What do we want as functionality? User can continue as long as the system is up? In automated cases this won't be an issue I assume.

- if one of the selected datafiles is by mistake chosen, ie he has no permission to do smthg with it 
(here would it be nice, if one sees an overview before job submission to check his input data)

- if potential memory leak occurs (in case a lot of datasets with many associated files are requested and smthg goes wrong)

- if the statusUpdate works for 10, 100, 1000 datasets


Probably to be refined when code is more ready.

rkweehinzmann:
After yesterdays discussion, this is what I recall to start with:

1) Automatic schema testing: Is this schema that I'm passing correct?
2) Given a user with a given jobs configuration, test if a job can be created or not, by either giving wrong credentials or correct credentials. Authenication is tested at 2 levels: Firstly, endpoint authorisation. Secondly, on the datasets (from JobParams from jobs config).

despadam:
@rkweehinzmann Yes, that is correct

sbliven:
How do we test the actions? These usually require interaction with external services.

sofyalaski:
https://github.com/SciCatProject/scicat-backend-next/blob/caf3eea1a3d91b0e8c11504d7a3fa8f1c83b2de6/src/jobs/jobs.controller.ts#L450-L474

I think this `else if` statement lacks a `return true;`  at the end.

sofyalaski:
1. When testing for normal users (not in the CREATE_JOB_GROUPS) I always got the forbidden error coming from no endpoint access:
https://github.com/SciCatProject/scicat-backend-next/blob/b5661d6b9528744267612a7001cb64a02ccde24c/src/casl/casl-ability.factory.ts#L896-L899

The issue was that this if condition always evaluates to false because there seems to be a type disagreement. 
I changed that to 
[...]

PR #1110, opened 2024-03-13
Job statusUpdate

despadam:
### Fixes:
#838  
The code as is now, compiles and runs: is able to create and update a job in the db.

### Questions:
Please see comments below


src/jobs/jobs.controller.ts
```diff
@@ -475,18 +476,98 @@ export class JobsController {
...
+  )
+  @Patch(":id")
+  @ApiOperation({
+    summary: "It updates an existing job.",
+    description: "It updates an existing job.",
+  })
+  @ApiBody({
+    description: "Input fields for the job to be updated",
+    required: true,
+    // schema: new UpdateJobStatusDto, // TBD
```
despadam: This part is exactly the same as in `create()`. However, in `update()` it produces the following syntax error (hence why the line is commented out):
`Type 'UpdateJobStatusDto' is not assignable to type 'SchemaObject | ReferenceOb-ject'.`

What is the difference between the two DTOs? Why `UpdateJobStatusDto` is not accepted in the `schema` here?

src/jobs/jobs.controller.ts
```diff
@@ -475,18 +476,98 @@ export class JobsController {
...
+
+
+  /**
+   * Find job by id
+   */
+  @UseGuards(PoliciesGuard)
+  @CheckPolicies((ability: AppAbility) =>
+    ability.can(AuthOp.Read, JobClass),
+  )
+  @Get(":id")
```
despadam, line 556: **All other endpoints** apart from `create()` and `update()`, have the following error. I log in and authorize in swagger as admin, however it seems that an admin does not have the rights to perform a GET or DELETE.
`{
  "statusCode": 403,
  "message": "Forbidden resource",
  "error": "Forbidden"
}`

src/jobs/jobs.controller.ts
```diff
@@ -475,18 +476,98 @@ export class JobsController {
...
+          status: HttpStatus.BAD_REQUEST,
+          message: "Invalid job id.",
+        },
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+    // Check job authorization
+    await this.instanceAuthorization(currentJob.jobParams, request.user as JWTUser);
+    // Update job in database
+    const updatedJob = await this.jobsService.statusUpdate(id, updateJobDto);
```
despadam, line 537: Currently `update()` only performs a **status update** (only `statusMessage`, not `statusCode`). 
1. Is there a need to update more fields? If yes, which ones, why and when? 
2. Do we need to perform an update after creation is complete? And is it just a status update? If yes, what would the new job status be? If no, which other fields do we also need to update?
3. Which are the acceptable values for `statusMessage`? What are their corresponding `statusCode` values?

src/jobs/schemas/job.schema.ts
```diff
@@ -107,7 +107,7 @@ export class JobClass extends OwnableClass {
     type: Object,
     required: false,
   })
-  configuration: Record<string, unknown>;
+  configuration: Record<string, unknown>;  // TBD
```
despadam, line 110: In the schema we store `jobParams`, so that is later accessible when performing an update.
1. Should we also store the `configuration`, and if yes why?
2. Should job `actions` be also somehow stored in the db? Yes/no and why?

src/jobs/config/jobConfig.example.json
```diff
@@ -0,0 +1,24 @@
...
+          "actionType": "log"
+        }
+      ]
+    },
+    "update": {
+      "auth": "#all"
+    }
+  },
+  {
+    "jobType": "public",
```
despadam, line 16: The current implementation removes the auth check for public datasets from the controller and moves it to the config file.

Is it safe to assume/request that auth options for public jobs will be always provided in this way?

src/jobs/jobs.controller.ts
```diff
@@ -369,21 +356,21 @@ export class JobsController {
...
         where: {
-          pid: { in: datasetIds },
+          pid: { $in: datasetIds },
           $or: [
-            { ownerGroup: { in: user.currentGroups } },
-            { accessGroups: { in: user.currentGroups } },
-            { isPublished: true }  // TODO
+            { ownerGroup: { $in: user.currentGroups } },
+            { accessGroups: { $in: user.currentGroups } },
+            { isPublished: true }  // TBD
```
despadam, line 368: Does `{ isPublished: true }` mean `isPublic`?

src/jobs/jobs.controller.ts
```diff
@@ -369,21 +356,21 @@ export class JobsController {
...
           ]     
         },
       });
 
-      const datasetsNoAccess = datasetIds.length - numberOfDatasets.count;
+      const datasetsNoAccess = datasetIds.length - numberOfDatasets.count; // TBD
```
despadam, line 373: The current implementation to check access to datasets was copied from the previous version.

Why was it written like this? Is it sufficient or should it change?

src/jobs/jobs.controller.ts
```diff
@@ -335,23 +318,27 @@ export class JobsController {
   /**
    * Checking if user is allowed to create job according to auth field of job configuration
    */
+  // TBD
   async instanceAuthorization(
-    jobCreateDto: CreateJobDto,
+    jobParams: Record<string, any> | undefined,
```
despadam, line 323: `instanceAuthorization` is currently implemented to accomodate both `create` and `update`.

What about get, delete etc? What would be the auth options for these? Would the same function be sufficient or should it be defined differently for each?

> TODO: Update needs an additional specific user that we set in configuration.
We might want to create a user with a specific jwt token dedicated for the specific job update.  I see this a update segregation. ---> @nitrosx 's comment via Slack

src/jobs/jobs.controller.ts
```diff
@@ -335,23 +318,27 @@ export class JobsController {
...
     user: JWTUser,
   ): Promise<boolean> {
     // Accepted options:
     // #all, #datasetOwner, #datasetOwnerOrAccess, #authenticated,
 
-    if (jobCreateDto.jobParams.auth != JobsAuth.All) {
+    if (jobParams === undefined) {
```
despadam, line 329: `jobParams` is currently optional. If it is not provided then what would be the default user auth? The current implementation allows for job creation and update when `jobParams` is undefined.

When `jobParams` is provided, we assume that it contains both `auth` and `datasetIds` fields, otherwise we throw an exception. Is some extra validation needed or is this enough?

src/jobs/jobs.controller.ts
sbliven: I looked into this issue. I think we are actually using this wrong everywhere, but coincidentally it compiles everywhere but here.

The ApiBody is expecting a JSON Schema description. The type definition is in `@nestjs/swagger/dist/interfaces/open-api-spec.interface.d.ts`. All fields are optional. Typescript enforces that the class passed have at least one of these properties and ignores additional properties. In the case of `CreateJobDto` (as well as `CreateDatasetDto`) there is a `type: string` parameter that happens to be compatible with the expected type. However, I guess that DTO validation isn't actually working in BE-next, and we just didn't notice until now.

sbliven: I think I solved this. `@ApiBody` should use the `type` param rather than `schema`. I'll push a commit on this branch, but we should test server-side validation.

src/jobs/jobs.controller.ts
nitrosx: There is no DELETE on the job.
I would need to check on the GET.

src/jobs/jobs.controller.ts
nitrosx: Status update should be able to update the status message and status code.

Every time the system receives an update request, it should update the statusMessage and statusCode fields, together with append them to the update history.

Job updates will be performed until the job is completed successfully or not. 
The only thing that we can do is status update, we are not allowed to change anything else.

statusMessages are just arbitrary strings. statusCode should be defined in the job definition... at least that was my idea.

src/jobs/config/jobConfig.example.json
nitrosx: YES!!!

src/jobs/jobs.controller.ts
nitrosx: The only other operation that we should allow is GET.
We should  brainstorm and come up with the permission matrix.

src/jobs/jobs.controller.ts
nitrosx: It is reasonable to assume that jobParams is always present.
I thought that the validation rules for jobParams were defined in the job definition.

src/jobs/jobs.controller.ts
nitrosx: Correct

src/jobs/jobs.controller.ts
nitrosx: In one of our discussions, we agreed that the job creation will succeed if the user has access to all the requested datasets.
I would keep it this way for now

src/jobs/schemas/job.schema.ts
nitrosx: If space and effort allows, I would store both of them.
I would do that for validation purposes. The job definition can change, so it would be nice to know what was the definition when the job was created.

nitrosx (approved):
Let's merge the PR and start testing more extensive.
Maybe we should start writing API tests.

src/jobs/jobs.controller.ts
despadam: Use params from config, not the DTO.

src/jobs/schemas/job.schema.ts
despadam: 1. Configuration should not be stored in the db. We can keep track of the config version (i.e. hash) and if config gets  updated, we stop all running jobs and restart.
2. Also actions should not be in the db, only the config file.

src/jobs/jobs.controller.ts
despadam: `instanceAuthorization` is different for `create` vs `update`.
For `create`, after endpoint auth we need to check if the user has access to the datasets.
For `update`, after endpoint auth we need to check whether the user has the right to update this specific job. To be covered by @nitrosx in the endpoint authorization.

src/jobs/jobs.controller.ts
despadam: - [x] We will keep: `create`, `update`, `get`, `get by id` and `delete`
- [x] We will remove `fullquery` and `fullfacet`
- [x] `get`, `get by id` and `delete` to be checked by @nitrosx for the `403 error` : the current user rights are probably not defined well
- [ ] For `delete`, we need to restrict user access

src/jobs/jobs.controller.ts
despadam: Testing to be covered by @rkweehinzmann, as mentioned in #626

src/jobs/jobs.controller.ts
despadam: - `update()` should perform a status update (`statusMessage` and `statusCode`)
- We will not impose any type-checking.
- After a successful creation, a job gets updated with these values: `{ "statusMessage": "Job has been created", "statusCode": "jobCreated" }`

nitrosx: Perfect!!!

PR #1111, opened 2024-03-15
Improve local development experience: provide default mongodb url and jwt token

Ingvord:
## Description

Add a few default values for local development

## Motivation

Simplify local development setup

## Fixes:

Default values for mongodb and jwt:
`mongodb://localhost:27017/dacat` and `xxx`

## Changes:

See diff

N/A

### DIFF
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -114,7 +114,7 @@ const configuration = () => {
     httpMaxRedirects: process.env.HTTP_MAX_REDIRECTS ?? 5,
     httpTimeOut: process.env.HTTP_TIMEOUT ?? 5000,
     jwt: {
-      secret: process.env.JWT_SECRET,
+      secret: process.env.JWT_SECRET ?? "xxx",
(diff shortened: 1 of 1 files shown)


src/config/configuration.ts
```diff
@@ -114,7 +114,7 @@ const configuration = () => {
     httpMaxRedirects: process.env.HTTP_MAX_REDIRECTS ?? 5,
     httpTimeOut: process.env.HTTP_TIMEOUT ?? 5000,
     jwt: {
-      secret: process.env.JWT_SECRET,
+      secret: process.env.JWT_SECRET ?? "xxx",
```
bpedersen2, line 117: I don't like this, as it introduces an unsafe default just  for developer convenience. 

This should be placed in a dev-only config.

bpedersen2 (changes requested):

sbliven:
I agree with @bpedersen2, these should be implemented with a dev .env file.

We had some discussions about having dev configurations as part of scicatlive, but they shouldn't be in the main config.

Maybe what you want is better log messages if these are undefined?

PR #1129, opened 2024-03-20
chore: update documentation and .env.example

Junjiequan:
## Description

- updated README documentation for loggers configuration. 
- updated .env.example according to the recent configuration changes.


README.md
```diff
@@ -77,6 +79,12 @@ automatically create the specified accounts on startup. If this file is not prov
 Follow the structure of [functionalAccounts.json.minimal.example](/functionalAccounts.json.minimal.example) to create
 your own _functionalAccounts.json_ file.
 
+### Loggers configuration
+
+Providing a file called loggers.json\_ at the root of the project, locally or in the container, and create an external logger class in the `src/loggers/loggingProviders/`directory will automatically create specified one or multiple loggers instances.
```
sbliven: Slight typo: underscore after `loggers.json`

.env.example
```diff
@@ -34,7 +34,48 @@ SMTP_HOST=<SMTP_HOST>
...
+MEM_LIMIT="4G"
+MONGODB_COLLECTION="Dataset"
+ES_MAX_RESULT=100000
+ES_FIELDS_LIMIT=400000
+ES_INDEX="dataset"
+ES_PORT=9200
+ES_HOST="https://localhost:9200"
+ES_USERNAME="elastic"
+ES_PASSWORD="duo-password"
+ES_REFRESH=<"wait_for"|"false">
```
sbliven, line 79: Should this be a boolean? It's a bit weird to have "false" as an enum value.

sbliven (approved):
This is very timely! I was just trying to work out how to configure logging.

.env.example
Junjiequan: Yes..It's indeed looking weird. Somehow we were using `yes|no` patterns for enable/disable flags before and I followed the pattern for consistency. Now we mixed `yes|no` and `true|false` for flags.
I will create an issue regarding the inconsistency, maybe we should replace all of the `yes|no` to `true|false`

PR #1132, opened 2024-03-22
history.schema.ts: drop default and required on managed fields

bpedersen2:
These fields don't need a default and should never be requried, as they are managed by mongoose (see l. 12)

- [ ] Docs updated?


Junjiequan (approved):

nitrosx (approved):
It looks fine to me. Let's merge and see if history elements still have a consistent dates fields.
We might want to work on a PR with new tests for history

PR #1127, opened 2024-03-19
[WIP] Implement job authorization

sbliven:
Update endpoint and instance authorization for the job controller.

This continues work on #620.

Issue #1119, opened 2024-03-18
Add job configuration version


sbliven:
Rather than storing the full configuration for each job in the database, allow a version string to be set in the config file.

Questions:
- Should the version default to a hash of the configuration?
- Should version be specific to each job type?

nitrosx:
The version is specific for each job definition.

sbliven:
Included in #1133

PR #1133, opened 2024-03-22
Jobs: statusHistory, instance auth based on config, Get functions

despadam:
Based on the comments from PR #1110 : 
- added `statusHistory` = every time we perform a `statusUpdate`, we store the new `statusCode` in an array, along with the timestamp of the update 
- `jobSchema` should **not** contain `configuration` and `jobParams` - we don't need to store it in the db
- `instanceAuthorization` now reads the `auth` field from the `configuration` instead of `jobParams`
- `endpoint auth` was fixed for GET with inspiration from #1127 (thanks @sbliven)
- updated swagger decoratos for `Get jobs` query and added a check for the filters
- `fullfacet` and `fullquery` were removed

Also covers:
- #1119 (TBD in case I missed something)


despadam:
Missing:
- Endpoint authorization for DELETE (see #1127)
- Endpoint authorization for UPDATE: Currently `instanceAuthorization` is used for both `create` and `update`, but it **should be different**. For `update`, we need to check whether the user has the right to update this specific job, and it should be covered in the endpoint authorization.

src/jobs/jobs.controller.ts
```diff
@@ -547,76 +629,94 @@ export class JobsController {
...
+    summary: "It returns the requested job.",
+    description: "It returns the requested job.",
   })
   @ApiResponse({
     status: HttpStatus.OK,
     type: JobClass,
     description: "Found job",
   })
   async findOne(
-    @Req() request: Request,
```
nitrosx: You should keep the request.
We use this to retrieve the user and decide if he/she has access to the job

src/jobs/schemas/job.schema.ts
```diff
@@ -83,45 +83,32 @@ export class JobClass extends OwnableClass {
...
-    required: false,
-    description:
-      "This is the equivalent object of the jobs parameters provided by the user.",
-  })
-  @Prop({
-    type: Object,
-    required: false,
-  })
-  jobParams: Record<string, unknown>;
-  // TODO email address for owner from scicat? see create example for job.recipients
```
nitrosx: We should keep this section because it is the only way to know which parameters the user sent with the create request.
...unless we have them somewhere else?

nitrosx (changes requested):
@despadam please check my comments and let me know your thoughts?

src/jobs/schemas/job.schema.ts
despadam: True, this information is not stored anywhere else. I restored it

src/jobs/config/jobConfig.example.json
```diff
@@ -1,6 +1,7 @@
 [
   {
     "jobType": "archive",
+    "configVersion": "v1.0 2024-03-01 6f3f38",
```
sbliven, line 4: Do we want a less complicated example version? Or keep it deliberately unrealistic like this?

src/jobs/config/jobConfig.schema.ts
```diff
@@ -44,7 +46,7 @@ export const JobConfigSchema = {
         additionalProperties: true
       },
     },
-    required: ["jobType"],
+    required: ["jobType", "configVersion"],
```
sbliven, line 49: I suggest configVersion be optional

src/jobs/config/jobConfig.schema.ts
```diff
@@ -1,12 +1,14 @@
 import { JobsAuth } from "../types/jobs-auth.enum";
 
 
+// We use this to validate the Json schema of the provided configuration
 export const JobConfigSchema = {
```
sbliven, line 5: Can the schema be generated with `SchemaFactory.createForClass(JobConfig[])`?

src/jobs/dto/update-jobstatus.dto.ts
```diff
@@ -3,14 +3,23 @@ import { IsOptional, IsString } from "class-validator";
...
   @ApiProperty({
     type: String,
     required: true,
     description: "Updated job status code for the current status.",
   })
   @IsString()
-  readonly jobStatus: string;
+  readonly jobStatusCode: string;
 
+  // TBD are 'message' and 'token' needed?
```
sbliven, line 22: Message, yes. It's the human-readable interpretation of the code. Maybe it could be optional in the DTO (but including it should be best practice).

@nitrosx Are you using token? Shouldn't authorization be done in the `Bearer` header? (This can be a separate issue, not necessarily part of this PR)

src/jobs/jobs.controller.ts
```diff
@@ -571,52 +653,73 @@ export class JobsController {
...
+  async findAll(
+    @Req() request: Request,
+    @Query("filter") filter?: string,
+  ): Promise<JobClass[] | null> {
     const parsedFilter: IFilters<
       JobDocument,
       FilterQuery<JobDocument>
     > = JSON.parse(filter ?? "{}");
+
+    if (!this.isFilterValid(Object.keys(parsedFilter))) {
```
sbliven, line 696: `isFilterValid` is a pretty limited check (keys only, top level only). Does more validation of the filter happen in the `@Query("filter")` magic?

src/jobs/jobs.controller.ts
```diff
@@ -475,25 +554,26 @@ export class JobsController {
...
     // Perform the action that is specified in the create portion of the job configuration
     await this.performJobCreateAction(createdJobInstance);
-    // TBD
-    // Update job instance with results of job create action
-    // return await this.jobsService.statusUpdate(createdJobInstance);
-    return createdJobInstance;
+    // Update job instance after the success of job create action
+    return await this.jobsService.statusUpdate(
+      createdJobInstance.id,
+      { "jobStatusMessage": "Job has been created.", "jobStatusCode": "jobCreated" } as UpdateJobStatusDto
```
sbliven: Why do this as a statusUpdate rather than part of the create call? I suspect that statusUpdate could have side effects like emitting events to any registered listeners. This could cause some actions to run twice.

Even without side effects, I think we want the job to have the correct statusCode before `performJobCreateAction`, so this should move up a line.

src/jobs/jobs.controller.ts
```diff
@@ -312,29 +372,75 @@ export class JobsController {
...
+      if (user === null) {
+        throw new HttpException(
+          {
+            status: HttpStatus.UNAUTHORIZED,
+          },
+          HttpStatus.UNAUTHORIZED,
+        );
+      }
+      return true;
+    } else if (jc[functionType].auth == JobsAuth.DatasetPublic) {
```
sbliven, line 424: It's not clear to me why `DatasetPublic` is needed. Isn't it covered by `DatasetAccess` already?

src/jobs/jobs.controller.ts
```diff
@@ -356,16 +462,16 @@ export class JobsController {
...
-    } else if (jobParams.auth == JobsAuth.DatasetAccess) {
+    } else if (jc[functionType].auth == JobsAuth.DatasetAccess) {
       // verify that all the pids listed in the property indicated are accessible by the user
-      const datasetIds = this.checkDatasetIds(jobParams);
+      const datasetIds = this.checkDatasetIds(createJobDto.jobParams);
       const numberOfDatasets = await this.datasetsService.count({
         where: {
           pid: { $in: datasetIds },
           $or: [
             { ownerGroup: { $in: user.currentGroups } },
```
sbliven, line 483: Doesn't this give an error for anonymous users (`user === undefined`)?

src/jobs/jobs.controller.ts
```diff
@@ -386,17 +492,6 @@ export class JobsController {
...
-      if (user === null) {
-        throw new HttpException(
-          {
-            status: HttpStatus.UNAUTHORIZED,
-          },
-          HttpStatus.UNAUTHORIZED,
-        );
-      }
-      return true;
     }
```
sbliven, line 507: If `auth` isn't an enum value then shouldn't it be interpreted as a group name? I forget what we decided on that, but it may need to be changed following this PR.

sbliven (approved):
I've left a few minor comments, but I think they could mostly be pushed to the next PR. Nice work!

src/jobs/dto/update-jobstatus.dto.ts
despadam: There is already `jobStatusMessage`, which is the human-readable interpretation of `jobStatusCode`. Why is `message` also needed?

src/jobs/jobs.controller.ts
despadam: `isFilterValid` was created as a basic syntax check for the GET filters, because I noticed there was none. We can discuss how it can be expanded. There is no additional validation happening.

src/jobs/jobs.controller.ts
despadam: I am removing this part. Now message gets set to `jobCreated` before `performJobCreateAction`

src/jobs/jobs.controller.ts
despadam: I am not sure either. I covered all options provided in `jobs-auth.enum.ts`. We can discuss this again

src/jobs/jobs.controller.ts
despadam: Adding one extra function to verify user is authenticated

src/jobs/jobs.controller.ts
despadam: I also forgot what we decided, let's discuss again

despadam:
I am merging so we can proceed with testing and will address all unresolved comments in a new PR.

src/jobs/jobs.controller.ts
sofyalaski: I am not quite sure I understand this. So there are two other versions GROUP and USER in the documentation that a facility can specify in the jobs configuration. In the current version of the code, there is this a line in `instanceAuthorizationJobCreate` function: `ability.can(AuthOp.JobCreateConfiguration, jobInstance)` which is responsible for that use case. In casl factory it will check if the user that wants to create a job is in `jobCreateInstanceAuthorizationValues`. This one is defined as `[#all, #authenticated, username (of user trying to do a job) and his/her groups]`. So it kind of does what Spencer asks in casl.

src/jobs/jobs.controller.ts
sofyalaski: The case of `DatasetPublic` that is not covered by `DatasetAccess`, is when an anonymous user tries to create a job. At least for ESS it seems to be an important use case. Also because `#all` is probably not going to be used for a real configuration anywhere, as it's too permissive ( at least it says so in the docs about `#all`)

PR #1173, opened 2024-04-12
Jobs authorization

nitrosx:
## Description
This PR implement the jobs authorization model which includes a mix of group authorization like all the other endpoints and individual job type authorization 

## Motivation
After collecting some use cases from multiple facility, we realized that the job authorization needed to be slightly more flexible

## Changes:
* jobs controller
* jobs services
* casl factory

- [X ] Docs updated?


sbliven:
Can this be rebased without the dependabot changes? It's hard to check what's new in this PR

src/casl/casl-ability.factory.ts
```diff
@@ -783,96 +784,175 @@ export class CaslAbilityFactory {
...
-      can(AuthOp.JobsCreateAny, JobClass);
-      can(AuthOp.JobsUpdateAny, JobClass);
-    } else if (
-      user.currentGroups.some((g) =>
-        configuration().createJobGroups.includes(g),
-      )
-    ) {
+      // instance authorization
+      can(AuthOp.JobCreateConfiguration,JobClass, {
+        'configuration.create.auth': '#all',
```
sbliven: We should add magic strings to an enum.

src/jobs/jobs.controller.ts
```diff
@@ -297,150 +311,131 @@ export class JobsController {
...
+            ? jobCreateDto.ownerGroup
+            : jobUser?.currentGroups[0]
+        )
+      } else {
+        jobInstance.ownerUser = user.username;
+        jobInstance.requesterEmail = user.email;
+        jobInstance.ownerGroup = (
+          user.currentGroups.includes(jobCreateDto.ownerGroup)
+            ? jobCreateDto.ownerGroup
+            : user.currentGroups[0]
```
sbliven: Should fail if an invalid group was specified

src/jobs/schemas/job.schema.ts
```diff
@@ -121,6 +132,30 @@ export class JobClass extends OwnableClass {
...
+    type: String,
+    required: false,
+    default: "",
+    description: "If the job is submitted anonymously, an email has to be provided"
+  })
+  @Prop({
+    type: String,
+    required: true,
+  })
+  requesterEmail: string;
```
sbliven: I suggest spelling it `contactEmail` for consistency with Dataset

src/jobs/jobs.controller.ts
```diff
@@ -297,150 +311,131 @@ export class JobsController {
...
+        jobInstance.ownerUser = jobUser?.username as string;
+        jobInstance.requesterEmail = jobUser?.email as string;
+        jobInstance.ownerGroup = (
+          jobCreateDto.ownerGroup
+            ? jobCreateDto.ownerGroup
+            : jobUser?.currentGroups[0]
+        )
+      } else {
+        jobInstance.ownerUser = user.username;
+        jobInstance.requesterEmail = user.email;
```
sbliven: Should throw an error if it differs

src/jobs/jobs.controller.ts
```diff
@@ -297,150 +311,131 @@ export class JobsController {
...
-        );
+
+    // NOTE: We need JobClass instance because casl module works only on that. 
+    // If other fields are needed can be added later.
+    const jobInstance = new JobClass();
+    jobInstance._id = "";
+    jobInstance.ownerUser = "";
+    jobInstance.ownerGroup = "";
+    jobInstance.accessGroups = [];
+    jobInstance.requesterEmail = jobCreateDto.requesterEmail;
```
sbliven: TODO: The EmailJobAction should validate that this is defined

sbliven: We should also document that configuring an email action with `#all` could lead to potential abuse for sending spam

src/jobs/config/jobConfig.example.json
```diff
@@ -16,7 +17,7 @@
       ]
     },
     "update": {
-      "auth": "#all"
+      "auth": "archivemanager"
```
despadam, line 21: Is `archivemanager` a valid auth option? I thought we only allow those that are defined in `src/jobs/types/jobs-auth.enum.ts`

src/casl/authop.enum.ts
```diff
@@ -177,16 +177,20 @@ export enum AuthOp {
...
+  JobStatusUpdate = "job_update",
+  JobDelete = "job_delete",
   // data instance authorization
-  JobsCreateOwner = "jobs_create_owner",
-  JobsCreateAny = "jobs_create_any",
-  JobsReadAccess = "jobs_read_access",
-  JobsReadAny = "jobs_read_any",
-  JobsUpdateAccess = "jobs_update_access",
-  JobsUpdateAny = "jobs_update_any",
+  JobCreateConfiguration = "job_create_configuration",
```
despadam, line 185: In which scenario is this used?

src/casl/authop.enum.ts
```diff
@@ -177,16 +177,20 @@ export enum AuthOp {
...
+  JobCreateOwner = "job_create_owner",
+  JobCreateAny = "job_create_any",
+  JobReadAccess = "job_read_access",
+  JobReadAny = "job_read_any",
+  JobStatusUpdateConfiguration = "job_status_update_configuration",
```
despadam, line 190: In which scenario is this used?

src/casl/casl-ability.factory.ts
```diff
@@ -783,96 +784,175 @@ export class CaslAbilityFactory {
...
-      cannot(AuthOp.JobsCreate, JobClass);
-      cannot(AuthOp.JobsUpdate, JobClass);
-    } else if (
-      user.currentGroups.some((g) => configuration().adminGroups.includes(g))
-    ) {
-      /**
-       * authenticated users belonging to any of the group listed in ADMIN_GROUPS
-       */
+      // job creation
+      if (configuration().jobConfiguration.some( (j) => j.create.auth == "#all")) {
```
despadam: Instead of the hardcoded strings `#all` etc, we can import the enum from `src/jobs/types/jobs-auth.enum.ts` and use the values defined there.

src/casl/casl-ability.factory.ts
```diff
@@ -783,96 +784,175 @@ export class CaslAbilityFactory {
...
-      // endpoint authorization
-      can(AuthOp.JobsRead, JobClass);
-      cannot(AuthOp.JobsCreate, JobClass);
-      cannot(AuthOp.JobsUpdate, JobClass);
-      // -------------------------------------
-      // data instance authorization
-      can(AuthOp.JobsReadAccess, JobClass, {
-        ownerGroup: { $in: user.currentGroups },
-      });
+      } else {
```
despadam, line 845: How can we test all the following code?

src/jobs/schemas/job.schema.ts
```diff
@@ -71,17 +82,17 @@ export class JobClass extends OwnableClass {
...
-    type: [Object],
-    required: false,
-    description: "Array of status updates containing status code and message",
-  })
-  @Prop({
-    type: [Object],
-    required: false,
-  })
-  statusHistory: Record<string, string>[];
+  // // history of status codes
```
despadam: Why is this commented out? Is it not needed?

src/jobs/jobs.controller.ts
```diff
@@ -297,150 +311,131 @@ export class JobsController {
...
-          status: HttpStatus.BAD_REQUEST,
-          message: "Invalid job type: " + jobInstance.type,
-        },
-        HttpStatus.BAD_REQUEST,
-      );
-    }
-    const jobActions = matchingConfig[0].create.actions;
-
-    for (var action of jobActions) {
+    for (var action of jobInstance.configuration?.create?.actions as Array<any>) {
```
despadam: Will this come across potential errors that we don't catch because we are using the `?` operators?

src/casl/authop.enum.ts
nitrosx: anytime access is controlled by job configuration

src/casl/authop.enum.ts
nitrosx: JobCreateAny: users belonging to ADMIN_GROUPS
JobReadAny: same as above
JobStatusUpdateConfiguration: the status update is controlled by the job configuration
JobReadAccess: users which do not belong to any GROUP_ADMINS

src/casl/casl-ability.factory.ts
nitrosx: I agree.

src/jobs/config/jobConfig.example.json
nitrosx: It is!!!. This is the use case where you specify an individual user that can perform the update. 
In this case, you are saying that this type of jobs can be updated only by user archivemanager.
The assumption is that you perform the request with a JWT token that has been created by such user and such user exists and it is a valid one

src/jobs/jobs.controller.ts
nitrosx: Yes, we should highlight the potential security issues that entails when the #all option is used.
I already talked to our security team to have a better idea and about how to phrase it.

src/jobs/schemas/job.schema.ts
nitrosx: We did not have time to talk about it in our meeting about this.
I was wondering if we should separate the history from the job object in database.
I was wondering if we can create an history collection and save all the updates in there. 
Although it requires more effort, it will clean up the job document in DB. 
Thoughts?

src/casl/casl-ability.factory.ts
nitrosx: We need to add a functional account that is only part of Job Group.
We can take this off-line

src/jobs/jobs.controller.ts
nitrosx: Yes. You are correct. 
I added the `?` operators because the typescript or linting was complaining. 
We need to review the code and make sure that is robust. I was concentrated on the functionality and left addressing  the typescript type annoying messages later

src/jobs/config/jobConfig.example.json
despadam: Ok I see! But in that case then I don't understand why it is not included in the `UpdateJobAuth enum` in `src/jobs/types/jobs-auth.enum.ts`. Is it some type of special case?

despadam: Also, should it perhaps have a `#` in the beginning? So it's `#archiveManager`

src/jobs/schemas/job.schema.ts
despadam: I would be prefer to keep this here, instead of a new collection. If we create a new collection then for each create and for each update we will have to modify two collections. Also imagine if we have a large amount of jobs that we will also have another collection of the same size just to store the status history.

src/jobs/schemas/job.schema.ts
```diff
@@ -121,7 +131,45 @@ export class JobClass extends OwnableClass {
...
+    type: String,
+    required: true,
+    default: {},
+    description: "Configuration that was used to create this job.",
+  })
+  @Prop({
+    type: Object,
+    required: true,
+  })
+  configuration: Record<string, unknown>;
```
despadam: I believed we had agreed to store only the config version instead of the full configuration, to avoid overloading the job schema

src/casl/casl-ability.factory.ts
```diff
@@ -783,96 +784,198 @@ export class CaslAbilityFactory {
...
-      can(AuthOp.JobsCreateAny, JobClass, {
-        ownerGroup: { $in: user.currentGroups },
+      // instance authorization
+      can(AuthOp.JobCreateConfiguration, JobClass, {
+        "configuration.create.auth": CreateJobAuth.All,
       });
-      can(AuthOp.JobsReadAccess, JobClass, {
-        ownerGroup: { $in: user.currentGroups },
+      can(AuthOp.JobCreateConfiguration, JobClass, {
+        "configuration.create.auth": CreateJobAuth.DatasetPublic,
```
despadam: I think this syntax is not the best approach. Same as the syntax using `?` that I mentioned in another comment, I believe it leaves room for ambiguity and potential errors

despadam (commented):
I pulled all changes and get a few compilation errors. Do you also get these errors when you try it?
I also left a couple of new comments.
Lastly, I marked as `unresolved` some of the comments above that I believe should be addressed again

src/jobs/config/jobConfig.example.json
nitrosx: archiveManager does not need `#` because it is a real user and it falls under the configuration specified `USER`

src/jobs/schemas/job.schema.ts
nitrosx: Sounds reasonable, although what if we have a job that has too many update and we reach the 16MB of the document in mongodb?

src/casl/casl-ability.factory.ts
nitrosx: @despadam can you make a suggestion on how to improve it?

src/jobs/schemas/job.schema.ts
nitrosx: we did!!!
Unfortunately when I went to code the solution, I realized that I needed the configuration in the job because it makes easier and safer to check permissions at creation and update time

src/jobs/config/jobConfig.example.json
despadam: Ok thank you, I get it now

src/jobs/schemas/job.schema.ts
despadam: Good point. Has it ever happened to reach the limit? Can we somehow calculate the average size of jobs document? I cannot really picture if for our use case 16MB is enough or not

src/jobs/schemas/job.schema.ts
despadam: I think this one it's better to discuss in a meeting. When we read to job config file we store its contents in the configuration so that it's easily accessible from there. However, I understand the pros and cons of both cases, so let's talk about it again

src/casl/casl-ability.factory.ts
despadam: Let me think about it and I will try to come up with something. However, in the meantime, this syntax is responsible for some of the compilation errors that we get. Can we somehow patch it for now?

src/jobs/schemas/job.schema.ts
bpedersen2: Which job statuses do you want to record (persistently) in der database? I would restrict that to something like 
submitted, started , finished , aborted and maybe failed. If you want progress status, this I would keep in a volatile field that just gets overriden. Then I don't see a problem with this growing out of control.

src/jobs/jobs.controller.ts
despadam: we can have #all but it should have a big disclaimer in the documentation about the potential risks

src/jobs/schemas/job.schema.ts
despadam: we reuse history class for a collection to store history for both jobs and datasets. we start with jobs first

src/jobs/jobs.controller.ts
despadam: enforce typing, remove isOptional wherever is possible, set default values for empty fields, use try-catch

src/jobs/schemas/job.schema.ts
despadam: we will explore different way to write this

src/casl/casl-ability.factory.ts
sbliven: How about something like
```
const allowedJobtypes: string[] = configuration().jobConfiguration
    .filter( (x) => x.create.auth == CreateJobAuth.All || x.create.auth == CreateJobAuth.DatasetPublic)
    .map( (x) => x.jobType);
can(AuthOp.JobCreateConfiguration, JobClass, {"jobType", allowedJobTypes});
```
This assumes that can(..., {"val", ["a", "b"]}) validates that job.value is either a or b. Maybe the casl syntax is slightly different.

src/jobs/schemas/job.schema.ts
despadam: to merge this PR: we keep statusHistory in the jobs schema. Then create a separate PR to split it

src/jobs/jobs.controller.ts
despadam: will create a separate issue for this

Junjiequan (changes requested):
Type issues need to be fixed.

nitrosx:
Type issues has been fixed by the latest PR to this branch

Junjiequan (approved):

Junjiequan:
I have tested this PR locally and everything seems working fine for me.

despadam:
- Since we have agreed to move it elsewhere, I removed `statusHistory` from the `job schema` because the code was compiling and running correctly without it.
- Instead, I kept `configuration` in the `job schema` and the code as is. 
- Just needed a couple of more fixes to resolve compilation errors and now it seems to be working fine. I was able to create and update a job via swagger.
- Will create separate issues for the rest of the topics that we discussed (see comments), so that we can address them in separate PRs and merge this one.

PR #1177, opened 2024-04-16
Injecting elastic apm

Ingvord:
*ACHTUNG!!!* For educational purpose only, do not merge!

Requires `npm install elastic-apm-node --save`, see https://www.elastic.co/guide/en/apm/agent/nodejs/current/express.html


Junjiequan:
The discussion page would be the better place to suggest for the new features.
This is a nice to have feature, but not on the priority list yet. I will close it for now, we can re-open this PR when we actually is going to work on this.

PR #1182, opened 2024-04-19
fix: update datasets fullquery to use pid instead of _id

Junjiequan:
## Description

The _id and pid for datasets can be different when created manually. This change will use pid as the source of truth for the dataset's fullquery.

```
    "_id" : "20.500.12269/34e117ea-9502-404f-8db1-0c1cd5f6d9d9",
    "updatedBy" : "admin",
    "createdAt" : ISODate("2023-09-05T09:51:49.954+0000"),
    "updatedAt" : ISODate("2023-09-05T09:51:49.954+0000"),
    "ownerGroup" : "262651",
    "accessGroups" : [
        "swap"
    ],
    "pid" : "20.500.12269/b140f8fd-660d-4429-af72-77b499153b16",
```

## Motivation

## Fixes:

* Items added

## Changes:

* changes made

- [ ] Docs updated?

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -845,7 +845,7 @@ export const datasetsFullQueryDescriptionFields =
   "metadataKey": "metadata", <optional>\n \
   "_id": "item id", <optional>\n \
   "userGroups": ["group1", ...], <optional>\n \
-  "sharedWith": "email", <optional>\n \
+  "sharedWith": ["email", ...], <optional>\n \
 }\n \
   </pre>';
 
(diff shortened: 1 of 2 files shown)


nitrosx:
If we are talking about datasets, _id, id and pid should never be different.
IF we have experienced this behavior, it is a major bug and needs to be fixed

nitrosx (approved):
We need to check if datasets document with `pid` different than `_id` have been created by mistake or they are systematic.
Approving this PR anyway, although the two fields should be exactly the same

PR #1273, opened 2024-06-19
Remove configVersion, fix statusUpdate fields, update GET/jobs

### LINKED ISSUE #1265: Finalize Jobs: Store the config in each job, fix Get filtering
Based on the meeting with @nitrosx and @sofyalaski on Jun 12, regarding storing the full configuration in each job vs only the configuration version, plus a few additional topics:

 Should we save the full config or only config version in a job?
--> Keep the full config in the schema instead of only version. In case of problems we re-address it. One solution could be to create a collection of configurations.

 Decide on `configVersion` format. Current (complicated) example is "configVersion": "v1.0 2024-03-01 6f3f38"
--> Should not be stored in the job schema, but the 1st line of the config file. It should be a string. Provide indicative example of what this string could look like in the documentation, but the format is in the end up to the instance admin to decide.

 On Job Update, should we check for matching configs of job vs running instance?
--> We don't need to check if they match, we use directly the config saved in the job.

 In the config file we use 'update' instead of 'statusUpdate'
--> Change it to 'statusUpdate', to be consistent 

 Should `isFilterValid` for checking the syntax of the job GET filters be expanded? If yes how?
--> For now is ok. Add a check for the user: if admin then can get all jobs, if not admin then get jobs based on ownership
--> Double check this, also for datasets

despadam:
Fixes #1265.

Details:
- In the config file, `update` has been replaced with `statusUpdate` for consistency reasons and to avoid confusion in case we would like to introduce a general update function in the future.
- As a result, `casl-ability` has been updated and a `statusUpdate` authentication option has been added in the `casl/AuthOp`.
- `UpdateActions` have been renamed to `StatusUpdateActions`.
- `UpdateStatusJobDto` has been renamed to `StatusUpdateJobDto`.
- `configVersion` has been removed from the Job Schema.
- As a result, the util function `addConfigVesrionField` has also been removed, since it is no longer used in the `jobs.service`.
- In `jobs.service` we also introduce an additional filter in the `findAll` function which is used for `GET/jobs`: Only an admin can see all jobs, so if the user is not an admin then he can see only the jobs that he owns.
- In `jobs.controller` the `GET/jobs` function has been updated in order to user a `try-catch` that checks that the filters provided by the user are a valid Json and contain only valid fields.
- `checkJobConfiguration` and `checkConfigurationVersion` have been removed from `jobs.controller` as they are no longer being used.

### DIFF
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -1,9 +1,8 @@
 import { Logger } from "@nestjs/common";
 import {
-  JobConfig,
   loadJobConfig,
   registerCreateAction,
-  registerUpdateAction,
+  registerStatusUpdateAction,
 } from "../jobs/config/jobconfig";
 import { LogJobAction } from "../jobs/actions/logaction";
 import { EmailJobAction } from "../jobs/actions/emailaction";
@@ -238,9 +237,9 @@ export function registerDefaultActions() {
   registerCreateAction(LogJobAction);
   registerCreateAction(EmailJobAction);
   registerCreateAction(URLAction);
-  // Update
-  registerUpdateAction(LogJobAction);
-  registerUpdateAction(EmailJobAction);
+  // Status Update
+  registerStatusUpdateAction(LogJobAction);
+  registerStatusUpdateAction(EmailJobAction);
 }
 
 export type OidcConfig = ReturnType<typeof configuration>["oidc"];
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -29,7 +29,7 @@ const configuration = () => {
     process.env.DATASET_CREATION_VALIDATION_REGEX || ("" as string);
 
   const createJobGroups = process.env.CREATE_JOB_GROUPS || ("" as string);
-  const updateJobGroups = process.env.UPDATE_JOB_GROUPS || ("" as string);
+  const statusUpdateJobGroups = process.env.UPDATE_JOB_GROUPS || ("" as string);
 
   const proposalGroups = process.env.PROPOSAL_GROUPS || ("" as string);
   const sampleGroups = process.env.SAMPLE_GROUPS || ("#all" as string);
@@ -81,7 +81,7 @@ const configuration = () => {
   //   "- Create dataset privileged groups : " + createDatasetPrivilegedGroups,
   // );
   // Logger.log("- Create job groups : " + createJobGroups);
-  // Logger.log("- Update job groups : " + updateJobGroups);
+  // Logger.log("- Update job groups : " + statusUpdateJobGroups);
 
   // Register built-in job actions
   registerDefaultActions();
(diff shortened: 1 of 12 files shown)


src/jobs/config/jobConfig.schema.ts
sofyalaski, line 1: We still require a `configVersion` here. We should remove it from the list of required parameters.  Otherwise, the schema won't be validated in `jobconfig.ts`.  Now it works because the config example still contains it ( which is probably fine since we want to change it in the future).

src/config/configuration.ts
sofyalaski, line 1: at line 110 we define :
`updateJobGroups: updateJobGroups`
Since we want to rename everything to statusUpdate maybe it also makes sense to rename the group name?

src/jobs/jobs.service.ts
```diff
@@ -51,10 +46,18 @@ export class JobsService {
 
   async findAll(
     filter: IFilters<JobDocument, FilterQuery<JobDocument>>,
+    user: JWTUser
   ): Promise<JobClass[]> {
-    const whereFilters: FilterQuery<JobDocument> = filter.where ?? {};
+    var whereFilters: FilterQuery<JobDocument> = filter.where ?? {};
     const { limit, skip, sort } = parseLimitFilters(filter.limits);
 
+    if (user.username !== "admin") {
```
sofyalaski: We talked with @despadam and decided to think about it after GET tests. Because CASL should cover correct permissions but the data instance authorization is never checked for GET endpoints, permissions are granted based solely on the endpoint authorization.

src/jobs/config/jobConfig.schema.ts
despadam: `configVersion` should still remain as a field in the job configuration file, hence why we kept it as `required` when validating the configuration schema. It is removed from being stored as a separate field in the job schema, and is now only stored inside the value of a job's `configuration` field.

despadam: The difference with what we decided in [issue #1265](https://github.com/SciCatProject/scicat-backend-next/issues/1265) is that it:
> Should not be stored in the job schema, but the 1st line of the config file
(see answer to 2nd question there)

Instead, what we do here it that we have a `configVersion` field inside each job definition in the configuration file, which I understand is not optimal. If you agree, I could update the config file and schema, along with `jobconfig.ts` to have only one unique `configVersion` field at the top of the config file, which would then be passed to each job and stored inside the value of the job's `configuration` field.

sofyalaski: yes, I think it makes sense since we discussed that SciCat Live only keeps track of one configuration, so it's safe to assume that the configuration file that the facility passes should contain one unique identifier for all job types.

src/jobs/jobs.service.ts
despadam: Done in https://github.com/SciCatProject/scicat-backend-next/pull/1273/commits/749edd95e0b4bfb0e5276c18ce594740f48a9a36

src/config/configuration.ts
despadam: Done in https://github.com/SciCatProject/scicat-backend-next/pull/1273/commits/749edd95e0b4bfb0e5276c18ce594740f48a9a36

The only thing I did not rename is `UPDATE_JOB_GROUPS` defined in `.env` to avoid creating a mess when we update to v4.

I also renamed `UpdateJobAuth` to `StatusUpdateJobAuth` in `jobs-auth.enum.ts`

src/jobs/config/jobConfig.schema.ts
despadam: Done in https://github.com/SciCatProject/scicat-backend-next/pull/1273/commits/c50335be4a43a3634ea069ffa843e019362c065f

sofyalaski (approved):

Issue #745, opened 2023-09-18
Proposal/measurement-peroid: Add id field


bpedersen2:
To allow easier mapping of measurement periods to e.g. the useroffice software,
 add an id field (could done the same way as for Proposal to auto-prepend the pid prefix).

nitrosx:
@bpedersen2  do you mean to add an id specific to the measurement periods?
Currently they do not have their own collection, but they are stored as sub-document of proposal.
Should we separate them and covert them in their own model with their own endpoint?

PR #1274, opened 2024-06-20
Allow user defined `id` field in `MeasurementPeriods`

fpotier:
#745

### DIFF
--- a/src/proposals/dto/create-measurement-period.dto.ts
+++ b/src/proposals/dto/create-measurement-period.dto.ts
@@ -4,6 +4,15 @@ import { IsDateString, IsOptional, IsString } from "class-validator";
 export class CreateMeasurementPeriodDto {
   @ApiProperty({
     type: String,
+    description:
+      "Unique identifier (relative to the proposal) of a MeasurementPeriod",
+  })
+  @IsString()
+  @IsOptional()
+  id?: string;
+
+  @ApiProperty({
+    type: String,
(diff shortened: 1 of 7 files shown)


nitrosx (changes requested):
@fpotier the code looks fine and it passes the test, so if we are in a hurry to get this feature in, I will be happy to approve it.
That said, I would like to propose two different approaches:
1. place the measurement period list in it's own collection with a proper schema, which will solve the unique id issue. 
2. expand the concept of proposal with proposal type, parent proposal, and metadata (similar to dataset scientificMetadata) which will allow to create a hierarchical structure of proposals entities which has at the root your proposal and under each visit (aka measurement period)

Let's discuss

fpotier:
@nitrosx Thanks, no hurry to merge it!
I'm okay with moving MeasurementPeriods to a dedicated collection but I'm not sure that the unique id issue would be solved.
Our user office uses a simple counter for each proposal (1..n), so in our case, we need uniqueness based on the proposal id. There can be an arbitrary number of period with id 1 as long as their proposal ids are different.

nitrosx:
@fpotier is this PR still relevant considering that proposals can now be hierarchical and, also, have a property type and metadata.

fpotier:
@nitrosx this PR and #1783 where for SOLEIL's proposal ingestion. Since I left, I don't know if this is still required or can be closed

PR #1284, opened 2024-06-26
Job Status Update: Implement RabbitMQ Action

### LINKED ISSUE #1121: Implement RabbitMQ Action
Implement a JobAction to create RabbitMQ messages

despadam:
## Description

Fixes #1121 

## Motivation

The RabbitMQ action is a requirement for PSI

## Fixes:

The changes made in `JobOperation class` in `jobconfig.ts` and in  `job-create.interceptor.ts` fix the way we are storing the configuration in each job created in the database. Previously, the part of configuration file for actions was replaced by the action instance information, which in some cases makes the result we see in the database incomprehensible. We now store the action information exactly as it comes from the json file. This fix provides readability and reproducibility. 

## Changes:

- `jobConfig.example.json`: added configuration for the rabbitMQ action
- `jobs.controller`: added function to call the relevant actions on job update
- `rabbitmqaction.ts`: makes a RabbitMQ connection, creates a channel and sends a message in the queue  

**Important:** In order to spin up RabbitMQ, one needs to update the `.env` and `docker-compose.yaml` files in `scicat-ci`. These changes have been made in `despadam/scicat-ci:dev-docs` fork and should be merged as well.

## Tests:

To test whether the message has been published, one can login to the RabbitMQ management page on `localhost:15672` with the default credentials and inspect the queue.

## Notes:

The discussion about whether we should store the full configuration in each job should remain open. Storing configuration per job means that each job can run with different configuration. However, even if this has been implemented in the job schema (see also the fix mentioned above), we tend to not utilize it. More specifically, authorization in `casl-factory` and action initialization are still using the SciCat running instance configuration, instead of the job specific one, which is contradictory.

### DIFF
--- a/src/jobs/actions/rabbitmqaction.ts
+++ b/src/jobs/actions/rabbitmqaction.ts
@@ -5,7 +5,7 @@ import { JobClass } from "../schemas/job.schema";
 
 
 /**
- * Publish a message following a job status update
+ * Publish a message in a RabbitMQ queue
  */
 export class RabbitMQJobAction<T> implements JobAction<T> {
   public static readonly actionType = "rabbitmq";
@@ -38,11 +38,12 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
       "RabbitMQJobAction",
     );
 
-    if ([undefined, ""].some(el => Object.values(this.connectionDetails).includes(el))) {
-      throw new NotFoundException("Configuration is missing connection details.");
+    const connectionDetailsMissing = [undefined, ""].some(el => Object.values(this.connectionDetails).includes(el));
+    if (connectionDetailsMissing) {
+      throw new NotFoundException("RabbitMQ configuration is missing connection details.");
     }
     if (this.queueName == undefined || this.queueName == "") {
-      throw new NotFoundException("Queue name is not defined.");
+      throw new NotFoundException("RabbitMQ queue name is not defined.");
     }
   }
 
@@ -54,8 +55,7 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
 
     amqp.connect(this.connectionDetails, (connectionError: Error, connection: Connection) => {
       if (connectionError) {
-        console.log(connectionError);
-        Logger.log(
+        Logger.error(
           "Connection error in RabbitMQJobAction: " + JSON.stringify(connectionError.message),
           "RabbitMQJobAction",
         );
@@ -64,7 +64,7 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
 
       connection.createChannel((channelError: Error, channel) => {
         if (channelError) {
-          Logger.log(
+          Logger.error(
             "Channel error in RabbitMQJobAction: " + JSON.stringify(channelError.message),
             "RabbitMQJobAction",
           );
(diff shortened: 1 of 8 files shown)


nitrosx (approved):

src/jobs/actions/rabbitmqaction.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  async performJob(job: JobClass) {
+    Logger.log(
+      "Performing RabbitMQJobAction: " + JSON.stringify(job),
+      "RabbitMQJobAction",
+    );
+
+    amqp.connect(this.connectionDetails, (connectionError: Error, connection: Connection) => {
+      if (connectionError) {
+        console.log(connectionError);
```
Junjiequan: Do we need console log here?

src/jobs/actions/rabbitmqaction.ts
```diff
@@ -0,0 +1,91 @@
...
+        Logger.log(
+          "Connection error in RabbitMQJobAction: " + JSON.stringify(connectionError.message),
+          "RabbitMQJobAction",
+        );
```
Junjiequan: Can we change all the expected error messages to use Logger.error instead for consistency?

src/jobs/actions/rabbitmqaction.ts
```diff
@@ -0,0 +1,91 @@
...
+      "Validating RabbitMQJobAction: " + JSON.stringify(dto),
+      "RabbitMQJobAction",
+    );
+
+    if ([undefined, ""].some(el => Object.values(this.connectionDetails).includes(el))) {
+      throw new NotFoundException("Configuration is missing connection details.");
+    }
+    if (this.queueName == undefined || this.queueName == "") {
+      throw new NotFoundException("Queue name is not defined.");
+    }
```
Junjiequan: I'm not sure about throwing here. Wouldn't it crash the application entirely? 
For the first if condition, it would be easier to read if we assign a meaningful name to the function, `const isConnectionDetailsMissing = [...].some()`

src/jobs/interceptors/job-create.interceptor.ts
```diff
@@ -78,6 +78,28 @@ export class JobCreateInterceptor implements NestInterceptor {
         });
       }),
     );
-    return jc;
+
+    return this.mergeActionsConfigurationFields(jc as any);
```
Junjiequan: Can we replace the `any` with the actual type?

Junjiequan (changes requested):

src/jobs/actions/rabbitmqaction.ts
nitrosx: @Junjiequan thank you for catching this. 
This should be using logger

src/jobs/actions/rabbitmqaction.ts
nitrosx: Valid point. We should have tests for this and implement the correct behavior.

despadam: Throwing an error here does not crash the application. If, for example, you want to create a RabbitMQ action when creating a job, but the RabbitMQ details are not defined properly, then you get an error and the job does not get created, but the application continues to run.

Junjiequan (approved):

PR #1286, opened 2024-06-26
Jobs tests

sofyalaski:
## Description

Implements API tests for jobs authorization endpoints and proposes fixes for missing cases 

## Motivation
Fixes are proposed to follow the authorization logic described in the documentation

### jobs.controller.ts
* use `CreateJobAuth` enum instead of `AuthOp` to define `jobDatasetAuthorization`
* in  `instanceAuthorizationJobCreate` shift an "if-block" checking the presence of `ownerGroup` down after the admin check was passed (reason: admin should be able to create a job for anonym)
*  in the same function, at the very beginning when initializing values of the JobClass instance add jobParams and leave ownerUser and ownerGroup undefined (for anonymous case)* 
* in the same function, check if the current authorization is in `jobDatasetAuthorization` (types  mismatching) 
*  in the same function,  add a check if an unauthorized user creates a job for some other group (unauthorized users should only create jobs for themselves)
* in the same function, for "dataset"-kind of job configurations, the "where" filtering wasn't grabbing the respective dataset values properly (maybe go back to Despina's implementation)
* in the `checkDatasetIds` function add a check if each requested datasetId exists in database ( for the "dataset" -kind of job configurations)
* implement `generateJobInstanceForPermissions` function, which is similar to the one in datasets.controller. mongoose `findOne` and `findAll` do not return an instance of JobClass as implemented but a simple JS Object. Since we still need to pass that object to casl( which expects the instance of JobClass) we need conversion. To not write all properties of JobClass, a minimal copy with only properties required for permission checking is created. The return instances are not affected by this and remain  JS Object.
*  functionality of `instanceAuthorizationJobStatusUpdate` is now implemented in the patch endpoint function. This is done to avoid redefining a job return by `findOne` as a instance of JobClass with all properties.
* for both GET endpoints only the endpoint authorisation was implemented. To add the instance authorization too, first create JobClass instance of the found job Objects (as described in two above items) and pass them to respective casl expressions. 
* adds implementation of delete endpoints. The rules are defined in the CheckPolicies decorator 

### jobs.service.ts 
* add a function to extract the username of a user sending a request ( should return "anonymous" if request sent by unauthenticated user). This is used both in `create` and`statusUpdate` functions 

### casl-ability.factory.ts   
* for statusUpdate endpoint the user from update_job_groups had less rights that a normal user -> fixed that 
* add instance authorization for statusUpdate for anonymous user
* implement delete endpoint authorization  
* fixed a negation typo 

## Changes
* add deleteJobGroups to `configuration.ts`

## Tests included

### DIFF
--- a/src/jobs/actions/rabbitmqaction.ts
+++ b/src/jobs/actions/rabbitmqaction.ts
@@ -3,7 +3,6 @@ import amqp, { Connection } from "amqplib/callback_api";
 import { JobAction } from "../config/jobconfig";
 import { JobClass } from "../schemas/job.schema";
 
-
 /**
  * Publish a message in a RabbitMQ queue
  */
@@ -12,6 +11,7 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
   private connection;
   private binding;
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   constructor(data: Record<string, any>) {
     Logger.log(
       "Initializing RabbitMQJobAction. Params: " + JSON.stringify(data),
@@ -28,7 +28,7 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
     this.binding = {
       exchange: data.exchange,
       queue: data.queue,
-      key: data.key
+      key: data.key,
     };
   }
 
@@ -42,14 +42,22 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
       "RabbitMQJobAction",
     );
 
-    const connectionDetailsMissing = [undefined, ""].some(el => Object.values(this.connection).includes(el));
+    const connectionDetailsMissing = [undefined, ""].some((el) =>
+      Object.values(this.connection).includes(el),
+    );
     if (connectionDetailsMissing) {
-      throw new NotFoundException("RabbitMQ configuration is missing connection details.");
+      throw new NotFoundException(
+        "RabbitMQ configuration is missing connection details.",
+      );
     }
 
-    const bindingDetailsMissing = [undefined, ""].some(el => Object.values(this.binding).includes(el));
+    const bindingDetailsMissing = [undefined, ""].some((el) =>
+      Object.values(this.binding).includes(el),
+    );
     if (bindingDetailsMissing) {
-      throw new NotFoundException("RabbitMQ binding is missing exchange/queue/key details.");
+      throw new NotFoundException(
+        "RabbitMQ binding is missing exchange/queue/key details.",
+      );
     }
   }
 
@@ -59,33 +67,47 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
       "RabbitMQJobAction",
     );
 
-    amqp.connect(this.connection, (connectionError: Error, connection: Connection) => {
-      if (connectionError) {
-        Logger.error(
-          "Connection error in RabbitMQJobAction: " + JSON.stringify(connectionError.message),
-          "RabbitMQJobAction",
-        );
-        return;
-      }
-
-      connection.createChannel((channelError: Error, channel) => {
-        if (channelError) {
+    amqp.connect(
+      this.connection,
+      (connectionError: Error, connection: Connection) => {
+        if (connectionError) {
           Logger.error(
-            "Channel error in RabbitMQJobAction: " + JSON.stringify(channelError.message),
+            "Connection error in RabbitMQJobAction: " +
+              JSON.stringify(connectionError.message),
             "RabbitMQJobAction",
           );
           return;
         }
 
-        channel.assertQueue(this.binding.queue, { durable: true });
-        channel.assertExchange(this.binding.exchange, "topic", { durable: true });
-        channel.bindQueue(this.binding.queue, this.binding.exchange, this.binding.key);
-        channel.sendToQueue(this.binding.queue, Buffer.from(JSON.stringify(job)));
+        connection.createChannel((channelError: Error, channel) => {
+          if (channelError) {
+            Logger.error(
+              "Channel error in RabbitMQJobAction: " +
+                JSON.stringify(channelError.message),
+              "RabbitMQJobAction",
+            );
+            return;
+          }
 
-        channel.close(() => {
-          connection.close();
+          channel.assertQueue(this.binding.queue, { durable: true });
+          channel.assertExchange(this.binding.exchange, "topic", {
+            durable: true,
+          });
+          channel.bindQueue(
+            this.binding.queue,
+            this.binding.exchange,
+            this.binding.key,
+          );
+          channel.sendToQueue(
+            this.binding.queue,
+            Buffer.from(JSON.stringify(job)),
+          );
+
+          channel.close(() => {
+            connection.close();
+          });
         });
-      });
-    });
+      },
+    );
   }
 }
--- a/src/jobs/actions/rabbitmqaction.ts
+++ b/src/jobs/actions/rabbitmqaction.ts
@@ -11,24 +11,42 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
   private connection;
   private binding;
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  constructor(data: Record<string, any>) {
+  constructor(data: Record<string, unknown>) {
     Logger.log(
       "Initializing RabbitMQJobAction. Params: " + JSON.stringify(data),
       "RabbitMQJobAction",
     );
 
+    // Validate that all necessary params are present
+    const requiredConnectionParams = [
+      "hostname",
+      "port",
+      "username",
+      "password",
+    ];
+    for (const param of requiredConnectionParams) {
+      if (!data[param]) {
+        throw new NotFoundException(`Missing connection parameter: ${param}`);
+      }
+    }
+
+    const requiredBindingParams = ["exchange", "queue", "key"];
+    for (const param of requiredBindingParams) {
+      if (!data[param]) {
+        throw new NotFoundException(`Missing binding parameter: ${param}`);
+      }
+    }
     this.connection = {
       protocol: "amqp",
-      hostname: data.hostname,
-      port: data.port,
-      username: data.username,
-      password: data.password,
+      hostname: data.hostname as string,
+      port: data.port as number,
+      username: data.username as string,
+      password: data.password as string,
     };
     this.binding = {
-      exchange: data.exchange,
-      queue: data.queue,
-      key: data.key,
+      exchange: data.exchange as string,
+      queue: data.queue as string,
+      key: data.key as string,
     };
   }
 
@@ -36,30 +54,8 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
     return RabbitMQJobAction.actionType;
   }
 
-  async validate(dto: T) {
-    Logger.log(
-      "Validating RabbitMQJobAction: " + JSON.stringify(dto),
-      "RabbitMQJobAction",
-    );
-
-    const connectionDetailsMissing = [undefined, ""].some((el) =>
-      Object.values(this.connection).includes(el),
-    );
-    if (connectionDetailsMissing) {
-      throw new NotFoundException(
-        "RabbitMQ configuration is missing connection details.",
-      );
-    }
-
-    const bindingDetailsMissing = [undefined, ""].some((el) =>
-      Object.values(this.binding).includes(el),
-    );
-    if (bindingDetailsMissing) {
-      throw new NotFoundException(
-        "RabbitMQ binding is missing exchange/queue/key details.",
-      );
-    }
-  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async validate(dto: T) {}
 
   async performJob(job: JobClass) {
     Logger.log(
--- a/src/jobs/actions/urlaction.ts
+++ b/src/jobs/actions/urlaction.ts
@@ -27,8 +27,7 @@ export class URLAction<T> implements JobAction<T> {
   private urlTemplate: Handlebars.TemplateDelegate<JobClass>;
   private method = "GET";
   private headers: Record<string, string> = {};
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  private body: Record<string, any> | null = null;
+  private body: Record<string, unknown> | null = null;
 
   getActionType(): string {
     return URLAction.actionType;
--- a/src/jobs/actions/urlaction.ts
+++ b/src/jobs/actions/urlaction.ts
@@ -1,9 +1,4 @@
-import {
-  Logger,
-  NotFoundException,
-  BadRequestException,
-  HttpException,
-} from "@nestjs/common";
+import { Logger, NotFoundException, HttpException } from "@nestjs/common";
 import { JobAction } from "../config/jobconfig";
 import { JobClass } from "../schemas/job.schema";
 import * as Handlebars from "handlebars";
@@ -32,12 +27,14 @@ export class URLAction<T> implements JobAction<T> {
   private urlTemplate: Handlebars.TemplateDelegate<JobClass>;
   private method = "GET";
   private headers: Record<string, string> = {};
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private body: Record<string, any> | null = null;
 
   getActionType(): string {
     return URLAction.actionType;
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async validate(dto: T) {}
 
   async performJob(job: JobClass) {
@@ -75,6 +72,7 @@ export class URLAction<T> implements JobAction<T> {
    *
    * @throws {NotFoundException} If the 'url' parameter is not provided in the data object
    */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   constructor(data: Record<string, any>) {
     if (!data["url"]) {
       throw new NotFoundException("Param 'url' is undefined in url action");
--- a/src/jobs/config/jobconfig.ts
+++ b/src/jobs/config/jobconfig.ts
@@ -31,20 +31,17 @@ export class JobConfig {
   jobType: string;
   configVersion: string;
   create: JobOperation<CreateJobDto>;
-  // read: JobReadAction[];
   statusUpdate: JobOperation<StatusUpdateJobDto>;
 
   constructor(
     jobType: string,
     configVersion: string,
     create: JobOperation<CreateJobDto>,
-    read = undefined,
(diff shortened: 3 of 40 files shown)


nitrosx (approved):

test/config/jobconfig.json
```diff
@@ -1,53 +1,80 @@
-[
+{
+  "configVersion": "v1.0 2024-03-01 6f3f38",
+  "jobs": [
   {
-    "jobType": "archive",
+    "jobType": "all_access",
+    "configVersion": "v1.0 2024-03-01 6f3f38",
```
despadam: There is a `configVersion` per job but also a general one at the top of the file. We should only keep the top one. the current implementation expects only that and then stores this information inside each job, but it's not needed to duplicate it per job in the configuration file.

test/Jobs.js
```diff
@@ -1,33 +1,151 @@
...
+};
+const jobGroup5 = {
+  ...TestData.Job, 
+  type: "group_access"
+};
+
+describe("1100: Jobs: Test New Job Model", () => {
+  before(() => {
+    db.collection("Dataset").deleteMany({});
+    db.collection("Job").deleteMany({});
```
despadam, line 145: I have been trying to figure out why this part is not working. We expect that it would empty the `Dataset` and `Job` collections, so that we can start the tests with an empty database, however this does not happen. 

One idea to solve this would be to prune the `mongodb volume` before starting the test, but I do not know whether this is the right approach. 

In my effort to understand what these two lines of code are doing, I came across the file `pretest.js` file, which contains the following:
```
const uri = "mongodb://localhost:27017/scicat";
const client = new MongoClient(uri);
global.db = client.db("scicat");
```
So now my question is why are we trying to connect to `scicat` db? Shouldn't it be `test_be_next`? Or perhaps there is something that I don't quite understand. Could someone please this to me? Perhaps @sofyalaski or @Junjiequan ?

despadam (changes requested):

test/Jobs.js
Junjiequan: It's my mistake. the db name should be customizable. In your case it didn't work because you are not creating db named `scicat`, so the local test would fail. I believe there's a ENV variable named ~~test_db or sth~~ `MONGODB_URI`, we can change to that instead hardcode here. I will fix it next week and open a PR.
Thanks for mentioning this.

src/jobs/jobs.controller.ts
```diff
@@ -319,9 +314,57 @@ export class JobsController {
...
+        where: {
+          pid: { $eq: datasetIds[i] }
+        },
+      };
+      const idCount = await this.datasetsService.count(filter);
+      if (idCount.count <1){
+        throw new HttpException(
+          {
+            status: HttpStatus.BAD_REQUEST,
+            message: "At least one dataset in the ids list doesn't.",
```
despadam: The error message here is not clear.

src/jobs/jobs.controller.ts
```diff
@@ -46,6 +48,9 @@ import {
   filterExampleSimplified,
 } from "src/common/utils";
 import { JobCreateInterceptor } from "./interceptors/job-create.interceptor";
+import { ConnectableObservable } from "rxjs";
+import { integer } from "@elastic/elasticsearch/lib/api/types";
+import { boolean } from "mathjs";
```
despadam: Are these needed?

src/jobs/jobs.controller.ts
```diff
@@ -16,13 +16,15 @@ import {
 } from "@nestjs/common";
 import { Request } from "express";
 import { FilterQuery } from "mongoose";
+import { plainToClass, plainToInstance } from 'class-transformer';
```
despadam: Are these needed?

src/jobs/jobs.controller.ts
```diff
@@ -368,11 +404,20 @@ export class JobsController {
             jobCreateDto.ownerUser,
           );
         }
-
         jobInstance.ownerUser = jobUser?.username as string;
         jobInstance.contactEmail = jobUser?.email as string;
         jobInstance.ownerGroup = jobCreateDto.ownerGroup;
```
despadam, line 436: What is the expected result of these lines? I tested it by logging in as an admin and creating a job without providing ownerUser or ownerGroup. I am printing the `jobInstance` after these lines but everything seems to be undefined: createdBy, ownerGroup,, ownerUser, contactEmail are all undefined.

src/jobs/jobs.controller.ts
```diff
@@ -390,44 +435,76 @@ export class JobsController {
...
+            {
+              status: HttpStatus.BAD_REQUEST,
+              message: "Dataset ids list was not provided in jobParams",
+            },
+            HttpStatus.BAD_REQUEST,
+          );
+        }
       // verify that the user meet the requested permissions on the datasets listed
-      const datasetIds = this.checkDatasetIds(jobCreateDto.jobParams);
+      // const datasetIds = await this.checkDatasetIds(jobCreateDto.jobParams);
```
despadam, line 491: Why is this commented out?

src/jobs/jobs.controller.ts
```diff
@@ -572,11 +626,18 @@ export class JobsController {
...
-    // Check job authorization
-    await this.instanceAuthorizationJobStatusUpdate(
-      request.user as JWTUser,
-      currentJob,
-    );
+    const currentJobInstance = await this.generateJobInstanceForPermissions(currentJob);
+
+    const ability = this.caslAbilityFactory.createForUser(request.user as JWTUser);
+    // check if he/she can create this dataset
+    const canCreate =
```
despadam: Change comment and variable name to `update` instead of `create`

src/jobs/jobs.controller.ts
```diff
@@ -610,7 +671,25 @@ export class JobsController {
...
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+    const currentJobInstance = await this.generateJobInstanceForPermissions(currentJob);
+    const ability = this.caslAbilityFactory.createForUser(request.user as JWTUser);
+    const canCreate =
+      ability.can(AuthOp.JobReadAny, JobClass) ||
+      ability.can(AuthOp.JobReadAccess, currentJobInstance);
+    if (!canCreate) {
+      throw new ForbiddenException("Unauthorized to update this dataset");
```
despadam, line 784: Rename variable and change error message to be representative to GET

src/jobs/jobs.controller.ts
```diff
@@ -670,9 +764,7 @@ export class JobsController {
    * Delete a job
    */
   @UseGuards(PoliciesGuard)
-  @CheckPolicies(
-    (ability: AppAbility) => ability.can(AuthOp.Delete, JobClass), // TBD
-  )
+  @CheckPolicies((ability: AppAbility) => ability.can(AuthOp.JobDelete, JobClass)&& ability.can(AuthOp.JobDeleteAny, JobClass) )
```
despadam: Need some more details on this. Do we need to check whether the specific used has the rights to delete a specific job? Is this check enough?

When logged in as an admin, I tried to delete a job that I created but got error `403: forbidden resource`

despadam (changes requested):
Reviewed `jobs.controller`

src/jobs/jobs.controller.ts
sofyalaski: it's already called in line 392. 
The reasoning behind that was that datasets were only checked when the configuration was set to "#datasetX". Now they are being checked for all configuration if "datasetIds" are passed to jobParams.

src/jobs/jobs.controller.ts
sofyalaski: "createdBy" field should be created by `utils.addCreatedByFields` from `jobs.service` and when I am logged in, it usually fills that field. Are you sure you were logged in? For the other fields, they are left empty for the case where admin creates a job for anonymous user. Then it's important to specify the email. For other cases, where users exist in the database, the email is going to be grabbed from the user data.

src/jobs/jobs.controller.ts
sofyalaski: Yes, If you look in the [docs](https://scicatproject.github.io/documentation/Development/v4.x/backend/authorization/authorization_jobs.html) (first table) admin shouldn't have rights to delete jobs, unless he/she is in the DELETE_JOB_GROUPS as defined in env file. I put archiveManager similar to datasets.

test/config/jobconfig.json
sofyalaski: Resolved in the latest commit

sofyalaski:
@despadam I made some changes you requested. For other comments, we can discuss tomorrow in person

sofyalaski:
@Despina, I've updated the change to configuration grabbing for statusUpdate as  we discussed

test/config/pretest.js
```diff
@@ -3,7 +3,7 @@
 
 var chaiHttp = require("chai-http");
 const { MongoClient } = require("mongodb");
-const uri = "mongodb://localhost:27017/scicat";
+const uri = process.env.MONGODB_URI;
```
Junjiequan: If we set the test DB URI as the MONGODB_URI, it will wipe the local development DB if someone forgets to change the URI before running the test. Therefore, I suggest keeping it unchanged for now.

Junjiequan (changes requested):

test/config/pretest.js
sbliven: Good point. `appUrl` is also hard-coded a few lines down. Unless there was a particular reason this needs to be changed at runtime?

src/jobs/config/jobconfig.ts
```diff
@@ -54,25 +50,21 @@ export class JobConfig {
    * @param data JSON
    * @returns
    */
-  static parse(
-    jobData: Record<string, any>,
-    configVersion: string
-  ): JobConfig {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
sbliven: Is `Record<string, any>` the best type for jobData? I had to disable linting because of the `any` here and in all subclasses. I think that's reasonable though, since it really is unchecked.

src/jobs/actions/urlaction.ts
```diff
@@ -32,12 +27,14 @@ export class URLAction<T> implements JobAction<T> {
...
   private method = "GET";
   private headers: Record<string, string> = {};
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private body: Record<string, any> | null = null;
 
   getActionType(): string {
     return URLAction.actionType;
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
sbliven, line 36: Weird that this empty method causes a lint warning; it overrides the interface method.

src/jobs/jobs.controller.ts
```diff
@@ -320,31 +310,77 @@ export class JobsController {
...
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+    for (const i in datasetIds) {
+      const filter: condition = {
+        where: {
+          pid: { $eq: datasetIds[i] },
+        },
+      };
+      const idCount = await this.datasetsService.count(filter);
```
sbliven: Wouldn't it be more performant to make a single database request here, rather than one request per datasetId?

src/jobs/jobs.controller.ts
```diff
@@ -320,31 +310,77 @@ export class JobsController {
...
+      if (idCount.count < 1) {
+        throw new HttpException(
+          {
+            status: HttpStatus.BAD_REQUEST,
+            message: "At least one dataset in the ids list doesn't exist.",
```
sbliven: Should include `datasetIds[i]` in the error message to make debugging easier.

src/jobs/jobs.controller.ts
sbliven: So delete it?

src/jobs/jobs.controller.ts
```diff
@@ -758,6 +866,7 @@ export class JobsController {
     @Req() request: Request,
     @Param("id") id: string,
   ): Promise<unknown> {
+    Logger.log("Deleting job!");
```
sbliven: Should include the id in the log message

test/Jobs.js
```diff
@@ -40,755 +158,3707 @@ describe.skip("1100: Jobs: Test New Job Model", () => {
...
+                          (tokenVal) => {
+                            accessTokenAdmin = tokenVal;
+                            utils.getToken(
+                              appUrl,
+                              {
+                                username:"archiveManager",
+                                password: TestData.Accounts["archiveManager"]["password"],
+                              },
+                              (tokenVal) => {
+                                accessTokenArchiveManager = tokenVal;
```
sbliven, line 206: That's a lot of nesting! Not necessary for this PR, but maybe `getToken` needs to be refactored to return a promise rather than use callbacks.

test/Jobs.js
```diff
@@ -40,755 +158,3707 @@ describe.skip("1100: Jobs: Test New Job Model", () => {
...
+  //     .send(newDataset)
+  //     .set("Accept", "application/json")
+  //     .set({ Authorization: `Bearer ${accessTokenAdmin}` })
+  //     .expect(TestData.BadRequestStatusCode)
+  //     .expect("Content-Type", /json/)
+  //     .then((res) => {
+  //       res.body.should.not.have.property("id");
+  //       res.body.should.have.propert("message").and.be.equal("Dataset ids list was not provided in jobParams");
+  //     });
+  // });
```
sbliven: Why commented?

test/config/pretest.js
Junjiequan: I don't think we need `appUrl` to be changed at run time. I'm not sure if we have environment variables to assign to the `appUrl`

Junjiequan: After discussing with @nitrosx regarding the test database, he suggested using the MONGODB_URI variable for the test database as well. In the documentation, we can mention that one should change the MONGODB_URI before running the test command to prevent accidental local data wipes.

Here are some minor suggestions for the code changes:

- Include require("dotenv").config() to load the environment variables. Without this, the environment variables will be undefined.
- For global.db, replace client.db(uri.split("/").pop()); with client.db().

```suggestion
require("dotenv").config();
var chaiHttp = require("chai-http");

const { MongoClient } = require("mongodb");

const client = new MongoClient(process.env.MONGODB_URI);

async function loadChai() {
  const { chai } = import("chai");
  chai.use(chaiHttp);
  await client.connect();
}

loadChai();
global.appUrl = "http://localhost:3000";
global.request = require("supertest");
global.db = client.db();
```

Junjiequan (approved):

sbliven (approved):
This looks pretty good. I added a couple minor comments but it looks good.

Can you verify that mocha passes before merging? @despadam and I both ran into issues running the tests. Mocha should be added to the CI following this PR.

src/jobs/config/jobconfig.ts
sbliven: I realized that `unknown` fits better here than any.

src/jobs/actions/rabbitmqaction.ts
```diff
@@ -12,80 +11,99 @@ export class RabbitMQJobAction<T> implements JobAction<T> {
...
-      throw new NotFoundException("RabbitMQ configuration is missing connection details.");
-    }
-
-    const bindingDetailsMissing = [undefined, ""].some(el => Object.values(this.binding).includes(el));
-    if (bindingDetailsMissing) {
-      throw new NotFoundException("RabbitMQ binding is missing exchange/queue/key details.");
-    }
-  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async validate(dto: T) {}
```
despadam, line 58: @sbliven I struggle to understand the need for `validate()` in the `action class`. I was always wondering whether it makes sense validating after calling the constructor, and your changes here I think prove my point that it probably doesn't. I would suggest following this example and moving all validation inside the constructor of each action. Then:
- On job `create` if creating an action fails, the job creation should also fail.
- On job `statusUpdate` things get a bit more complicated. Since we are using a global job configuration, instead of job specific, if no changes were made to the configuration then no validation is needed at all and we can proceed directly to performing the job. But, if the configuration has changed, we would need a method that would first update the values of the class' attributes, before performing the action.

test/Jobs.js
sofyalaski: yes, I was confused by the logic myself a bit. I changed this one into 0065. Now these tests at the beginning cover:
- `jobParams` is passed with property `datasetIds`, but the array is empty --> fail (0040)
- `jobParams` is passed with property `datasetIds`, but it refers to datasetIds that are not in database --> fail (0050)
- `jobParams` is not passed at all --> fail (0060)
- `jobParams` is passed as empty object --> fail (0065)

I additionally implemented the check for `jobParams` property in the `create` endpoint. It checks that `jobParams` is passed and is not empty. Before that if no `jobParams` were passed a 500 internal error was thrown. Now the error message is explicitly saying that `jobParams` need to be defined and not empty.   

the last two tests are new. 0060 before was testing with an empty object as jobParams, which passed before. Now, as I I changed the implementation a little, this test fails. I think it is logical that if a job has no context it shouldn't be accepted? This test now is numbered as 0065.
0060 tests the case where no `jobParams` passed at all.

src/jobs/jobs.controller.ts
sofyalaski: addressed in the last commit

src/jobs/jobs.controller.ts
sofyalaski: Yes, you're right, I changed that in the last commit

src/jobs/jobs.controller.ts
sofyalaski: addressed in the last commit

src/jobs/actions/rabbitmqaction.ts
despadam: After discussing with @sofyalaski, I will create a separate PR to address this.

sofyalaski:
@sbliven in[ the commit](https://github.com/SciCatProject/scicat-backend-next/pull/1286/commits/9fa1b8ef069f9a924e2493abc70a42400e78dfbb) two weeks ago I added running the tests on PR to the "relase-*" branches, and you can see the bars here at the bottom, I assume. For me, the API tests seem to be passing (API test include both `test:api:jest` and `test:api:mocha`. It might be because of some differences in the way we deploy then? We tried running the brand new docker container and running tests in it with Despina and it worked. Also, sorry for not telling that you need to start the app first!

test/Jobs.js
sofyalaski: This one is addressed in PR #1350

src/jobs/actions/rabbitmqaction.ts
despadam: Will be covered by validate action

Issue #1294, opened 2024-06-28
Re-evaluate whether a job should have individual configuration


despadam:
Based on previous discussions, the current implementation for reading and storing `job configuration` is as follows:

- The job configuration file (see `jobConfig.example.json`) contains a field called `configVersion` at the top, which indicates the version of the file. 
- When creating a job, we store the full configuration definition for this specific job type, inside the job schema, appended with the configuration file version.
- When we read the configuration file in `config/configuration.js`, we store the extracted information in property `jobConfiguration`. This corresponds to the running instance's configuration.
- Based on `jobConfiguration` we also initialize the job actions. This means that actions don't have job-specific configuration but instance-specific.
- Jobs Authorization in `cals-ability.factory` is also based on the instance-specific configuration.

The above statements mean that even though we store the job-specific configuration in the db, in the end we do not use it. As a result, we perform authorization and actions based on different parameters than expected.

We need a solid use case for which job-specific configuration is required if we would like to push for this implementation. If we agree on that, then the above-mentioned parts of the code would change significantly. 
Otherwise, implementation stays for the most part as is and we finalize how we store the configuration information in the db. In that case this information will not need to be duplicated inside each job, but stored globally somewhere else. If we need the full configuration information inside the db, then one idea would be to create a Configuration collection and only store a foreign key to specific config version in the job schema. If not, then perhaps storing only the configVersion in the job schema would be enough.

despadam:
The `job-create.interceptor` performs `validateJob` which also uses the intance-specific configuration, instead of job-specific.

Additionally, there is only an interceptor for `create`. Shouldn't we also add one for `statusUpdate`?

despadam:
`instanceAuthorizationJobCreate` in `jobs.controller` uses `jobCreateDto.configuration`

despadam:
See last related changes: https://github.com/SciCatProject/scicat-backend-next/issues/1265

despadam:
Based on the discussion with @nitrosx, @sofyalaski and @sbliven during **SciCatCon**:

For the MVP we will enforce the use of `instance-specific` configuration instead of `job-specific`. This will allow us to fix and merge the current PRs and finish the testing.

Afterwards, we will invite participants from the rest of the facilities to present their use cases and requirements, in order to make our final decision.

despadam:
All jobs are based on the same configuration, individual configuration is not allowed, apart from action selection. 
[...]

PR #1320, opened 2024-07-09
Jobs should only use the SciCat instance configuration

despadam:
## Description

Based on the last comment under topic https://github.com/SciCatProject/scicat-backend-next/issues/1294.

> For the MVP we will enforce the use of instance-specific configuration instead of job-specific. This will allow us to fix and merge the current PRs and finish the testing.

## Changes:

In `jobs.controller`: `instanceAuthorizationJobCreate` and `performJobCreateAction` have been updated to extract the necessary information for the job configuration per job type, from the global configuration variable that is created when the SciCat instance starts running. Previously they were using the job-specific configuration that is being stored within each job instance.

## Notes:

- The full job type specific configuration is still being stored in the db when a new job instance is created. We keep it like this for now, so that we don't apply changes to the job schema, as it has been finalized. After the MVP implementation we will decide whether this should change too.

### DIFF
--- a/src/jobs/jobs.controller.ts
+++ b/src/jobs/jobs.controller.ts
@@ -362,7 +362,6 @@ export class JobsController {
     // If other fields are needed can be added later.
     const jobInstance = new JobClass();
     const jobConfiguration = this.getJobMatchingConfiguration(jobCreateDto);
-    console.log(jobConfiguration);
 
     jobInstance._id = "";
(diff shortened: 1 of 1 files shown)


sofyalaski (approved):
Thanks Despina, 
I will need to add changes to the [PR 1286](https://github.com/SciCatProject/scicat-backend-next/pull/1286)  to agree with this

Issue #1152, opened 2024-04-04
EmailJobAction should accept the email body from a file


sbliven:
The body of the email was typically a fairly long handlebars template. See [job-template.html](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/common/email-templates/job-template.html).

`EmailJobAction` should have a configuration option to specify this as a filename rather than directly in the config file. It should default to `common/email-templates/job-template.html`.

despadam:
Done in https://github.com/SciCatProject/scicat-backend-next/pull/1326

PR #1326, opened 2024-07-22
Jobs: templating for email action

despadam:
## Motivation:

https://github.com/SciCatProject/scicat-backend-next/issues/1152

## Description:

The email action is implemented so that it validates all relevant values provided in the configuration file, inside the constructor. When performing the action, the email body is read from a template file, the path for which is provided in the configuration. We then fill in the values in the template with the respective job-specific information.

Please note that this is a first and basic implementation of the email action, which acts as a proof of concept for the Jobs MVP. The mailer (i.e. SMTP) is not property set up, so in the end we never actually send the email. Instead, we only perform validation and templating, and print the result of what the email content would look like, in the console.

## Changes:

- In `jobConfig.example.json`: included configuration for the email action. It is possible that this is not the final schema, especially when it comes to providing the value for `password`. TBD as it is dependent on how the mailer service is set up in the facility. Please note the assumption that `bodyTemplate` should be the path to the template file.
- `job-template-simplified.html` is a minimal template example that requires only some basic job information. To be expanded.
- `emailaction.ts`: based on previous changes in the Action class, all validation is performed inside the constructor. Setting up `mailService` is commented out as we are currently not actually making the connection to the service. We open and read the file for which the path is provided in `bodyTemplate`. We fill in the template with the job details. Only the necessary fields are exposed to the template via `jobTemplateOptions`. In the end, instead of sending an actual email, we print the result in the console.
- `jobs.controller.ts`: applied a fix in order to always correctly provide to the template a list of `datasetIds` inside `jobParams`.


src/jobs/actions/emailaction.ts
```diff
@@ -38,31 +55,49 @@ export class EmailJobAction<T> implements JobAction<T> {
...
       throw new NotFoundException("Param 'to' is undefined");
     }
     if (!data["from"]) {
       throw new NotFoundException("Param 'from' is undefined");
     }
     if (typeof data["from"] !== "string") {
-      throw new TypeError("from should be a string");
+      throw new TypeError("Param 'from' should be a string");
+    }
+    if (!data["password"]) {
```
bpedersen2: Password should be optional, as not all mailhosts do require one.  And From and auth may differ.

I would suggest an optional auth field that can be passed as-is into the  auth member of createTransport.
This also allows to configure the other possible auth schemes easier.

despadam: Thanks for the suggestion @bpedersen2, I added the optional `auth` field in the latest commit https://github.com/SciCatProject/scicat-backend-next/pull/1326/commits/64c897195a5d3cd1b967dca4bdcc9a94870d3c77

src/jobs/actions/emailaction.ts
```diff
@@ -38,31 +60,49 @@ export class EmailJobAction<T> implements JobAction<T> {
...
     }
-    if (typeof data["from"] !== "string") {
-      throw new TypeError("from should be a string");
-    }
     if (!data["subject"]) {
       throw new NotFoundException("Param 'subject' is undefined");
     }
-    if (!data["body"]) {
-      throw new NotFoundException("Param 'body' is undefined");
+    if (!data["bodyTemplate"]) {
```
sbliven: Should we be explicit about calling this being a filename by calling it eg `"bodyTemplateFile"`?

sbliven (approved):
This looks fine with the understanding that the SMTP part is a stub. For instance, in the real thing the user/password should not come from jobConfig.json but from environmental variables. We will need a GraphQL implementation too for MS365 emails.

src/jobs/actions/emailaction.ts
despadam: Renamed the field here: https://github.com/SciCatProject/scicat-backend-next/pull/1326/commits/3a75dde9760ed89eb5584d8abffc6913bc24ad9d

PR #1327, opened 2024-07-22
Feat: add endpoint to check user access for sample and proposal

Junjiequan:
## Description

This pull request introduces new endpoints to check user access for specific samples and proposals. It includes updates to the authorization mechanisms and related test cases to ensure proper access control based on user roles and permissions.

Additional: 
1.  Added configurable test database environment variable is added for local testing.
2. Replaced deprecated casl-factory AbilityBuilder with updated one.

## Motivation

When a user opens a dataset detail page, the frontend makes requests to fetch the Sample and Proposal attached to the dataset, regardless of whether the user has permission to access them. By utilizing user access endpoints, we can pre-check if the user has access to the Sample or Proposal before making subsequent requests. This can reduce the collection of `FALSY` errors.

## Changes
**New Endpoints:**
- Added `/samples/:id/authorization` and `/proposals/:pid/authorization` endpoints to check if the user has access to specific samples and proposals.

**Authorization Updates:**
- Modified CASL ability factory to include updated types and conditions for checking access permissions.
- Updated policies guards to handle access control for the new endpoints.

**Tests:**
- Added test cases for new endpoints in ProposalAuthorization.js and SampleAuthorization.js to validate access control for different user roles.

## Tests included

## Documentation


Junjiequan:
The decision on whether to allow users to update the isPublished status of proposals requires further investigation. In the proposalAuthorization test file, test cases for public proposals are temporarily commented out.

.env.example
```diff
@@ -17,6 +17,7 @@ LOGBOOK_BASE_URL="http://localhost:3030/scichatapi"
 METADATA_KEYS_RETURN_LIMIT=100
 METADATA_PARENT_INSTANCES_RETURN_LIMIT=100
 MONGODB_URI="mongodb://<USERNAME>:<PASSWORD>@<HOST>:27017/<DB_NAME>"
+MONGDB_TEST_NAME="scicat-test"
```
nitrosx: Why do we need this variable? The database name is already included in MONGODB_URI.
IF we are in a testing environment, we should change MONGODB_URI locally.
If we are going to keep it, can you fix the misspelling? It should MONGODB_TEST_NAME

README.md
```diff
@@ -148,6 +149,7 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
 | `METADATA_KEYS_RETURN_LIMIT` | number | Yes | The return limit for the `/Datasets/metadataKeys` endpoint. | |
 | `METADATA_PARENT_INSTANCES_RETURN_LIMIT` | number | Yes | The return limit of Datasets to extract metadata keys from for the `/Datasets/metadataKeys` endpoint. | |
 | `MONGODB_URI` | string | | The URI for your MongoDB instance. | |
+| `MONGDB_TEST_NAME` | string | | The DB name for your MongoDB instance that runs on the local testing environment. |"scicat-test"|
```
nitrosx: Same comment as before

src/proposals/proposals.controller.ts
```diff
@@ -546,9 +545,49 @@ export class ProposalsController {
...
+
     return proposal;
   }
 
+  // GET /proposals/:pid/access
+  @UseGuards(PoliciesGuard)
+  @CheckPolicies((ability: AppAbility) =>
+    ability.can(Action.ProposalsRead, ProposalClass),
+  )
+  @Get("/:pid/access")
```
nitrosx: Can we name this endpoint /proposal/:id/authorization?

src/samples/samples.controller.ts
```diff
@@ -546,12 +546,54 @@ export class SamplesController {
...
+    return sample;
   }
 
-  // PATCH /samples/:id
+  // GET /samples/:id/access
   @UseGuards(PoliciesGuard)
+  @CheckPolicies((ability: AppAbility) =>
+    ability.can(Action.SampleRead, SampleClass),
+  )
+  @Get("/:id/access")
```
nitrosx: Same as for proposal. I think this endpoint should be named /samples/:id/authorization

test/ProposalAuthorization.js
```diff
@@ -104,7 +114,7 @@ describe("1400: ProposalAuthorization: Test access to proposal", () => {
     done();
   });
 
-  it("0010: adds proposal 1", async () => {
+  it("0000: adds proposal 1", async () => {
```
nitrosx: Can we maintain the old test number?

test/ProposalAuthorization.js
```diff
@@ -120,7 +130,7 @@ describe("1400: ProposalAuthorization: Test access to proposal", () => {
       });
   });
 
-  it("0020: adds proposal 2", async () => {
+  it("0010: adds proposal 2", async () => {
```
nitrosx: Can we maintain the old test number?

test/ProposalAuthorization.js
```diff
@@ -136,7 +146,7 @@ describe("1400: ProposalAuthorization: Test access to proposal", () => {
       });
   });
 
-  it("0030: adds proposal 3", async () => {
+  it("0020: adds proposal 3", async () => {
```
nitrosx: Can we maintain the old test number?

test/ProposalAuthorization.js
```diff
@@ -152,14 +162,38 @@ describe("1400: ProposalAuthorization: Test access to proposal", () => {
       });
   });
 
+  // it("0030: adds proposal 10", async () => {
```
nitrosx: Renumber this test to 0035

test/SampleAuthorization.js
```diff
@@ -485,7 +485,7 @@ describe("2250: Sample Authorization", () => {
       .post("/api/v3/Samples")
       .send(sample)
       .set("Accept", "application/json")
-      .expect(TestData.CreationForbiddenStatusCode);
+      .expect(TestData.UnauthorizedStatusCode);
```
nitrosx: Why are we changing from CreationForbiddenStatusCode to UnauthorizedStatusCode?

test/SampleAuthorization.js
```diff
@@ -498,7 +498,7 @@ describe("2250: Sample Authorization", () => {
...
```
nitrosx: Why are we changing from CreationForbiddenStatusCode to UnauthorizedStatusCode?

test/SampleAuthorization.js
```diff
@@ -511,7 +511,7 @@ describe("2250: Sample Authorization", () => {
...
```
nitrosx: Why are we changing from CreationForbiddenStatusCode to UnauthorizedStatusCode?

test/SampleAuthorization.js
```diff
@@ -524,7 +524,7 @@ describe("2250: Sample Authorization", () => {
...
```
nitrosx: Why are we changing from CreationForbiddenStatusCode to UnauthorizedStatusCode?

test/SampleAuthorization.js
```diff
@@ -977,31 +977,31 @@ describe("2250: Sample Authorization", () => {
       .post("/api/v3/Samples/" + sampleId1 + "/attachments")
       .send(TestData.AttachmentCorrect)
       .set("Accept", "application/json")
-      .expect(TestData.CreationForbiddenStatusCode);
+      .expect(TestData.UnauthorizedStatusCode);
```
nitrosx: Why are we changing from CreationForbiddenStatusCode to UnauthorizedStatusCode?

test/SampleAuthorization.js
```diff
@@ -977,31 +977,31 @@ describe("2250: Sample Authorization", () => {
...
   });
 
   it("0420: adds an attachment as an unauthenticated user to sample 2, which should fail", async () => {
     return request(appUrl)
       .post("/api/v3/Samples/" + sampleId2 + "/attachments")
       .send(TestData.AttachmentCorrect)
       .set("Accept", "application/json")
-      .expect(TestData.CreationForbiddenStatusCode);
+      .expect(TestData.UnauthorizedStatusCode);
```
nitrosx: Why are we changing from CreationForbiddenStatusCode to UnauthorizedStatusCode?

test/SampleAuthorization.js
```diff
@@ -977,31 +977,31 @@ describe("2250: Sample Authorization", () => {
...
   });
 
   it("0430: adds an attachment as an unauthenticated user to sample 3, which should fail", async () => {
     return request(appUrl)
       .post("/api/v3/Samples/" + sampleId3 + "/attachments")
       .send(TestData.AttachmentCorrect)
       .set("Accept", "application/json")
-      .expect(TestData.CreationForbiddenStatusCode);
+      .expect(TestData.UnauthorizedStatusCode);
```
nitrosx: Why are we changing from CreationForbiddenStatusCode to UnauthorizedStatusCode?

test/SampleAuthorization.js
```diff
@@ -977,31 +977,31 @@ describe("2250: Sample Authorization", () => {
...
   });
 
   it("0440: adds an attachment as an unauthenticated user to sample 4, which should fail", async () => {
     return request(appUrl)
       .post("/api/v3/Samples/" + sampleId4 + "/attachments")
       .send(TestData.AttachmentCorrect)
       .set("Accept", "application/json")
-      .expect(TestData.CreationForbiddenStatusCode);
+      .expect(TestData.UnauthorizedStatusCode);
```
nitrosx: Why are we changing from CreationForbiddenStatusCode to UnauthorizedStatusCode?

test/config/pretest.js
```diff
@@ -3,7 +3,8 @@
 
 var chaiHttp = require("chai-http");
 const { MongoClient } = require("mongodb");
-const uri = "mongodb://localhost:27017/scicat";
+const testDbName = process.env.MONGODB_TEST || "scicat-test";
+const uri = `mongodb://localhost:27017/${testDbName}`;
```
nitrosx: I'm not sure what is the correct solution, but I think that have a separate variable to set the dataset in testing will lead to confusion.
I think we should rely only on the value MONGODB_URI. If you are testing locally, you will change the value of MONGODB_URI according in order to use your local database.

nitrosx (changes requested):
Please check my comments and let me know if we need to discuss

test/config/pretest.js
Junjiequan: I think we should try to avoid relying on MONGODB_URI for testing. 
If you by accident, run the test locally without changing the MONGODB_URI the test will wipe off all the local DB data

test/SampleAuthorization.js
Junjiequan: To give you a brief context, we now have two policy guards for APIs, PoliciesGuard and AuthenticatedPoliciesGuard.

PoliciesGuard is used for endpoints that unauthenticated users can access. Permission is checked at the instance scope, and errors are controlled by the permissionChecker function, which throws a Forbidden 403 error if access is denied.

AuthenticatedPoliciesGuard is used for endpoints that only authenticated users can access. Errors are controlled by the guard scope + instance scope, and a Unauthorized 401 error is thrown when an unauthenticated user sends a request to an endpoint decorated with AuthenticatedPoliciesGuard.

So, for endpoints that can be accessed by unauthenticated users, we should use Forbidden 403. For endpoints that require authentication, we should use Unauthorized 401.

nitrosx: Understood.
Maybe a better solution would be to change `TestData.CreationForbiddenStatusCode` to `TestData.CreationUnauthorizedStatusCode`.

Which status is returned if an unauthorized user calls this endpoint? 

What matters is that we keep the status returned consistent.

test/config/pretest.js
nitrosx: I see your point, but by using the two environmental variables, we are forcing developers to change the database name in two places instead of one.
Also I feel that we are not minimizing the risk to wipe a local DB data. It is still very much present.

Junjiequan: I'd suggest to just hardcode test db as `scicat-test-db`  in this file like the appUrl

test/SampleAuthorization.js
Junjiequan: In this case, it will return 401 Unauthorized

nitrosx (changes requested):
Waiting for the changes on pretest

nitrosx (approved):

PR #1337, opened 2024-07-29
feat: enable filters config and scientific conditions on the backend

### LINKED ISSUE #604: Backend development to enable table for user configuration of Facets and columns.
## Issue Name
Backend development to enable table for user configuration of Facets and columns.

### Summary
The user configuration needs to be stored to the main Scicat database so that preferences for what columns can be seen, what facets can be filtered and the order and type of the metadata and scientific metadata sections.

Ingvord:
This one is to resolve #604 

- [x] extend UserSettings document schema: add filters and conditions
- [x] add tests 
- [x] add `default ` endpoints


bpedersen2 (commented):
Looks fine.

src/users/users.controller.ts
```diff
@@ -307,6 +313,42 @@ export class UsersController {
...
+  @UseGuards(
+    class ByPassAuthenticatedPoliciesGuard
+      extends PoliciesGuard
+      implements CanActivate
+    {
+      async canActivate(): Promise<boolean> {
+        return Promise.resolve(true);
+      }
+    },
+  )
```
Junjiequan, line 326: If the endpoint should open to unauthenticated users, we can just use PoliciesGuard or remove UseGuards at all I suppose. Since we are not checking policies here

src/users/schemas/user-settings.schema.ts
```diff
@@ -5,6 +5,43 @@ import { Document } from "mongoose";
...
+}
+
+// Define the Condition interface
+export interface ScientificCondition {
+  field: string;
+  value: string;
+  operator: string;
+}
+
+export const kDefaultFilters: FilterConfig[] = [
```
Junjiequan: k stands for?

Junjiequan (approved):
Looks good to me overall.
Just commented few minor things.

PS: merge with master branch will fix the docker-compose error in the test

src/users/schemas/user-settings.schema.ts
Ingvord: Sorry, old habit - roots from my C++ experience to follow Google code styling [standards ](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables)

`k` - prefix for a static const

I used to maintain JavaScriptMVC (pre-Angular SPA vanilla JS framework), together with the folks we adopted that standard also for JS 

If that makes any sense

Junjiequan: Generally it would be more appreciated if the naming is consistent with existing naming convention within the application scope.

Ingvord:
Sure, what is the current convention?

src/users/schemas/user-settings.schema.ts
Junjiequan: #Sure, what is the current convention?

We don't have a strict naming convention in this project yet, but we should have one documented in the future.
Google also has a code styling standard for [typescript](https://google.github.io/styleguide/tsguide.html#naming) as well, which, generally speaking is like unspoken standard for most of the typescript applications

Ingvord: Thanks! I will check that out!

Ingvord:
resolves #604

Ingvord:
@Junjiequan how can I fix that PR title check? Thanks

Junjiequan:
@Ingvord 
feat: enable filters config and scientific conditions on the backend

The rules for PR title can be found in https://github.com/SciCatProject/scicat-backend-next/blob/master/.github/pr-title-checker-config.json

Ingvord:
Cool! That worked. Thanks!

Also the whole thing looks a bit encryptic to me. Why do we need this? Is it possible to provide more descriptive error message?

Thanks again for the help

Junjiequan:
@Ingvord 
It helps with mainly two things:
1. PR title prefix makes it easy to understand what kind of changes will be included in the PR.
2. Helps with complete & well organized release notes.

I believe the error messages won't matter too much, it just takes a little time to accommodate the title rules.

Ingvord:
That's true if you do the project on a daily basis. Although if one only
rarely contributes something, they will have to accommodate with rules
every time 😆

Issue #1338, opened 2024-07-29
Can not build latest master


Ingvord:
Hi, I have update my BE master branch today and found out I can not build it anymore:

```console
ingvord@<>:~/Projects/scicat/scicat-backend-next$ npm run start

> scicat-backend-next@4.0.0 start
> nest start

src/datasets/datasets.service.ts:295:5 - error TS2740: Type 'ModifyResult<Document<unknown, {}, DatasetDocument> & DatasetClass & Document<any, any, any> & { _id: ObjectId; }>' is missing the following properties from type 'DatasetClass': pid, _id, owner, contactEmail, and 17 more.

295     return await this.datasetModel.findOneAndDelete({ pid: id });
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/users/users.service.ts:257:5 - error TS2740: Type 'ModifyResult<Document<unknown, {}, UserIdentityDocument> & UserIdentity & Document<any, any, any> & { _id: ObjectId; }>' is missing the following properties from type 'UserIdentity': authStrategy, created, credentials, externalId, and 4 more.

257     return removedUserIdentity;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~

Found 2 error(s).

```

Is it a real issue, or am I missing anything?

Thanks

Junjiequan:
Could you try to delete the node_modules folder and then re-install the packages with `npm install`

Ingvord:
Yes, it is the same

bpedersen2:
No problems building here (master @ d42f8c487c00879f3a70d71d082f44fd490c502e) .

Ingvord:
I did a fresh setup and indeed it now builds. However takes a few seconds (15!) to fire up the BE:

![Peek 2024-07-30 20-04](https://github.com/user-attachments/assets/0dad75aa-2fba-4a3f-9598-0f6bcdeb9022)

I have a feeling (sorry no recordings) it was much faster... is it only me? Or is it a regression?

Ingvord:
I guess we can close this

PR #1340, opened 2024-07-29
Fetch correct functional accounts file for tests

Ingvord:
Provide test data with test functional accounts in environment agnostic way


Junjiequan:
I don't understand what is this PR for. Could you explain a little bit why is this change needed?

Ingvord:
@Junjiequan my intention was to load test accounts explicitly in the TestData.js

For that we need to rename `functionalAccount.json.test` to `functionalAccounts.test.json`

Basically that it. However I realized that that file has to be copied around anyway before running the BE. So probably does not make much of a sense...

Is there any config option to specify functional accounts file location for BE?

Junjiequan:
For users runnning the application for the first time they need to `npm run generate:local`, which handles `functionalAccounts.json` file generation and pulls required docker images. 
So, to run the test locally you can choose to `npm run generate:local` or create your own `functionalAccounts.json` in the root directory.

Ingvord:
@Junjiequan Thanks! Good to know!

nitrosx:
@Ingvord we decided to make it harder to use the functional accounts file present in the repo for security reason.
We want site admins or user that deploy scicat to define theirs functional accounts and not use the default ones.
Hopefully this info helps

Junjiequan:
Close this PR as the issue is fixed https://github.com/SciCatProject/scicat-backend-next/issues/1338

Issue #1341, opened 2024-07-29
Validate actions during their creation, store only configVersion in the Job schema


despadam:
Based on the comment under PR #1286 and discussing afterwards with @sofyalaski:

- There is no need for method `validate()` in the action class. Validation should happen inside the constructor of each action. 
- Since we are now using a global job configuration - instead of job specific - we should only store the configuration version inside each job in the db. 
- Before `create`: No need for additional logic. If there is any error while constructing the actions, the application fails. When all actions have been initialized successfully, then each job that gets created will do so by automatically using the current global configuration.
- Before `statusUpdate`: We need to check if the `configVersion` is the same. If no changes were made to the configuration, then we can proceed directly to performing the update. But if the configuration has changed, would we need to update the value that is stored in the db? Note that when updating the configuration, the application needs to get restarted, so the actions have already been reconstructed automatically with the new configuration values.

sbliven:
I'm back from my vacation and just seeing this.

Validation of the job config file needs to happen in the constructor, as you say.

The reason for the validate() method on actions was to allow validation of `CreateJobDto.jobParams`. For instance, one could imagine a job where users have to supply an authentication token when creating a job, which then gets validated by URLAction before making the request.

None of the MVP actions used jobParams (other than the datasetIds list) so including validate() in the abstract interface was more about future-proofing the API than from an immediate need. But if we're removing validate() then we should probably remove `jobParams` itself (making datasetIds a top-level field as proposed elsewhere).

sofyalaski:
@sbliven  I was thinking about that validate() function when I stumbled upon this enum [JobsConfigSchema](https://github.com/SciCatProject/scicat-backend-next/blob/release-jobs/src/jobs/types/jobs-config-schema.enum.ts). According to this enum, job wouldn't have the `jobParams`, but only `datasetIds`, so essentially the case you propose. However, in the [dto](https://github.com/SciCatProject/scicat-backend-next/blob/release-jobs/src/jobs/dto/create-job.dto.ts) the `jobParams` is the required parameter. We should make it consistent. We could introduce another enum for job parameters, that would include datasetIds but I don't know what else this enum should include for now

despadam:
`configVersion` in the job schema: has been resolved by PRs #1364 and #1395 

TODO: validation action

sbliven:
`validate()` restored in #1421

PR #1354, opened 2024-08-05
Validate actions during their creation

despadam:
## Description

As seen in issue #1341 

## Motivation

Method `validate()` in the `JobAction` class was never used. All validation was done inside the action's constructor.

## Changes:

- Removed method `validate()` from the `JobAction` class and all its subclasses.
- There is no longer the need to validate the actions before creating a job. This part has been commented out in the interceptor.
- In fact, there is no longer the need to use `job-create.interceptor` at all. It has been removed from `create` in the `jobs.controller`. The reason for this is that the interceptor was doing two things: Apart from action validation which has already been removed, it was also adding the `configuration` field in the job instance to be created. However this part is already covered by the `instanceAuthorizationJobCreate` function in  `jobs.controller`.
- `checkAuthenticatedUser` has been removed from `jobs.controller`, as it was never used.
- Before `statusUpdate` we check whether the configuration file has been updated by comparing the new with the one stored inside the job. However, even if they are different, we still proceed with the update as usual, as all actions have been automatically re-initialized with the new configuration. _This is a point to be discussed later._

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -178,5 +178,5 @@ jobs:
         run: |
           cp CI/ESS/docker-compose.api.yaml docker-compose.yaml
           cp functionalAccounts.json.test functionalAccounts.json
-          docker-compose up --build -d
+          docker compose up --build -d
           npm run test:api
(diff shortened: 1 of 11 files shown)


sofyalaski (approved):
looks good!

PR #1364, opened 2024-08-07
Store configVersion in the job schema

despadam:
## Description

As seen in issue #1341

## Motivation

Since we are now using a global job configuration - instead of job specific - we should only store the configuration version inside each job in the db.

## Changes:

- `job.schema` was updated to only include `configVersion`
- In `jobs.controller`: `instanceAuthorizationJobCreate` and was updated to store `configVersion` in `jobInstance`

## Tests:

Tests should be updated to conform with the new job schema and retrieve the configuration from elsewhere.


src/jobs/jobs.controller.ts
```diff
@@ -749,11 +737,11 @@ export class JobsController {
...
       request.user as JWTUser,
     );
-    const canCreate =
+    const canGet =
       ability.can(AuthOp.JobReadAny, JobClass) ||
       ability.can(AuthOp.JobReadAccess, currentJobInstance);
-    if (!canCreate) {
-      throw new ForbiddenException("Unauthorized to update this dataset");
+    if (!canGet) {
+      throw new ForbiddenException("Unauthorized to get this job.");
```
Junjiequan:
```
      const canView =
        ability.can(Action.DatasetReadAny, DatasetClass) ||
        ability.can(Action.DatasetReadOneOwner, datasetInstance) ||
        ability.can(Action.DatasetReadOneAccess, datasetInstance) ||
        ability.can(Action.DatasetReadOnePublic, datasetInstance);

      if (!canView) {
      }
```
I suggest we either follow the variable namings for can ability from the `datasets.controller.ts`, or just to change all namings with canRead, canCreate. Having a vairious naming for the samething makes confusion sometimes.

Junjiequan (dismissed):

src/jobs/jobs.controller.ts
despadam: Covered in commit 4aedbc27d5667476a6d87cef1e06194d2be1b6df

src/jobs/jobs.controller.ts
```diff
@@ -749,11 +737,11 @@ export class JobsController {
     const ability = this.caslAbilityFactory.createForUser(
       request.user as JWTUser,
     );
-    const canCreate =
+    const canRead =
```
bpedersen2, line 748: some test fail with permission problems, i suspect ti could be related to this change here

despadam: The latest commit fixes the failing tests

PR #1375, opened 2024-08-16
refactor(casl-ability): split factory function

sofyalaski:
## Description

This PR splits a one single function defined in casl-ability.factory.ts into logical parts.

## Motivation

* Readability/maintainability 
* Later will add changes to the one function specific to jobs (data instance authorization) in release-jobs. Now the access there is granted based on the configuration passed as Object's property. Since we don't want to store the whole configuration this could be a work-around. 

## Changes:

* I restructured casl function such that each endpoint gets a separate function for endpoint access and a data instance access. Functions for endpoint access are called `NameEnpointAccess` and the ones for data instance access are called `NameDataInstanceAccess`. Not all of the endpoints have this function, since there were no use cases ( for instance users).
* The endpoint functions are the ones that are called in the `@CheckPolicies` decorator when user starts to create any operation. To still be able to use this decorator, I implemented this function in casl: `endpointAccess`. It takes two arguments: the JWTUser and the string containing the endpoint name. Based on that a correct function for endpoint authentication will be called. (The user-identities.controller.ts uses the rules defined for user, so in the new casl function `endpointAccess` the one for users will be called too. )
* `policies.guard.ts` now calls `endpointAccess` instead of the `createForUser`. The second argument (endpoint name) is extracted from the request. 
* To check the data instance authorization the casl function was usually called one more time. Here, just the one defined specifically for this endpoint will be called.
 
## Issues

* While I was restructuring I saw some line that is never used: 
https://github.com/SciCatProject/scicat-backend-next/blob/b218952495b36d836c155e0ec0e7af9e3baf0832/src/casl/casl-ability.factory.ts#L219
* "ListAll" and "ListOwn" rules for polices don't exist in casl: 
https://github.com/SciCatProject/scicat-backend-next/blob/b218952495b36d836c155e0ec0e7af9e3baf0832/src/policies/policies.controller.ts#L88-L89. I think it was never obvious because the tests that exist for policies don't cover the GET operation, but only GET/:id, which doesn't check these rules. 

## Tests included

## Documentation

### DIFF
--- a/src/casl/casl-ability.factory.ts
+++ b/src/casl/casl-ability.factory.ts
@@ -56,119 +56,129 @@ export type AppAbility = MongoAbility<PossibleAbilities, Conditions>;
 export class CaslAbilityFactory {
   accessEndpointForUser(user: JWTUser, subject: string) {
     switch (subject) {
-      case 'elastic-search':
-        return this.accessElasticSearchEndpointForUser(user);
-      case 'logbooks':
-        return this.accessLogbookEndpointForUser(user);
-      case 'publisheddata':
-        return this.accessPublishedDataEndpointForUser(user);
-      case 'policies':
-        return this.accessPolicyEndpointForUser(user);
-      case 'users':
-        return this.accessUserEndpointForUser(user);
-      case 'datasets':
+      case "datasets":
         return this.accessDatasetEndpointForUser(user);
-      case 'origdatablocks':
-        return this.accessOrigDatablockEndpointForUser(user);
-      case 'jobs':
+      case "elastic-search":
+        return this.accessElasticSearchEndpointForUser(user);
+      case "jobs":
         return this.accessJobsEnpointForUser(user);
-      case 'proposals':
-        return this.accessProposalsEndpointForUser(user);
-      case 'samples':
-        return this.accessSamplesEndpointForUser(user);
-      case 'instruments':
+      case "instruments":
         return this.accessInstrumentEndpointForUser(user);
+      case "logbooks":
+        return this.accessLogbookEndpointForUser(user);
+      case "origdatablocks":
+        return this.accessOrigDatablockEndpointForUser(user);
+      case "policies":
+        return this.accessPolicyEndpointForUser(user);
+      case "proposals":
+        return this.accessProposalsEndpointForUser(user);
+      case "publisheddata":
+        return this.accessPublishedDataEndpointForUser(user);
+      case "samples":
+        return this.accessSamplesEndpointForUser(user);
+      case "users":
+        return this.accessUserEndpointForUser(user);
+      case "useridentities":
+        return this.accessUserEndpointForUser(user);
       default:
-        throw new Error(`No endpoint access policies defined for subject: ${subject}`);
+        throw new Error(`
+        No endpoint access policies defined for subject: ${subject}`);
     }
   }
 
-  accessElasticSearchEndpointForUser(user:JWTUser){
-    const { can, cannot, build } = new AbilityBuilder(
+  accessElasticSearchEndpointForUser(user: JWTUser) {
+    const { can, build } = new AbilityBuilder(
       createMongoAbility<PossibleAbilities, Conditions>,
     );
 
-    if (user && 
-        user.currentGroups.some((g) => configuration().adminGroups.includes(g))
-      ) {
-        /*
+    if (
+      user &&
+      user.currentGroups.some((g) => configuration().adminGroups.includes(g))
+    ) {
+      /*
         / user that belongs to any of the group listed in ADMIN_GROUPS
         */
-        can(Action.Manage, ElasticSearchActions);
-      }
+      can(Action.Manage, ElasticSearchActions);
+    }
     return build({
       detectSubjectType: (item) =>
         item.constructor as ExtractSubjectType<Subjects>,
     });
   }
 
-  accessLogbookEndpointForUser(user:JWTUser){
-    const { can, cannot, build } = new AbilityBuilder(
+  accessLogbookEndpointForUser(user: JWTUser) {
+    const { can, build } = new AbilityBuilder(
       createMongoAbility<PossibleAbilities, Conditions>,
     );
 
-    if (user
-      ) {
-        /*
+    if (user) {
+      /*
         / authenticated user
         */
-       can(Action.Read, Logbook);
-      }
+      can(Action.Read, Logbook);
+    }
     return build({
       detectSubjectType: (item) =>
         item.constructor as ExtractSubjectType<Subjects>,
     });
   }
-    
-  accessPublishedDataEndpointForUser(user:JWTUser){
-    const { can, cannot, build } = new AbilityBuilder(
+
+  accessPublishedDataEndpointForUser(user: JWTUser) {
+    const { can, build } = new AbilityBuilder(
       createMongoAbility<PossibleAbilities, Conditions>,
     );
-    if (user){
+    if (user) {
       can(Action.Read, PublishedData);
       can(Action.Update, PublishedData);
       can(Action.Create, PublishedData);
     }
 
-    if (user && user.currentGroups.some((g) => configuration().deleteGroups.includes(g))){
-        /*
+    if (
+      user &&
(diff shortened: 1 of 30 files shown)


sbliven (approved):
Seems like a good refactor to me!

@nitrosx should probably review as well regarding ListAll

bpedersen2 (commented):
LGTM

src/casl/casl-ability.factory.ts
```diff
@@ -54,40 +54,199 @@ export type AppAbility = MongoAbility<PossibleAbilities, Conditions>;
...
 export class CaslAbilityFactory {
-  createForUser(user: JWTUser) {
+  accessEndpointForUser(user: JWTUser, subject: string) {
+    switch (subject) {
+      case "datasets":
+        return this.accessDatasetEndpointForUser(user);
+      case "elastic-search":
+        return this.accessElasticSearchEndpointForUser(user);
+      case "jobs":
+        return this.accessJobsEnpointForUser(user);
```
Junjiequan: Typo here, accessJobsEndpointForUser

src/casl/casl-ability.factory.ts
```diff
@@ -744,51 +780,225 @@ export class CaslAbilityFactory {
...
+        isPublished: true,
+      });
+    }
+    return build({
+      detectSubjectType: (item) =>
+        item.constructor as ExtractSubjectType<Subjects>,
+    });
+  }
 
+  accessJobsEnpointForUser(user: JWTUser) {
```
Junjiequan: Typo here accessJobsEndpointForUser

src/casl/casl-ability.factory.ts
```diff
@@ -54,40 +54,199 @@ export type AppAbility = MongoAbility<PossibleAbilities, Conditions>;
...
+      case "proposals":
+        return this.accessProposalsEndpointForUser(user);
+      case "publisheddata":
+        return this.accessPublishedDataEndpointForUser(user);
+      case "samples":
+        return this.accessSamplesEndpointForUser(user);
+      case "users":
+        return this.accessUserEndpointForUser(user);
+      case "useridentities":
+        return this.accessUserEndpointForUser(user);
```
Junjiequan: We can avoid duplication here by 
```suggestion
      case "users":
      case "useridentities":
        return this.accessUserEndpointForUser(user);
```

src/casl/casl-ability.factory.ts
```diff
@@ -54,40 +54,199 @@ export type AppAbility = MongoAbility<PossibleAbilities, Conditions>;
...
 @Injectable()
 export class CaslAbilityFactory {
-  createForUser(user: JWTUser) {
+  accessEndpointForUser(user: JWTUser, subject: string) {
+    switch (subject) {
+      case "datasets":
+        return this.accessDatasetEndpointForUser(user);
+      case "elastic-search":
+        return this.accessElasticSearchEndpointForUser(user);
+      case "jobs":
```
Junjiequan: I think the naming can be simplified as `endpointAccess`, `datasetEndpointAccess` etc, which to me is clear enough. 
What do you think?

Junjiequan (changes requested):
Great job, thanks for the efforts!
Just commend on few minor issues.

src/casl/casl-ability.factory.ts
sofyalaski: yes, that might be better :)

src/casl/casl-ability.factory.ts
sofyalaski: oh thanks for noticing!

Junjiequan (approved):
Looks good to me.
Let's see what @nitrosx says about the ListAll policy.

src/casl/guards/policies.guard.ts
```diff
@@ -20,14 +20,15 @@ export class PoliciesGuard implements CanActivate {
 
     const req = context.switchToHttp().getRequest();
     const user = req.user;
-    const ability = this.caslAbilityFactory.createForUser(user);
+    const endpoint = req.route.path.split("/")[3];
```
nitrosx: I worry that this solution is a little bit to hard coded. What if we have routes that have different structure and the endpoint/subsystem is no tin position 3?

src/casl/casl-ability.factory.ts
```diff
@@ -155,54 +121,740 @@ export class CaslAbilityFactory {
...
+      can(Action.UserUpdateOwn, User, { _id: user._id });
+      can(Action.UserDeleteOwn, User, { _id: user._id });
+    }
+    return build({
+      detectSubjectType: (item) =>
+        item.constructor as ExtractSubjectType<Subjects>,
+    });
+  }
+
+  datasetDataInstanceAccess(user: JWTUser) {
```
nitrosx: Can we change the name to datasetInstanceAccess?
This would  make it consistent with terminology that we use every where else.

src/casl/casl-ability.factory.ts
```diff
@@ -54,40 +54,47 @@ export type AppAbility = MongoAbility<PossibleAbilities, Conditions>;
...
+        return this.policyEndpointAccess(user);
+      case "proposals":
+        return this.proposalsEndpointAccess(user);
+      case "publisheddata":
+        return this.publishedDataEndpointAccess(user);
+      case "samples":
+        return this.samplesEndpointAccess(user);
+      case "users":
+      case "useridentities":
+        return this.userEndpointAccess(user);
```
nitrosx: Do we really need to have this switch statement?
Can we load all the endpointAccess loaded?

nitrosx (changes requested):
Great job!!!
Minor changes on the names and feedback on the way you extract the endpoint.

nitrosx:
I will check regarding tomorrow regarding the authorization that are missing.

src/casl/casl-ability.factory.ts
sofyalaski: So yes, regarding this and the previous comment about the way the endpoint is extracted, I was also worried that it might break something in future. The alternative would be to unite the endpoint functions back into one, eliminating the switch statement because it is then based on the user only again. This is then not as granular as now, but at the same time the endpoint functions are also not that long/complex.

src/casl/guards/policies.guard.ts
sofyalaski: One option would be to unite the endpoint functions at least back into one, s.t the rules are based only on the user again. The second option would be to pass a second argument to the @CheckPolicies decorator, based on which the correct function for endpoint will be called.

src/casl/casl-ability.factory.ts
nitrosx: @sofyalaski what do you think is the best to move forward?

nitrosx:
@sofyalaski regarding the issues that you have listed above.
When you check if an action can be performed on a data class or data instance, like
`ability.can(AuthOp.ListAll, Policy)`
if such policy is not defined in the ability factory, it will be a `cannot` by default and return `false` when can is called.
I hope this answer your questions and clears your doubts.

src/casl/casl-ability.factory.ts
sofyalaski: I added a new commit, it now changes the decorator for @CheckPolices to accept two arguments, the first being the keyword, according to which the function will be sent to the correct endpoint function. What do you think?

sofyalaski:
Regarding the policies, isn't then the `updateMergedFiltersForList` function not needed anymore? since both abilities evaluate to false, as they are not defined the `mergedFilters` are never changed in this function.

nitrosx:
@sofyalaski most likely the `updateMergedFiltersForList` has not been updated since it was introduced. Most likely the tests do not cover it also.
Please refer to the same function that is present in datasets controller if you want to try to update it.
https://github.com/SciCatProject/scicat-backend-next/blob/9216dea598a9ad05ab06b0c8917b950cd0a30f01/src/datasets/datasets.controller.ts#L157

Leave the code that you presented in the `issues` section of the PR as it is, and let's focus to merge the PR. 
Please open an issues regarding policies endpoint and its authorization with the reference to the code that you have in this PR. We should also open another issue to review and update the tests for policies.

nitrosx (approved):

PR #1387, opened 2024-08-20
chore: upgrade from node 18 to 20

fpotier:
## Description

Upgrade to node 20

## Motivation

Node 18 end of life is coming in a few months.


fpotier:
Are the files in `CI` still used? Should I also bump the node version there or let each facility take care of it?

bpedersen2 (approved):

PR #1394, opened 2024-08-23
fix: fixed permissions on user identities and improved swagger documentation

nitrosx:
## Description
fixed permissions on user identities and improved swagger documentation

## Motivation
E2E testing was failing due to wrong permissions on user identities endpoints.

## Fixes:
Please provide a list of the fixes implemented by this PR
* User Identities controller
* common utils

## Tests included

## Documentation

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -925,19 +925,28 @@ export const samplesFullQueryDescriptionFields =
 }\n \
   </pre>';
 
-export const filterUserIdentityExample = '{ "where": { "field": "value" }}';
+export const filterUserIdentityExample =
+  '{ "profile.email": "this_email@your.site" }';
 
 export const filterUserIdentityDescription =
   '<pre>\n \
-{\n \
-  "field": "value"\n \
-}\n \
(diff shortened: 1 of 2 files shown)


src/common/utils.ts
```diff
@@ -925,6 +925,21 @@ export const samplesFullQueryDescriptionFields =
...
+  "field": "value"\n \
+}\n \
+  or \n \
+{\n \
+  "where?": {\n \
+    "field": "value"\n \
+  }\n \
+}\
+</pre>';
+
```
Junjiequan: It's useful to allow both a key-value pair like "profile.email":"value" and just "value" without a key to work interchangeably. 

Since this endpoint is meant for validating emails, using "profile.email":"value" instead of "field":"value" would be clearer for users. 

Additionally, the "where?":{"field":"value"} format is currently only used by the frontend, and should be deprecated for this endpoint in the future, as it can be confusing.

Junjiequan (approved):
Looks good in general.
Just one minor comment

PR #1404, opened 2024-09-02
refactor: separate search ui logic from frontend

Junjiequan:
## Description

This PR aims to separate the implementation of default user settings logic from FE.

## Motivation

Reduce maintenance costs for the future.

## Changes:

- Updated `frontend.config` to include new labelMaps and reorganize filter settings.
- Renamed and moved localColumns from `frontend.config`  into defaultDatasetsListSettings field.
- Removed unused interceptors for user settings.
- Added validation for filter components in findByIdUserSettings to check for valid filters based on FilterComponentType, and reset if invalid filters are found during validation.
- Adjusted findOneAndUpdateUserSettings to ensure proper upsert and default behavior

## Tests included

## Documentation


src/users/schemas/user-settings.schema.ts
```diff
@@ -74,12 +71,12 @@ export class UserSettings {
...
-    default: FILTER_CONFIGS,
+    default: [],
     description: "Array of filters the user has set",
   })
   @Prop({
     type: [{ type: Object }],
-    default: FILTER_CONFIGS,
+    default: [],
   })
   filters: FilterConfig[];
```
Junjiequan: FilterConfig needs to be changed.
It should be similar to:
```suggestion
export type FilterComponentType =
  | "LocationFilterComponent"
  | "PidFilterComponent"
  | "PidFilterContainsComponent"
  | "PidFilterStartsWithComponent"
  | "GroupFilterComponent"
  | "TypeFilterComponent"
  | "KeywordFilterComponent"
  | "DateRangeFilterComponent"
  | "TextFilterComponent";

// Define the Filter interface
export interface FilterConfig {
  FilterComponentType: string;
  visible: boolean;
}
```

src/users/users.controller.ts
```diff
@@ -100,6 +94,7 @@ export class UsersController {
 
   @ApiBody({ type: CredentialsDto })
   @AllowAny()
+  @UseInterceptors(UserSettingsInterceptor)
```
nitrosx: I do not think we need this interceptor on login endpoint.

src/users/users.service.ts
```diff
@@ -273,16 +274,27 @@ export class UsersService implements OnModuleInit {
   }
 
   async findByIdUserSettings(userId: string): Promise<UserSettings | null> {
```
nitrosx, line 275: _findUserSettingsById_ is a better name

src/users/interceptors/user-settings.interceptor.ts
nitrosx, line 1: Do we need this interceptor?

nitrosx (changes requested):

src/users/users.service.ts
nitrosx: The renaming of the internal functions needs more brainstorming.
It requires more effort in order to keep naming consistent.

src/users/dto/update-user-settings.dto.ts
```diff
@@ -18,13 +10,14 @@ export class UpdateUserSettingsDto {
...
-  @IsArray()
-  readonly conditions: ScientificCondition[];
+  @ApiProperty({
+    type: "object",
+    additionalProperties: { type: "array", items: {} },
+    required: false,
+  })
+  @IsOptional()
+  @IsObject()
+  readonly frontendSettings?: Record<string, unknown[]>;
```
nitrosx: On second thought, can we call them just `extraSettings`?

nitrosx (changes requested):
Minor changes.

src/users/dto/update-user-settings.dto.ts
Junjiequan: extraSettings are too vague. I strongly against naming it as extraSettings.

src/users/users.controller.ts
Junjiequan: It's removed in the following commit

src/users/dto/update-user-settings.dto.ts
Junjiequan: ExtraSettings seems a bit too vague. I would kindly suggest reconsidering the name to something more specific.

src/users/interceptors/user-settings.interceptor.ts
Junjiequan: It's removed in the following commit.

nitrosx (approved):

Issue #1388, opened 2024-08-21
jobs api: jobResultObject missing in jobs-release


phwissmann:
### Summary
jobResultObject (or an alternative) is missing in the `jobs-release` branch. In `main` this was for example used to report urls for downloading retrieved datablocks:

```
  "jobResultObject": {
    "result": [
      {
        "datasetId": "20.500.11935/2f919991-2197-41bc-9edd-cbbecc2339ce",
        "name": "2f919991-2197-41bc-9edd-cbbecc2339ce_0_2024-07-12-23-47-53.tar",
        "size": 572897,
        "archiveId": "/STANDARD1S/oneCopySmall/p18724/raw/2024/07/12/20.500.11935/2f919991-2197-41bc-9edd-cbbecc2339ce_0_2024-07-12-23-47-53.tar",
        "url": "https://object.cscs.ch/v1/AUTH_ba83c1e328974db299bef2291c21a770/noderedp/828b948e4443242a514791ff8764cea2/2f919991-2197-41bc-9edd-cbbecc2339ce_0_2024-07-12-23-47-53.tar?temp_url_sig=5792fc444b4531ee50d7b4ed5065b0f8395b7486&temp_url_expires=1724658421"
      }
    ]
  },
```

The code change seemed to have happened when changing `update-job.dto` to `status-update-job.dto`, 
see
https://github.com/SciCatProject/scicat-backend-next/blob/release-jobs/src/jobs/dto/status-update-job.dto.ts
vs
https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/dto/update-job.dto.ts

sbliven:
Does this need to update the job in the database, or does the result just get passed through to the actions triggered by the status update?

sbliven:
After discussion, we believe that this should be included in the job.schema (thus persisted in the database) and the jobStatusUpdate.dto (allowing it to be set by the archive system)

despadam:
Solved by https://github.com/SciCatProject/scicat-backend-next/pull/1410

PR #1410, opened 2024-09-03
Add jobResultObject field

despadam:
## Motivation

#1388 

## Description

A `jobResultObject` field should be included in the `job.schema`, thus persisted in the database. The field contains necessary information for the location of a dataset.

Upon job creation the field should be empty. The value of the field is the result of a dataset archival, meaning we can only have the value after this is successfully executed.

This means that the value can only be provided via the `StatusUpdateJobDto` coming from the archive system.

## Changes:

- Added a non required field `jobResultObject` in the `job.schema`
- Added an optional field `jobResultObject` in the `StatusUpdateJobDto`
- Added the new field in `addStatusFields()` function in `util.ts`
- Passed the extra argument when `addStatusFields()` is called in `jobs.service`. On `create` the field is empty. On `statusUpdate` we pass the value coming from the `StatusUpdateJobDto`.


src/jobs/jobs.service.ts
```diff
@@ -112,6 +113,7 @@ export class JobsService {
           ),
           statusUpdateJobDto.statusCode,
           statusUpdateJobDto.statusMessage!,
+          statusUpdateJobDto.jobResultObject!,
```
Junjiequan: In the StatusUpdateJobDto, both statusMessage and jobResultObject are marked as optional fields, but they are being treated as required fields with the exclamation mark. I’m not sure if I missed something, or if this is a mistake

Junjiequan (dismissed):

src/jobs/jobs.service.ts
despadam: Thanks! `statusMessage` should be required and `jobResultObject` should be optional. Fixed it the latest commit (https://github.com/SciCatProject/scicat-backend-next/pull/1410/commits/c2af58544f3e8b30fae3664aff7f72a6eed35cf9)

Junjiequan (approved):
Looks fine to me.

sbliven (approved):
I tested this out and everything looks good to me!

Issue #1101, opened 2024-03-04
Upgrade API version to v4


nitrosx:
Migration to v4 has been completed for a while.
The community has decided to upgrade the BE URL from `api/v3` to `api/v4`.
API version in Swagger should be consistent.

sbliven:
I thought we decided not to do this because there aren't actually any breaking API changes (#760)

sbliven:
We should do this now that there are clear breaking changes.

sbliven:
Martin gave a nice demo of the work on datasets.v4.controller.ts which utilizes nest's [versioning](https://docs.nestjs.com/techniques/versioning#multiple-versions) capabilities to provide different /api/v3 and /api/v4 implementations. Backwards-compatible controllers can even specify `version: VERSION_NEUTRAL` to provide both. Maybe we should add version annotations to all controllers.

PR #1414, opened 2024-09-05
refactor: prep for unified dataset schema

nitrosx:
## Description
This PR rename the current datasets DTOs and schemas and introduce a new unified dataset schema/dto.

## Motivation
For historical reason, datasets were implemented with two different schemas depending if they were of type raw or derived.
This design introduced complexity in code and increased maintenance cost. 
After 2 years of BE v4.x, we came to the conclusion that such differentiation is not needed and datasets can be described with a single schema and dto.

## Fixes:
- reviewed dataset schema
- renamed current dto to obsolete

## Changes:
- added new unified dto

## Tests included
- [x] No test changes needed

## Documentation


src/datasets/dto/update-dataset.dto.ts
```diff
@@ -211,7 +208,7 @@ export class UpdateDatasetDto extends OwnableDto {
   })
   @IsOptional()
   @IsBoolean()
-  readonly isPublished?: boolean;
+  readonly isPublished?: boolean = false;
```
Junjiequan: It's already set default to false on line 206

src/datasets/dto/update-dataset.dto.ts
```diff
@@ -292,14 +289,134 @@ export class UpdateDatasetDto extends OwnableDto {
...
+    type: String,
+    required: false,
+    description:
+      "ID of the instrument or instruments where the data included or used in this datasets was collected on.",
+  })
+  @IsOptional()
+  @IsString({
+    each: false,
+  })
+  readonly instrumentId?: string[];
```
Junjiequan: Should it be array of string or just string

src/datasets/dto/update-dataset.dto.ts
```diff
@@ -292,14 +289,134 @@ export class UpdateDatasetDto extends OwnableDto {
...
+    type: [String],
+    required: true,
+    description:
+      "Array of input dataset identifiers used in producing the derived dataset. Ideally these are the global identifier to existing datasets inside this or federated data catalogs.",
+  })
+  @IsOptional()
+  @IsString({
+    each: false,
+  })
+  readonly inputDatasets?: string[];
```
Junjiequan: why `each:false` here?
shouldn't it validate each items to make sure they are all string?

src/datasets/schemas/dataset.schema.ts
```diff
@@ -453,44 +451,47 @@ export class DatasetClass extends OwnableClass {
...
+    type: [String],
     required: false,
-    description: "ID of the instrument where the data was created.",
+    description:
+      "Id of the instrument or array of IDS of the instruments where the data contained in this dataset was created/acquired.",
   })
-  @Prop({ type: String, ref: "Instrument", required: false })
-  instrumentId?: string;
+  @Prop({ type: [String], ref: "Instrument", required: false })
+  instrumentId?: string[];
```
Junjiequan: We need to change the naming to `sampleIds` and `instrumentIds` to avoid confusion

src/datasets/dto/output-dataset.dto.ts
```diff
@@ -0,0 +1,41 @@
...
+  updatedBy: string;
+
+  @ApiProperty({
+    type: Date,
+    required: true,
+    description:
+      "Date and time when this record was created. This field is managed by mongoose with through the timestamp settings. The field should be a string containing a date in ISO 8601 format (2024-02-27T12:26:57.313Z)",
+  })
+  @IsString()
+  createdAt: Date;
```
Junjiequan: Maybe we need `@IsDateString()` here as well for consistent Date format?

src/datasets/dto/output-dataset.dto.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  @ApiProperty({
+    type: Date,
+    required: true,
+    description:
+      "Date and time when this record was updated last. This field is managed by mongoose with through the timestamp settings. The field should be a string containing a date in ISO 8601 format (2024-02-27T12:26:57.313Z)",
+  })
+  @IsString()
+  updatedAt: Date;
+}
```
Junjiequan: Maybe we need `@IsDateString()` here as well for consistent Date format?

Junjiequan (changes requested):
Left few comments

sbliven:
This is a pretty major breaking change (but a good one). I think we should change the prefix to `v4` (#1101)

sbliven:
Also note that the change should be documented clearly in the migration page

src/datasets/dto/update-dataset.dto.ts
nitrosx: You are right

src/datasets/dto/update-dataset.dto.ts
nitrosx: There might be some datasets that are produced on multiple instruments or a the results of data analysis/reduction on data collected on the same sample but on multiple instruments.
Array of strings allows that

nitrosx: The frontend will not allow multiple instruments,  yet

src/datasets/dto/update-dataset.dto.ts
nitrosx: FIxed

src/datasets/schemas/dataset.schema.ts
nitrosx: Done for all the relevant fields

src/datasets/dto/output-dataset.dto.ts
nitrosx: Followed your suggestion and changed to `@isDateString`

src/datasets/dto/output-dataset.dto.ts
nitrosx: Done. Converted `createdAt` and `updatedAt` to decorator `@IsDateString()`

src/datasets/datasets.controller.ts
```diff
@@ -61,7 +61,7 @@ import { DatablocksService } from "src/datablocks/datablocks.service";
 import { Datablock } from "src/datablocks/schemas/datablock.schema";
 import { CreateDatablockDto } from "src/datablocks/dto/create-datablock.dto";
 import { PartialUpdateDatablockDto } from "src/datablocks/dto/update-datablock.dto";
-import { UpdateQuery } from "mongoose";
+import { Document, UpdateQuery } from "mongoose";
```
Junjiequan: Lint error - 'Document' is defined but never used

src/datasets/datasets.controller.ts
```diff
@@ -2087,7 +2244,7 @@ export class DatasetsController {
       Action.DatasetLogbookRead,
     );
 
-    const proposalId = dataset?.proposalId;
+    const proposalId = (dataset?.proposalIds || [])[0];
```
Junjiequan, line 2247: It means the sequence of proposalIds is crucial to retrieve correct logbook messages. 
Do we have control over arbitrary proposalIds update with dataset PATCH or PUT methods?

src/datasets/dto/update-dataset.dto.ts
```diff
@@ -292,14 +290,153 @@ export class UpdateDatasetDto extends OwnableDto {
...
```
Junjiequan: It's still `each: false`

src/datasets/schemas/dataset.schema.ts
Junjiequan, line 1: Some lint warnings from unused imports

Junjiequan (changes requested):
Left few comments

src/datasets/datasets.controller.ts
nitrosx: For now, there is only one proposal per dataset, so the problem does not exists.
When we refactor the dataset endpoints, we will brainstorm and think about what to do.

nitrosx: ...but yes the order of proposals will matter.

src/datasets/dto/update-dataset.dto.ts
nitrosx: fixed!!!

src/datasets/datasets.controller.ts
nitrosx: Fixed

src/datasets/schemas/dataset.schema.ts
nitrosx: fixed most of them

Junjiequan (approved):

Issue #1120, opened 2024-03-18
Parse config files as YAML


sbliven:
Since YAML is a strict superset of JSON, we can accept YAML for all config files in a fully backwards-compatible manner. This is also fully supported by NestJS.

1. Add `js-yaml` and `@types/js-yaml` packages
2. Replace `json.load` with `yaml.load`
3. Accept `.json`, `.yaml`, and `.yml` file extensions for the config files (stopping with the first one found)

bpedersen2:
Good idea! :+1:

despadam:
As seen also in the comment here: https://github.com/SciCatProject/scicat-backend-next/issues/841#issuecomment-2391485867

sbliven:
jobConfig implements this in #1463.

Should I keep this open to do the same for functionalAccounts.json?

PR #1421, opened 2024-09-10
feat(jobs): ValidateAction

sbliven:
## Description

Add ValidateAction to check Job DTO formatting

## Motivation

When creating a job clients can provide an untyped `jobParams` object in the DTO, or a `jobResultObject` when updating. These were previously not validated.

## Fixes:
This PR adds a new `validate` job action which can test for the presence of particular attributes in the DTO. Currently the attributes are specified using a [JSONPath](https://github.com/JSONPath-Plus/JSONPath) and are only verified to be non-null. If needed, future implementations could add JSONSchema support to do deeper validation.

For example, the following `jobConfig.json` section would configure the `public` job to require at least one datasetId:
```
{
      "jobType": "public",
      "create": {
        "auth": "#all",
        "actions": [
          {
            "actionType": "validate",
            "required": ["jobParams.datasetIds[0]"]
          }
        ]
      },
      "statusUpdate": {
        "auth": "#all"
      }
    }
```

## Changes:
- Implement `validate` action
- New dependency: jsonpath-plus
- Add tests
- Add example configuration

## Tests included

## Documentation


sbliven:
What does everyone think of the use of JSONPath?

Pro: Simple syntax for simple reference, while keeping enough power for more complex structures like arrays
Pro: Well-known standard (eg used by `jq` command)
Con: Adds a dependency
Con: Only checks element presence, not type

A more powerful option would be to validate against a JSON Schema. This would allow type checking. However I think it would be much more verbose.

sofyalaski:
I think the JSONpath solution is appropriate for now. 
However, there is a minor issue in the jobs.controller: `performJobCreateAction` now accesses only the `jobInstance` of `JobClass` as an argument (this also needed to call the other function that performs the actions).  `performJobCreateAction` should take two arguments, the second being of `JobDto` type so that we can call the `validate` function. Now in the controller `validate` is called on `JobClass` instance, which is not how `validate` is defined.

sbliven:
Good catch, @sofyalaski. Why does this even compile?

sbliven:
Ah, it compiles because JobClass is structurally compatible with CreateJobDto. Working on a fix now.

sbliven:
This implementation has a couple problems.

First, the JSONPath solution doesn't perform any type checking. It would be nice to do something like
```
{
  "actionType": "validate",
  "required": {
    "jobParams.datasetIds": "object",
    "jobParams.datasetIds[*]": "string"
  }
}
```
This looks fine for simple types (string, number, list of literal values) but could quickly grow in complexity until it reaches a full type system. One solution would be to supply a JSON Schema to validate the DTO (which provides full type support). However including JSON schemas in the config file seems overly complex for operators. Is there a happy medium?

Second, the validate method currently acts against the DTO body. Half of the checks in v3 are actually against database entities referenced in the DTO or Job itself. For instance, "archive" jobs should check that all datasets associated with the job have "datasetLifecycle.archivable". Or "publish" jobs with file listings should check that all files exist, requiring checking associated DataBlocks from the dataset. Should these be added to a generic ValidateAction or be a more specific check? We are leaning towards handling those as special cases in the jobs controller.

sbliven:
I changed how this works somewhat. Now the ValidateAction config looks like:

```
{
  "actionType": "validate",
  "request": {
    "JSONPath": JSONSchema,
    ...
  }
}
```

The schema gets validated by avj, so you theoretically could use the full power of a schema. However combining it with the path allows most use cases to be covered by short, easily readable one-liners. Examples:

```
{
  "actionType": "validate",
  "request": {
    "jobParams.name": { "type": "string"}, // match simple types
    "jobParams.answers[*]": { "enum": ["yes","no"]}, // literal values (here applied to an array)
    "jobResultObject": {"$ref": "https://json.schemastore.org/schema-org-thing.json"}, // JSON Schema
  }
}
```

I spelled the top-level `request` to because it gets applied to the DTO (either the create or update DTO). If it needed to get applied somewhere else (eg to associated datasets) we can add that in the future.

test/Jobs.js
```diff
@@ -40,7 +39,7 @@ let datasetPid1 = null,
   jobIdGroup4 = null,
   encodedJobIdGroup4 = null,
   jobIdGroup5 = null,
-  encodedJobIdGroup5 = null,  
+  encodedJobIdGroup5 = null,
```
sbliven, line 40: Sorry for the long diff. Everything down to like 4160 should be pure whitespaces changes (my editor removes trailing whitespace).

We should update eslint to cover js files to minimize these formatting changes.

test/Jobs.js
```diff
@@ -4158,4 +4156,192 @@ describe("1100: Jobs: Test New Job Model", () => {
         res.body.should.be.an("array").that.deep.contains({ all: [{ totalSets: 2 }] });
       });
   });
+
+  describe("3120: Validate Job Action", () => {
```
sbliven, line 4464: This file is massive. I think it would be good to either split it into several files or at least group the tests with additional `describe` blocks.

Grouping them is nice for testing since you can just add `describe.only(` to run tests from a single section.

test/TestData.js
```diff
@@ -694,62 +694,6 @@ const TestData = {
     ],
   },
 
-  ArchiveJob: {
```
sbliven, line 735: These weren't used by any tests. OK to delete them?

src/jobs/config/jobConfig.example.json
```diff
@@ -58,7 +58,18 @@
...
-        "auth": "#all"
+        "auth": "#all",
+        "actions": [
+          {
+            "actionType": "validate",
+            "request": {
+              "jobParams.datasetIds[*]": {
+                "type": "object",
+                "required": ["pid","files"]
+              }
```
sbliven, line 69: Not the best example, as it's a bit complicated.

src/jobs/config/jobConfig.updateExample.json
```diff
@@ -18,7 +18,7 @@
         {
           "actionType": "validate",
           "jobParamsSchema": {
-            "$schema": "https://json-schema.org/draft/2020-12/schema",
+            "$schema": "http://json-schema.org/draft-07/schema#",
```
sbliven: One downside of Avj is that it needs to know the schema version during import. The default supports draft v6 and v7.

src/jobs/config/jobConfig.example.json
```diff
@@ -58,7 +58,18 @@
     {
       "jobType": "public",
       "create": {
-        "auth": "#all"
+        "auth": "#all",
+        "actions": [
+          {
+            "actionType": "validate",
+            "request": {
+              "jobParams.datasetIds[*]": {
```
despadam: Should now be `datasetList` instead of `datasetIds`

despadam (changes requested):
Would it make more sense to try and also cover https://github.com/SciCatProject/scicat-backend-next/issues/1120 in this PR?

despadam:
Agreed to merge https://github.com/SciCatProject/scicat-backend-next/pull/1440, then remove `checkDatasetState` from the controller and incorporate its logic into the `validate` action, which will allow for configuring `jobType - datasetLifecycle` combinations and deleting the hardcoded `job-types.enum`. These changes will come in a new PR which will also incorporate this: https://github.com/SciCatProject/scicat-backend-next/issues/1120

despadam (approved):

src/jobs/config/jobConfig.example.json
sbliven: changed

PR #1422, opened 2024-09-12
Add fullfacet and fullquery for Jobs

despadam:
## Description

Add `fullfacet` and `fullquery` endpoints, which were omitted before, for the new Jobs implementation.

## Motivation

While trying to see whether the `release-jobs` backend is compatible with the frontend, we realized that these two endpoints were still needed.

## Changes:

- Adding the new endpoints in `jobs.controller`. To implement them, we use the same authorization model as the one for `GET/jobs`, combined with their previous implementations, as seen on `master` branch [here](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/jobs.controller.ts#L313).
- As `fullfacet` was still producing errors, due to its complicated nature, together with @nitrosx we decided it is not a necessary fix for the MVP implementation, so most of its code is commented out and it is set to always return an empty array. It will be updated in an upcoming PR.
- In `utils.ts`, the examples that were added are being used to describe the expected parameters for the `fullquery` endpoint.

## Fixes:

- Fixed the timestamps that are automatically added to the `Job schema`. Previously `createdAt` would be renamed to `DateTime` and `updatedAt` would be omitted. Now both fields are added as expected.
- In `jobs.service` fixed the order of parameters passed when calling `addCreatedByFields` during `create`. Previously `statusCode` and `statusMessage` were passed in opposite places and one would take the place the other. To complete this fix, all relevant tests were updated.

## Tests included

- Added 12 tests for `fullquery`.
- Added 2 tests for `fullfacet`. More will be added when the implementation is finished.

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -873,9 +873,6 @@ export const filterDescription =
 export const fullQueryExampleLimits =
   '{"limit": 1, "skip": 1, "order": "creationTime:desc"}';
 
-export const datasetsFullQueryExampleFields =
-  '{"mode":{},"ownerGroup":["group1"],"scientific":[{"lhs":"sample","relation":"EQUAL_TO_STRING","rhs":"my sample"},{"lhs":"temperature","relation":"GREATER_THAN","rhs":10,"unit":"celsius"}]}';
-
 export const fullQueryDescriptionLimits =
   '<pre>\n \
 {\n \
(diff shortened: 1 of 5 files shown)


sbliven (approved):
Can you make an issue for finishing the fullFacet implementation?

FullQuery looks good.

PR #1432, opened 2024-09-25
fix: fix dataset output dto

nitrosx:
## Description
This PR fix the issues with the dataset output dto found while working on the frontend. It also extends the text to test for ProposalId, SampleId and InstrumentId fields

## Motivation
While working on the search UI release, we discovered that the field's names for proposal, sample and instrument id where wrongly named.

## Fixes:
Please provide a list of the fixes implemented by this PR
- test data
- dataset tests
- dataset dtos
- dataset controller

## Tests included
- [ x] Included for each change/fix?
- [ x] Passing? (Merge will not be approved unless this is checked) 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -409,16 +409,16 @@ export class DatasetsController {
       version: "v3",
     };
 
+    if ("proposalId" in inputObsoleteDataset) {
+      propertiesModifier.proposalIds = [
+        (inputObsoleteDataset as CreateRawDatasetObsoleteDto).proposalId,
+      ];
+    }
     if (
       inputObsoleteDataset instanceof CreateRawDatasetObsoleteDto ||
       inputObsoleteDataset instanceof UpdateRawDatasetObsoleteDto ||
       inputObsoleteDataset instanceof PartialUpdateRawDatasetObsoleteDto
     ) {
-      if ("proposalId" in inputObsoleteDataset) {
-        propertiesModifier.proposalIds = [
-          (inputObsoleteDataset as CreateRawDatasetObsoleteDto).proposalId,
-        ];
-      }
       if ("sampleId" in inputObsoleteDataset) {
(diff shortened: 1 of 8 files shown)


src/datasets/datasets.controller.ts
```diff
@@ -409,16 +409,16 @@ export class DatasetsController {
       version: "v3",
     };
 
+    if ("proposalId" in inputObsoleteDataset) {
+      propertiesModifier.proposalIds = [
+        (inputObsoleteDataset as CreateRawDatasetObsoleteDto).proposalId,
+      ];
+    }
```
Junjiequan, line 416: Is it intentional that it's been scoped out of the if condition below? instead of putting together with sampleId and instrumentId conversions?

src/datasets/dto/output-dataset-obsolete.dto.ts
```diff
@@ -92,14 +92,14 @@ export class OutputDatasetObsoleteDto extends UpdateDatasetObsoleteDto {
...
-  @IsString()
-  readonly proposalId?: string;
+  // @ApiProperty({
+  //   type: String,
+  //   required: false,
+  //   description: "The ID of the proposal to which the dataset belongs.",
+  // })
+  // @IsOptional()
+  // @IsString()
+  // readonly proposalId?: string;
```
Junjiequan: Why is proposalId commented out from the dto? Does it mean we no longer provide proposalId for retrieved dataset?

src/datasets/datasets.controller.ts
nitrosx: Yes, as I added proposal id also to derived datasets

src/datasets/dto/output-dataset-obsolete.dto.ts
nitrosx: I transfer the field in the output dataset obsolete and I forgot to delete it.

Junjiequan (approved):

Issue #1411, opened 2024-09-04
Allow empty jobParams


sbliven:
jobParams is currently required and also required to be non-empty. This should be optional for job types that don't require a `datasetIds` list.

Example: POST to /api/v3/jobs
```
{
  "type": "test",
  "jobParams": {},
  "ownerUser": "user1",
  "ownerGroup": "group1",
  "contactEmail": "a@b.com"
}
```

Gives error 
```
[Nest] 418  - 09/04/2024, 6:52:46 PM   ERROR {"status":400,"message":"Job parameters need to be defined."}
[Nest] 418  - 09/04/2024, 6:52:46 PM   ERROR HttpException: Job parameters need to be defined.
[Nest] 418  - 09/04/2024, 6:52:46 PM   ERROR Object:
{
  "requestUrl": "/api/v3/jobs",
  "requestUser": {
    "_id": "66cf29d54611a38cf2c100ba",
    "username": "admin",
    "email": "admin@scicat.project",
    "currentGroups": [
      "admin",
      "globalaccess",
      "ess"
    ],
    "authStrategy": "local"
  },
  "statusCode": 400
}
```

sofyalaski:
I first wanted to fix it but then wasn't sure. This only throws an error if there are truly no jobParams. If you would put something like ProposalIds with an empty list, it creates a job ( doesn't do any checks though). I was just thinking that if there are no parameters - there is no job content.

sbliven:
Not necessarily. For instance, if you had a `test` job that just had a LogAction configured then it wouldn't need any jobParams but could still be useful for checking network connectivity.

sofyalaski:
So you think we should accept both no jobParams and empty ones? Or as Despina proposed, no jobParams is accepted but if they are provided they cannot be empty?

sofyalaski:
addressed in PR #1433. 
If you have no jobParams you must pass an empty object. JobParams is a required property in DTO. 
Empty object as jobParams is legit now.

PR #1433, opened 2024-09-26
fix issue 1411: allow jobParams 

sofyalaski:
## Description
addresses issue #1411. Now user can pass empty object as jobParams.
You still must pass the jobParams however, as it is a required property in the DTO.

## Motivation

testing, etc

## Fixes:
Doesn't throw an error on empty JobParams anymore

## Changes:

test counts are adopted 

## Tests included

## Documentation

### DIFF
--- a/src/jobs/jobs.controller.ts
+++ b/src/jobs/jobs.controller.ts
@@ -621,19 +621,6 @@ export class JobsController {
     @Body() createJobDto: CreateJobDto,
   ): Promise<JobClass | null> {
     Logger.log("Creating job!");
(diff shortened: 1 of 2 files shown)


despadam (approved):
Looks good, we just need to make sure to document that it is necessary to still pass an empty object when there are no `jobParams`. 
Then, in the case we do have `jobParams`, their validation should be handled by the Validation Action, so indeed no code for this is needed in the controller.

sofyalaski:
can we merge then?

despadam:
> can we merge then?

Yes, go ahead

PR #1436, opened 2024-10-09
Jobs: implement fullfacet

despadam:
## Description

Fully implement the `fullfacet` endpoint for Jobs.

## Motivation

As seen here: https://github.com/SciCatProject/scicat-backend-next/pull/1422, the `fullfacet` endpoint was previously set to always return an empty array.

## Changes

To be able to properly check the authorization, we first execute a `fullquery` with the same filters that the `fullfacet` is going to be executed with. This allows us to access the job configuration, based on the job type, which is needed to perform `jobsInstanceAccess` authorization and ability checks. This was the only feasible workaround to be able to retrieve this information. After performing the checks we keep a list of ids of the jobs that are accessible by the user, and add it to the facet filters. This way when the `fullfacet` is executed, it will only run on these specific jobs accessible. 

## Tests

Added more `fullfacet` tests in the `Jobs.js` test file.

### DIFF
--- a/src/jobs/interfaces/job-filters.interface.ts
+++ b/src/jobs/interfaces/job-filters.interface.ts
@@ -1,0 +1,28 @@
+interface IDateRange {
+  begin: string;
+  end: string;
+}
+
+interface IJobFieldObject {
+  $regex: string;
+  $options: string;
+}
+
+interface IJobIdsFieldObject {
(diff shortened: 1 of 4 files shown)


sofyalaski (approved):
That looks and works very nicely!

PR #1440, opened 2024-10-15
Complete Jobs controller with datasetList checks

despadam:
## Description

This PR reintroduces additional checks for Job creation that were previously commented out, namely:
- checkDatasetPids
- checkDatasetState
- checkDatasetFiles

Previously, only `checkDatasetPids` was covered by `instanceAuthorizationJobCreate`. We have now created the function `validateDatasetList` which will first validate the field `datasetList` that is passed inside `jobParams` and then execute the 3 new functions that are mentioned above.
This means that what was previously the `datasetIds` field in `jobParams` has been updated, so that it is compatible with the previous version, and now looks like:
```
datasetList: [
    {
        pid: 'someDatasetPid',
        files: [ 'file1', 'file2' ]
    }
]
```
so that it is compatible with the previous version.

## Motivation

We revert back to using `datasetList` so that we are also able to accommodate use cases where someone would only want to pass specific files of a dataset to a job.

## Fixes

- Introduced `jobs/types/job-types.enum.ts` to avoid hardcoded strings
- `common/handlebars-helpers.ts` use to map `datasetIds` to `datasetList` to provide compatibility with v3 --> no longer needed
- `jobs/config/job-example.json` and `jobs/config/jobConfig.updateExample.json` are never used and are no longer needed as examples --> have been deleted

## Tests included

- Added tests for the new functions in `test/Jobs.js`
- Updated `config/jobconfig.json` so that it also includes jobTypes `archive`, `retrieve` and `public` which are important in order to test the `checkDatasetState` function.
- Fixed `jobs/config/jobconfig.spec.ts` to also keep count of the new added jobTypes


sofyalaski (approved):
Looks good to me!

src/jobs/types/job-types.enum.ts
```diff
@@ -0,0 +1,17 @@
+export enum JobType {
```
sbliven, line 1: This is the only place where job types are hard-coded. I dislike adding it back; is there can add the checks to the job config rather than having special cases for some job names?

sbliven:
As discussed, this will be merged as well as #1421. Then we will remove `checkDatasetState` and `JobType` and migrate them to ValidateAction.

Issue #1439, opened 2024-10-14
Rename `statusUpdate` in jobconfig to just `update`


sbliven:
### Current Behaviour
Updating jobs is now done through a simple `PATCH /jobs/:id` call, rather than a distinct `/jobs/statusUpdate/:id` as in the past. However the jobConfig still includes `statusUpdate`

### Expected Behaviour
Rename `statusUpdate` to just `update` consistently.
- StatusUpdateJobDto -> UpdateJobDto
- JobConfig.statusUpdate -> JobConfig.update
etc.

despadam:
(draft) PR: https://github.com/SciCatProject/scicat-backend-next/pull/1460

PR #1460, opened 2024-10-24
Jobs: rename `statusUpdate` to `update`

despadam:
## Motivation
https://github.com/SciCatProject/scicat-backend-next/issues/1439

## Changes:
Rename `statusUpdate` to `update` as well as all relevant variable names, properties, classes


sbliven (approved):
This is a good change, and makes the API more predictable.

We do need to document the endpoint change in the migration guide. In v3 it was `POST /api/v3/Jobs/update` while now it's `PATCH /api/v3/jobs/:id`. I don't think that changed from this PR though.

sbliven (approved):
Looks good. Note that everyone needs to update their jobConfig.yaml files.

PR #1473, opened 2024-10-30
feat: Validate dataset properties in jobs

sbliven:
## Description

Allows jobs to validate properties of the dataset.

## Motivation

The `validate` action currently allows operators to enforce attributes of the DTO.

However, many jobs depend on `jobParams.datasetLists` to link a job to certain datasets. This allows validation of properties on datasets associated with the dataset.

The motivating example would be checking datasetLifecycle properties for certain jobs. These are currently hard-coded for special job types. This would allow them to be applied to custom job types as well.

```
configVersion: v1.0
jobs:
  - jobType: custom_archive
    create:
      auth: "#datasetAccess"
      actions:
        - actionType: validate
          datasets:
            datasetlifecycle.archivable:
              const: true
```

## Changes:

* Add `datasets` property to validate actions (only in `create` operations) to validate linked dataset properties

## Tests included

(WIP pending tests and documentation)

- [x] Passing? 

## Documentation

Docs are included in https://github.com/SciCatProject/documentation/pull/56


sbliven:
Documentation is now added to https://github.com/SciCatProject/documentation/pull/56

sbliven:
The unit tests pass because I'm manually mocking the datasetService, so the core functionality works.

The mocha tests fail because the datasetService doesn't get injected properly by nest. I need to figure out how to properly initialize the `ValidateCreateAction` before merging this.

sbliven:
I rebased this onto the current release-jobs.

Tests are currently failing due to a circular dependency in the module DI. I'm looking into how to fix it.

sbliven:
Tests are fixed, so this is ready for review. Some tests changed, so maybe double check that we intended the new behavior (eg I requiring more properties in the example jobs and datasets)

jobConfig.example.yaml
```diff
@@ -2,9 +2,12 @@ configVersion: v1.0 2024-03-01 6f3f38
...
   - jobType: archive
     create:
-      auth: "#all"
+      auth: "#datasetOwner"
       actions:
-        - actionType: log
+        - actionType: validate
+          datasets:
+            "datasetlifecycle.archivable":
+              const: true
```
despadam, line 10: I have to admit that I am a bit confused with this approach. I understand that the idea is to avoid hardcoded job types and allow users to easily configure their own jobs, however isn't this now also hardcoded in the config file? And also now more prone to errors since it's provided by the user. I realize that for new facility-specific job types that we don't cover in the core, perhaps this is the only way. But thinking about it the other way round, why would someone define an `archive` job for datasets with `archivable=false`? Shouldn't this already be the default? Otherwise we always force users to add `archivable=true` for `archive` jobs even though is it self-explanatory. 
So my suggestion would be that for `archive/retrieve` jobs we do the dataset lifecycle validation by default, and only ask users to provide this when they want to define a new job type. But that leads us back to hardcoded job types I think? I am happy to discuss this further, as I am clearly confused 😅  Perhaps I am misunderstanding the approach?

sbliven: > why would someone define an archive job for datasets with archivable=false?

They shouldn't. This was previously enforced by scicat for the `type:archive` job type; now it can also be configured with custom job names.

> Shouldn't this already be the default?

I could see an argument for automatically adding this validate action to any job called `archive` (and the equivalent for `retrieve` and `public` as a special case, for backwards compatibility. Or at least warning if it isn't configured. I think that can be a later discussion, as I can think of several ways to implement it, most of which boil down to syntactic sugar for the syntax here.

sbliven: #1692

src/config/job-config/actions/validateaction/validateaction.ts
```diff
@@ -100,3 +123,149 @@ export class ValidateJobAction<T extends JobDto> implements JobAction<T> {
...
+    const datasetIds = datasetList.map((x) => x.pid);
+
+    // Load linked datasets
+    const filter = {
+      where: {
+        pid: {
+          $in: datasetIds,
+        },
+      },
+    };
```
despadam, line 235: I fail to understand how the `datasetlifecycle` state is being checked here. Could you provide an explanation please?

despadam (commented):
Perhaps then we could proceed with this approach for now and discuss #1692 also with Max before making a final decision. I just have one question below.

despadam (approved):

despadam:
There is one conflict in `jobs.controller`. This block of code was removed in `release-jobs` branch, because I am assuming that it is now covered by `RabbitMQAction`:
```
  publishJob() {
    if (parseBoolean(this.configService.get<string>("rabbitMq.enabled"))) {
      // TODO: This should publish the job to the message broker.
      // job.publishJob(ctx.instance, "jobqueue");
      Logger.log("Saved Job %s#%s and published to message broker");
    }
  }
  ```
Is it still needed?

sbliven:
Yeah, that's not needed.

We also aren't using the eventEmitter anymore. I wonder if this is going to surprise anyone. Anyways, nothing to do with this PR.

despadam:
Feel free to merge then 👏

PR #1485, opened 2024-11-05
feat: added run number field to dataset and relative tests

nitrosx:
## Description
This PR adds the field runNumber to the high level fields of the datasets. The field is optional, but it has come up many times in our assessments with users that it makes sense to promote it to high level, making it easier to create the proper search functionalities.

## Motivation
In our assessments with users, they have mention run number as a piece of information with high relevance. 
So far it was included in the scientific metadata, but it had suffer from the variability introduced by the freedom provided by the field. 
Promoting  run number to high level field, it will allow us to reduce such variability and provide a more consistent search functionality.

## Changes:
* dataset schema
* dataset dtos
* raw dataset tests

## Tests included
- [x] Passing? 

## Documentation

### DIFF
--- a/src/datasets/dto/update-dataset-obsolete.dto.ts
+++ b/src/datasets/dto/update-dataset-obsolete.dto.ts
@@ -297,6 +297,16 @@ export class UpdateDatasetObsoleteDto extends OwnableDto {
   @IsOptional()
   @IsString()
   readonly proposalId?: string;
+
+  @ApiProperty({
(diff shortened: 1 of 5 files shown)


Junjiequan (approved):
Looks good to me in general.
Assigning a string value to the field named runNumber is a bit confusing, but I guess it's necessary.

bpedersen2:
:+1:

PR #1486, opened 2024-11-06
ci(py-sdk): fix python package names

cchndl:
Python package names are not allowed to contain '-', which is a syntax error. Replace with '_'.

## Description

The package names in Python shouldn't contain hyphens, as this is a syntax error.

Replacing with underscore fixes this, but the drawback is that now project and package names differ. This is fine but not recommended. But whether only the pacakge name should be changed or you also modify the project names on Pypi is your call. 

## Motivation
Both Python sdk are not usable as the import fails with a syntax error. 

## Changes:

* rename Python sdk package names

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/.github/openapi/python-config.json
+++ b/.github/openapi/python-config.json
@@ -1,5 +1,5 @@
 {
   "generatorName": "python",
-  "packageName": "scicat-sdk-py",
+  "packageName": "scicat_sdk_py",
   "projectName": "scicat-sdk-py"
 }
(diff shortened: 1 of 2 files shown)


Junjiequan:
Thanks for the fix.
I've just included a small change so that package name is consistent with project name to avoid confusion.

Junjiequan (approved):

PR #1499, opened 2024-11-12
feat: add parent proposal to the proposal document

martin-trajanovski:
## Description
This PR aims to add the new `parentProposalId` field in the proposal document.

## Motivation
There was a feature requested to add new field on a proposal called parent proposal

## Fixes

* https://jira.ess.eu/browse/SWAP-4298

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/proposals/proposals.controller.ts
+++ b/src/proposals/proposals.controller.ts
@@ -17,6 +17,7 @@ import {
   BadRequestException,
   Logger,
   InternalServerErrorException,
+  NotFoundException,
 } from "@nestjs/common";
 import { Request } from "express";
 import { ProposalsService } from "./proposals.service";
@@ -578,7 +579,10 @@ export class ProposalsController {
     const proposal = await this.proposalsService.findOne({
       proposalId,
     });
-    if (!proposal) return { canAccess: false };
+
+    if (!proposal) {
+      throw new NotFoundException(`Proposal with ${proposalId} not found`);
+    }
 
     const canAccess = await this.permissionChecker(
       Action.ProposalsRead,
--- a/src/proposals/proposals.controller.ts
+++ b/src/proposals/proposals.controller.ts
@@ -14,10 +14,8 @@ import {
   Req,
   ForbiddenException,
   ConflictException,
-  BadRequestException,
   Logger,
   InternalServerErrorException,
-  NotFoundException,
 } from "@nestjs/common";
 import { Request } from "express";
 import { ProposalsService } from "./proposals.service";
@@ -90,14 +88,17 @@ export class ProposalsController {
     return proposalInstance;
   }
 
-  private async permissionChecker(
+  private permissionChecker(
     group: Action,
-    proposal: ProposalClass | CreateProposalDto,
+    proposal: ProposalClass | CreateProposalDto | null,
     request: Request,
   ) {
-    const proposalInstance = this.generateProposalInstanceForPermissions(
-      proposal as ProposalClass,
-    );
+    if (!proposal) {
+      return false;
+    }
+
+    const proposalInstance =
+      this.generateProposalInstanceForPermissions(proposal);
 
     const user: JWTUser = request.user as JWTUser;
     const ability = this.caslAbilityFactory.proposalsInstanceAccess(user);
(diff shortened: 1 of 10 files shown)


src/proposals/proposals.controller.ts
```diff
@@ -578,7 +579,10 @@ export class ProposalsController {
     const proposal = await this.proposalsService.findOne({
       proposalId,
     });
-    if (!proposal) return { canAccess: false };
+
+    if (!proposal) {
+      throw new NotFoundException(`Proposal with ${proposalId} not found`);
+    }
```
nitrosx: Probably we should just log a message that the parent proposal is not valid, but return the proposal anyway

test/Proposal.js
```diff
@@ -216,7 +218,46 @@ describe("1500: Proposal: Simple Proposal", () => {
...
+      .then((res) => {
+        res.body.should.have.property("ownerGroup").and.be.string;
+        res.body.should.have.property("proposalId").and.be.string;
+        proposalWithParentId = res.body.proposalId;
+        res.body.should.have.property("parentProposalId").and.be.string;
+        res.body.parentProposalId.should.be.equal(proposalId);
+      });
+  });
+
+  it("0120: updates a proposal with a new parent proposal", async () => {
```
nitrosx, line 244: Number it as 0125

test/RawDataset.js
```diff
@@ -280,6 +283,26 @@ describe("1900: RawDataset: Raw Datasets", () => {
     );
   });
 
+  it("01250: adds a new proposal for pattching in the existing dataset", async () => {
```
nitrosx: fix test string to "0124: adds a new proposal for patching in the existing dataset"

src/proposals/proposals.controller.ts
Junjiequan: This endpoint is to avoid error throwing to the frontend for users who have not permission to access the specific proposal. 
It should only return canAccess true or false, and additional information should just log with `logError`  without throwing error.

Junjiequan (changes requested):

src/proposals/proposals.controller.ts
martin-trajanovski: I understand that and it should work exactly as you described. Even if you don't have access to the proposal as you can see the `permissionChecker` is right after the proposal query. For example this throws error only if you provide a `proposalId` that doesn't exist. I think in this situation it is good to throw error. Otherwise after that where we check the permission we will return `true` or `false` accordingly. For me this is the best and most reasonable way. I don't see a point returning `200` success with `canAccess: false` if the queried proposal doesn't even exist.

martin-trajanovski: @nitrosx please check the comment that I left little bit bellow where I explained to Jay why I did this. I think that is the most reasonable way to move forward.

test/Proposal.js
martin-trajanovski: I will try to fix the numbering but honestly I am not a big fan of this way for naming the tests. I understand it why it is done but now if I want to add a test in the middle somewhere I need to update all the numbers after that one which is a bit odd.

nitrosx: I agree with you, although it makes it easier for wider range of people to be more effective when working with tests... including myself!!! :)

src/proposals/proposals.controller.ts
nitrosx: At this moment, I do not know which is the best and most sensible way, although I would just return false if the proposal does not exists.
What I suggest is that both endpoints `proposal/{id}` and `proposal/{id}/authorization` behave the same way.

martin-trajanovski: Yes I understand you but then maybe it is better to just have the `permissionChecker` function to decide about this and not even have this condition in the `findByIdAccess` function. I will submit new improved solution and you can check it if you like it or not.

Junjiequan: The endpoint is designed to return only true or false to keep the frontend unaware of backend details. It’s used exclusively on the dataset details page on frontend , so we avoid throwing errors and simply show nothing if access fails for any reason, however, logging error to the grayLog would help us monitoring the issues.  
Ideally, error handling should occur within findById, not findByIdAccess. I don't know which way is better, what you suggested is good practice, I'm not sure if it's what we want

test/Proposal.js
Junjiequan: It is quite handy, especially when you need only to run specific test, like this
`npx mocha ./test/config/pretest.js ./test/DerivedDataset.js -r chai/register-should.js  —grep “0125”`

src/proposals/proposals.controller.ts
Junjiequan: That being said, if you can convince Max, I'm okay with this approach, as it is just to handle edge cases

martin-trajanovski: @Junjiequan this is being fixed by me and refactored a bit. You are welcome to double check it once more.

nitrosx (approved):

Junjiequan (approved):

PR #1503, opened 2024-11-13
feat: add type property to the proposal document with database migration

martin-trajanovski:
## Description
This PR aims to add the `type` property on a proposal. It also adds the support for configuring the proposal types using `proposalTypes.json` file located in the root directory of the project. If no configuration is provided the default value is `Default Proposal` and that's why we have added the database migration as well to update all the existing proposals with default value.

## Motivation
`type` property was needed on a proposal and for ESS we will use only the default one `Default Proposal` for now.

## Fixes

* https://jira.ess.eu/browse/SWAP-4299

## Changes:

* changes made

## Tests included

- [x] Passing? 

## Documentation


src/proposals/dto/update-proposal.dto.ts
```diff
@@ -4,12 +4,14 @@ import {
...
   IsDateString,
   IsEmail,
+  IsEnum,
   IsObject,
   IsOptional,
   IsString,
   ValidateNested,
 } from "class-validator";
 import { OwnableDto } from "../../common/dto/ownable.dto";
+import { ProposalType } from "../proposal-type.enum";
```
nitrosx: We should have the proposal list in configuration not in an hard coded enum

src/proposals/proposal-type.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum ProposalType {
+  DefaultProposal = "Default Proposal",
+  DOORProposal = "DOOR Proposal",
+  Beamtime = "Beamtime",
+}
```
nitrosx: We can enforce `default proposal` in the code, but the additional types must be in configuration.

src/proposals/schemas/proposal.schema.ts
```diff
@@ -3,6 +3,7 @@ import { ApiProperty } from "@nestjs/swagger";
 import { Document } from "mongoose";
 
 import { OwnableClass } from "src/common/schemas/ownable.schema";
+import { ProposalType } from "../proposal-type.enum";
```
nitrosx: See my comment above

nitrosx (changes requested):

proposalTypes.example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "DefaultProposal": "Default Proposal",
```
nitrosx: Could we remove `Default Proposal` from here into code?
Meaning can it be added in code automatically.

src/config/configuration.ts
```diff
@@ -57,6 +58,25 @@ const configuration = () => {
...
 
+  let proposalTypesData;
+
+  try {
+    proposalTypesData = fs.readFileSync("proposalTypes.json", "utf8");
+  } catch (error) {
+    // NOTE: If there is no proposalTypes.json file provided we just use the DEFAULT_PROPOSAL_TYPE type
+    console.warn(error);
+  }
+
```
nitrosx: Add here to `proposalTypesData` the option `DefaultProposal`.

nitrosx (changes requested):

proposalTypes.example.json
martin-trajanovski: Do you mean to append it in code if not present in the configuration file or?

src/config/configuration.ts
martin-trajanovski: Ok you mean to check here if the `DefaultProposal` is present in the config then move forward if not then add it.

proposalTypes.example.json
nitrosx: Correct. I will assume that it is not present and add it.

src/config/configuration.ts
nitrosx: as I mentioned above, assume that default proposal is not there. Add the option here

src/config/configuration.ts
```diff
@@ -57,6 +58,26 @@ const configuration = () => {
...
+    console.warn(
+      `Proposal types configuration file not provided. The "${DEFAULT_PROPOSAL_TYPE}" type will be used`,
+    );
+  }
+
+  const proposalTypes: Record<string, string> = {
+    ...JSON.parse(proposalTypesData || "{}"),
+    DefaultProposal: DEFAULT_PROPOSAL_TYPE,
+  };
+
```
Junjiequan: I think this can be included in the jsonConfigFileList

```
  const jsonConfigMap: { [key: string]: object[] | boolean } = {};
  const jsonConfigFileList: { [key: string]: string } = {
    loggers: process.env.LOGGERS_CONFIG_FILE || "loggers.json",
  };
  Object.keys(jsonConfigFileList).forEach((key) => {
    const filePath = jsonConfigFileList[key];
    if (fs.existsSync(filePath)) {
      const data = fs.readFileSync(filePath, "utf8");
      try {
        jsonConfigMap[key] = JSON.parse(data);
      } catch (error) {
        console.error(
          "Error json config file parsing " + filePath + " : " + error,
        );
        jsonConfigMap[key] = false;
      }
    }
  });
```

martin-trajanovski: Do you want me to hardcode it or add another variable for the `proposalTypes.json` file?

nitrosx (approved):

PR #1532, opened 2024-11-27
feat: allow custom dataset types defined in configuration

HayenNico:
## Description
Change dataset model and endpoints to accept custom dataset types at creation. Custom types are defined in a json file, default location is "datasetTypes.json". Preexisting base types Raw and Derived are always supported.
Custom types assume use of the new unified CreateDatasetDto/UpdateDatasetDto.

## Motivation
Additional dataset types besides raw and derived are required e.g. for the proposed dataset collections (hierarchical grouping). Control of which types are allowed should be on the site administration side and be configurable.

## Fixes:

* Add missing runNumber property to new unified UpdateDatasetDto for compatibility

## Changes:

* configuration: add new datasetTypes field and support to read in a json file for custom datasest types (either by env variable or default location "datasetTypes.json")
* add support for custom dataset types in dataset schema
* dataset controller: update endpoints and validation methods to accept custom dataset types. 

## Tests included

- [x] Passing? 

## Documentation


src/datasets/datasets.controller.ts
```diff
@@ -536,15 +556,29 @@ export class DatasetsController {
     @Body()
     createDatasetObsoleteDto:
       | CreateRawDatasetObsoleteDto
-      | CreateDerivedDatasetObsoleteDto,
+      | CreateDerivedDatasetObsoleteDto
+      | CreateDatasetDto,
   ): Promise<OutputDatasetObsoleteDto> {
     // validate dataset
+    let dtoType;
+    switch (createDatasetObsoleteDto.type) {
```
martin-trajanovski, line 592: We can maybe extract this in it's own function and reuse it in all the places needed.

src/datasets/datasets.module.ts
```diff
@@ -1,11 +1,11 @@
-import { forwardRef, Module } from "@nestjs/common";
+import { BadRequestException, forwardRef, Module } from "@nestjs/common";
 import { MongooseModule } from "@nestjs/mongoose";
 import { DatasetClass, DatasetSchema } from "./schemas/dataset.schema";
 import { DatasetsController } from "./datasets.controller";
 import { DatasetsService } from "./datasets.service";
 import { CaslAbilityFactory } from "src/casl/casl-ability.factory";
 import { AttachmentsModule } from "src/attachments/attachments.module";
-import { ConfigModule } from "@nestjs/config";
+import { ConfigModule, ConfigService } from "@nestjs/config";
```
martin-trajanovski: removing non used imports

martin-trajanovski (approved):
Looks good to me now. 🚀

src/datasets/datasets.controller.ts
HayenNico: Had to revert this change, the return Dto type (e.g. Create, Update and PartialUpdate) is different in each of the three instances this happens, so the logic is not identical

martin-trajanovski: Yes I see it now. OK then maybe we can keep it as it is for now as this will be refactored and cleaned up little bit soon where we will use mostly the standardized DTOs for all the dataset endpoints accordingly

martin-trajanovski:
@nitrosx maybe we should assign someone else from the collaborators to take a look at this and if they are fine we can move forward and merge it. What do you think?

nitrosx (approved):

HayenNico:
@nitrosx Regarding the new changes to the isValid endpoint: Should this work with the custom dataset types or not? There is still some code left in the function to facilitate validation, but the function arguments suggest it's meant explicitly for obsolete DTOs.

nitrosx:
@HayenNico my bad. I forgot about that. Please add the DatasetDto back.
Sorry about that.

PR #1534, opened 2024-11-27
fix: delete_job_groups instance authorization

sofyalaski:
## Description
When in .env delete-job-group was set to admin, flowing the authorization logic, admin should be able to delete a job. Instead, this user would get denied access. 

## Motivation
fix this

## Fixes
user can delete a job based on delete-job-group

## Changes:
* In casl, the endpoint authorization lines for JobDelete are separated into their own if/else clauses
* added admin to delete_job_group in tests workflow
* 1920 test shour return a successful status code
* test for getting are adapted to minus one deleted test. 

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -155,7 +155,7 @@ jobs:
           ACCESS_GROUPS_STATIC_VALUES: "ess"
           CREATE_JOB_GROUPS: group1,group2
           UPDATE_JOB_GROUPS: group1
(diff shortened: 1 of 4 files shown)


sbliven (approved):
Thanks for the fix!

sofyalaski:
> Thanks for the fix!

then I will create another PR for scicatlive. there was a change in env for tests again..

nitrosx (approved):

PR #1536, opened 2024-12-03
feat: migrate dataset principalInvestigator to principalInvestigators

martin-trajanovski:
## Description
This PR aims to migrate the existing field of `principalInvestigator` -> `principalInvestigators` and allow the datasets to store multiple principal investigators as we can have multiple proposals.

## Motivation
We can have multiple proposals on a dataset but not multiple principal investigators.

## Changes:
Changes are mostly in the schema, providing migration script and conversion of the obsolete to new schema and the opposite.

## Tests included

- [x] Passing? 

## Documentation


nitrosx (approved):

src/datasets/schemas/dataset.schema.ts
```diff
@@ -362,8 +362,8 @@ export class DatasetClass extends OwnableClass {
     description:
       "First name and last name of principal investigator(s). If multiple PIs are present, use a semicolon separated list. This field is required if the dataset is a Raw dataset.",
   })
-  @Prop({ type: String, required: false })
-  principalInvestigator?: string;
+  @Prop({ type: [String], required: false })
+  principalInvestigators?: string[];
```
Junjiequan, line 365: The description might also need to be changed?

Something like `Multiple PIs can be provided as separate strings in the array`?

src/datasets/dto/update-dataset.dto.ts
```diff
@@ -291,13 +291,14 @@ export class UpdateDatasetDto extends OwnableDto {
...
 
   @ApiProperty({
-    type: String,
-    required: true,
+    type: [String],
+    required: false,
     description:
       "First name and last name of principal investigator(s). If multiple PIs are present, use a semicolon separated list. This field is required if the dataset is a Raw dataset.",
   })
+  @IsOptional()
```
Junjiequan: The description might also need to be changed?

Something like `Multiple PIs can be provided as separate strings in the array`?

Junjiequan (approved):
Commentted minor things, otherwise looks fine to me

Issue #1154, opened 2024-04-08
Refactor JobConfig as a nest module


sbliven:
The `JobConfig` class is currently a vanilla TS class with some singleton code to avoid reloading it too much. Instead, it should be implemented as a NestJS module. This would be more idiomatic, make mocking easier (#1114) and eliminate code. NestJS should handle parsing, loading the singleton, and registering JobActions (via input injection).

sbliven:
Closed with #1539, although suggestions are still welcome on the architecture.

PR #1539, opened 2024-12-03
refactor: Implement JobConfig as a NestJS module

sbliven:
## Description

Refactor the JobConfig following #1154.

## Motivation

The main motivation is to make it easier to implement JobActions that depend on NestJS modules. For instance, `EmailJobAction` now can call MailService, `ValidateJobAction` will be able to make database calls (#1421 ), etc.

Registering JobActions was also awkward with the previous implementation, and tests were required to call some setup code. NestJS is great for managing singletons, so all this setup and caching can be replaced with standard module imports.

## Details

Implementing an Action now requires 4 (short) files:
- `action.ts` contains a class implementing `JobAction`. It's constructor can take any arguments, but the existing actions take an `options` argument mirroring the expected yaml config. It does not need to be `@Injectable()`, since it is constructed by the factory.
- `action.interface.ts` can contain additional types, e.g. the definition of the expected `options` and a type guard for casting to the options.
- `action.factory.ts` implements `JobActionFactory`. The factory is provided by NestJS as a singleton, so it must be `@Injectable()`. This means that dependencies can be injected into the factory. It has a `create(options)` method, which constructs the action itself by combining the options from the yaml file with any dependencies injected by NestJS.

The lists of known factories are provided to Nest with the `CREATE_JOB_ACTION_FACTORIES` and `UPDATE_JOB_ACTION_FACTORIES` symbols. The top-level AppModule imports built-in actions from `DefaultJobActionFactories`. Future plugins can then use the NestJS dependency injection system to extend the lists if needed.

Parsing jobConfig.yaml is handled by the JobConfigService. NestJS injects the lists of factories (and the file path) during construction. When parsing reaches a new action, the correct factory is used to create an instance of the JobAction with the current list of options.

Code which needs the configuration for a particular job type (eg jobs.controller.ts) injects the `JobConfigService` and can then call `jobConfigService.get(jobType)`.

## Changes:

* Refactor Job Configuration system (#1154)

## Tests included

- [x] Passing? 

## Documentation


sbliven:
Unit tests should now pass again. I still need to confirm that mocha tests pass.

There are some lint errors due to `any` being used. I will fix these.

The rabbitmq and urlactions still need to be migrated.

sbliven:
I've squashed all previous commits to make it easier to review.

src/common/mail.service.ts
```diff
@@ -1,28 +1,25 @@
...
   constructor(private readonly mailerService: MailerService) {}
 
-  async sendMail(
-    to: string,
-    cc: string,
-    subject: string,
-    mailText: string | null,
-    html: string | null = null,
-  ) {
+  async sendMail(options: ISendMailOptions): Promise<SentMessageInfo> {
```
Junjiequan, line 14: I see that we have `nodemailer` package installed separately, and it is used only as a type (which is Any type) in this file. However, `@nestjs-modules/mailer` already has nodemailer as part of the dependency package. 
I'd suggest to remove `nodemailer` from the package.json if we have no intention to use this package separately and rely on dependency nodemailer from @nestjs-modules/mailer exclusively

src/auth/auth.module.ts
```diff
@@ -13,7 +13,7 @@ import { OidcConfig } from "src/config/configuration";
 import { BuildOpenIdClient, OidcStrategy } from "./strategies/oidc.strategy";
 import { accessGroupServiceFactory } from "./access-group-provider/access-group-service-factory";
 import { AccessGroupService } from "./access-group-provider/access-group.service";
-import { CaslAbilityFactory } from "src/casl/casl-ability.factory";
+import { CaslModule } from "src/casl/casl.module";
```
nitrosx, line 16: I am curious about renaming the module?

src/datasets/datasets.module.ts
nitrosx, line 1: What are the differences between using casl module as a providers and as an import? What are the advantages of the new implementation?

nitrosx (commented):
Just a couple of questions.
Other than that, I think this PR is good to go.

src/config/job-config/jobconfig.interface.ts
```diff
@@ -0,0 +1,85 @@
...
+  configVersion?: string;
+  jobs: JobConfigOptions[];
+}
+
+/**
+ * Encapsulates all responses to a particular job type (eg "archive")
+ */
+export interface JobConfig {
+  jobType: string;
+  configVersion: string;
```
despadam, line 21: Why do we also need `configVersion` per job? Isn't this information only on the file level?

src/config/job-config/jobconfig.interface.ts
```diff
@@ -0,0 +1,85 @@
...
+  create: JobOperation<CreateJobDto>;
+  statusUpdate: JobOperation<StatusUpdateJobDto>;
+}
+export interface JobConfigOptions {
```
despadam, line 25: In general I am struggling a bit to understand this file. What is the difference between `JobConfig` and `JobConfigOptions` and why do we need both? The same with `JobOperation` etc ...

src/config/job-config/jobconfig.service.ts
```diff
@@ -0,0 +1,132 @@
...
+   */
+  private loadJobConfig(filePath: string): Record<string, JobConfig> {
+    // If no file is given don't configure any jobs
+    if (!filePath) {
+      return {};
+    }
+    const yaml = fs.readFileSync(filePath, "utf8");
+    const jobConfigListOptions = load(yaml, { filename: filePath });
+
+    // TODO: Do we like the schema-based validation, or should it be pure typescript?
```
despadam, line 65: For me it's ok to leave it as is. The schema only checks for minimal information as everything else can be configured. In case we decide to expand the schema, we could then change the validation approach. But for now I think it is not a concern

src/config/job-config/actions/defaultjobactions.module.ts
```diff
@@ -0,0 +1,90 @@
...
+  ],
+  providers: [
+    {
+      provide: CREATE_JOB_ACTION_FACTORIES,
+      useFactory: (
+        logJobActionFactory,
+        emailJobActionFactory,
+        validateJobActionFactory,
+        urlJobActionFactory,
+        rabbitMQJobActionFactory,
```
despadam: Maybe I am misunderstanding this or the name `defaultjobactions` is confusing me, but by doing this don't we by default enable all actions that are written here? Instead we should only enable those that are declared in the jobConfig file. Perhaps if this was split to `default` and `additional` actions it would be clearer? Where in `additional` someone could declare whether the use rabbitMQ or Kafka for example, instead of by default having rabbitMQ

src/config/job-config/actions/defaultjobactions.module.ts
```diff
@@ -0,0 +1,90 @@
...
+        EmailJobActionFactory,
+        ValidateJobActionFactory,
+        URLJobActionFactory,
+        RabbitMQJobActionFactory,
+      ],
+    },
+  ],
+  exports: [CREATE_JOB_ACTION_FACTORIES, UPDATE_JOB_ACTION_FACTORIES],
+})
+export class DefaultJobActionFactories {}
```
despadam: Is this a factory or a module?

src/auth/auth.module.ts
sbliven: It's not renamed. Importing the module gives you the app-wide singleton instance. The previous pattern was instantiating a new CaslAbilityFactory outside of the dependency injection framework.

src/datasets/datasets.module.ts
sbliven: See above.

My understanding is that it is an error to directly instantiate a provider (eg anything marked `@Injectable` or `@Controller`). Instead you need to import the module and get the singleton from that module.

src/common/mail.service.ts
sbliven: Good suggestion. I think `reflect-metadata` is also used only indirectly, but I will save removing that for a different PR.

despadam (commented):
Overall looks good, thanks for all the great work @sbliven! I left a few comments, plus have a general question: 
Going through these changes I realized that the controllers import factories, which in my understanding perhaps is not a great approach, as controllers should import services. More specifically, I am referring to the `CaslAbilityFactory` that keeps getting imported and reused multiple times. 
I think the topic of refactoring Casl has come up a few times already and I believe it is something that we will have to look into. But in anyway it should be in a separate PR, so we can discuss this at a later time

sbliven:
Thanks for the review @despadam.

> controllers import factories, which in my understanding perhaps is not a great approach, as controllers should import services. More specifically, I am referring to the CaslAbilityFactory that keeps getting imported and reused multiple times.

I think this is a problem with inconsistent naming. NestJS doesn't give clear definitions of all of these terms. Here's my own (possibly hallucinated by chatGPT) definitions:
- *provider*: A provider is a class or a value that can be injected into other classes (via dependency injection) to encapsulate reusable logic or state. Classes decorated with `@Injectable()` are providers, as are objects with a `{provides:token}`
- *service*:  A specific type of provider primarily used to define reusable business logic or data access operations. Every service is a provider, but not all providers are services.
- *module*: a namespace for dependency injection, responsible for instantiating providers, injecting them into other providers as needed, and exporting the singletons to other modules.
- *factory*: a function that generates a value or an object to be used as a provider. It can dynamically determine the provider’s instance at runtime based on logic or configuration.

Many of our classes don't follow these terms
- `CaslAbilityFactory` is actually a provider, not a factory. It is an `@Injectable()` class intended as a singleton
- `accessGroupServiceFactory` is a provider that uses an (anonymous) factory for its instantiation
- `JobActionFactory` subclasses are providers, since they use `@Injectable()`! They do create `JobAction` instances, but this is done outside of nestjs.

I played around a bit with trying to make JobActionFactories actual nest factories (eg functions that return providers). All that would be needed to make JobAction implementations into providers would be to add `@Injectable()`, so this seems within reach. But after trying to understand how dependency injection would work with a factory my brain threw a 'RecursionError: Maximum Recursion Depth Exceeded' and so I just committed the version here. I would be open to either suggestions as to changing the name or refactoring to make it an actual factory.

src/config/job-config/jobconfig.interface.ts
sbliven: This is how it was before. We don't store the top-level object (`JobConfigListOptions` here) so the root `configVersion` has to get copied into each JobConfig section. I'm not sure what happened before if you specified it down a level too. Maybe we should remove it from `JobConfigOptions` entirely, rather than using the top-level field as a default that can still be overridden one level down?

There's potentially a use case where you only update one job type and so only bump one version, but that sounds complicated to track.

src/config/job-config/jobconfig.interface.ts
sbliven: `JobConfigOptions` represents the POD object parsed from yaml. This is then converted to the real `JobConfig` instance. They have similar structure except that the `JobConfig` has `JobAction` instances at its leaves, while `JobConfigOptions` has the configuration strings as leaves.

sbliven: It's possible this could be done directly in the yaml parser as a custom deserializer, but this follows the patterns used in Nest's documentation for custom providers.

src/config/job-config/jobconfig.service.ts
sbliven: I'm not sure validation is even necessary, since all the constructors throw errors if the yaml doesn't follow the expected structure. I expect ajv gives clearer error messages though.

src/config/job-config/actions/defaultjobactions.module.ts
sbliven: When the yaml file gets parsed you need to have a mapping between `actionType` strings and the `JobActionFactory` instance that will parse that part of the config and create a `JobAction`. Would it be clearer if the file were called `defaultjobactionfactories.module.ts` like the module? Somehow the name should reflect that this is the stuff that gets created before the yaml file is parsed, not the stuff configured by the yaml.

If you wanted to add new actions (say using a plugin) you would provide new lists for `CREATE_JOB_ACTION_FACTORIES` and `UPDATE_JOB_ACTION_FACTORIES`. I guess this could be split into the built-in ones and `additional` factories loaded from plugins, but it's not clear to me what you would gain from that.

src/config/job-config/actions/defaultjobactions.module.ts
sbliven: It's a module in the NestJS sense, but its role is to provide the factories-that-create-JobAction-instances. Please suggest better names, this is starting not to even make sense to me.

Junjiequan:
That's a very clear explaniation of the nestJs module system. Thanks @sbliven 

To add a little extra information when creating a module:
exports: Makes services/providers available for injection in other modules when this module is imported.
controllers: Declares which controllers should handle HTTP requests for the routes managed by this module.

src/config/job-config/jobconfig.interface.ts
despadam: You are right that this is how it was before, I got confused. I think the confusion came from the fact that we also store `configVersion` in `JobConfigListOptions` but never actually access it from there. From what I see, we extract it directly either from `JobConfig` or `JobConfigOptions`, depending on the case.

We have agreed that we will not cover the use case of updating individually the versions, exactly because it becomes too tricky. There is only a check in the controller that will print an error message when a job's version is not the same as the current running config version - I believe this is enough

sbliven: I've made `JobConfigListOptions.configVersion` required and made sure it always gets passed through to the JobConfig level.

The ajv schema enforces that configVersion can only appear as a top-level field.

src/config/job-config/actions/defaultjobactions.module.ts
sbliven: I've renamed this to `JobActionCreator`.

sbliven:
In ec245bb I switch the code to use `JobActionCreator` to avoid confusion with the NestJS term `factory`. I suggest in #1560 renaming some of the other "factories" too, but for now I will only rename the ones introduced in this PR.

PR #1541, opened 2024-12-05
feat: add new dataset v4 controller

martin-trajanovski:
## Description
This PR aims to introduce the new dataset controller v4 (version 4). IT introduces breaking changes to the dataset API by cleaning up all the redundant endpoints and using the unified Dataset types. Dataset endpoints like: find, findOne, count, findById now use mongo aggregation for easier relational field resolution.

## Motivation
Dataset controller was a beast with a lot of redundant endpoints and the types were a complete mess.

## Fixes

* https://jira.ess.eu/browse/SWAP-4318

## Changes:

* Changes are mostly in introducing the new controller and adjusting dataset service methods for aggregation.
* This PR improves the v4 dataset endpoints swagger documentation and makes it easier to use

## Tests included

- [ ] Passing? 

## Documentation


martin-trajanovski:
@martin-trajanovski Do not forget to do "Squash and merge" instead of just normal merge

Junjiequan:
It looks more like a breaking change to me

martin-trajanovski:
> It looks more like a breaking change to me

You are right. But I wasn't sure if we should name it like that as we keep both for now and not removing the old v3 apis. Maybe once we remove those it should be named as a breaking change otherwise we are backwards compatible for now.

.github/workflows/release-and-publish-sdk.yml
```diff
@@ -112,6 +112,7 @@ jobs:
         env:
           MONGODB_URI: "mongodb://localhost:27017/scicat"
           JWT_SECRET: thisIsTheJwtSecret
+          SDK_PACKAGE_SWAGGER_HELPERS_DISABLED: true
```
nitrosx, line 117: What is the purpose for this variable?

src/config/configuration.ts
```diff
@@ -81,24 +80,24 @@ const configuration = () => {
...
+        .map((v) => v.trim()),
+      createDatasetPrivileged: createDatasetPrivilegedGroups
+        .split(",")
+        .map((v) => v.trim()),
+      proposal: proposalGroups.split(",").map((v) => v.trim()),
+      sample: sampleGroups.split(",").map((v) => v.trim()),
+      samplePrivileged: samplePrivilegedGroups.split(",").map((v) => v.trim()),
+      createJob: createJobGroups,
+      updateJob: updateJobGroups,
+    },
```
nitrosx, line 99: I wonder the reasons for these changes. Can you clarify, please?

src/datasets/datasets.controller.ts
```diff
@@ -125,7 +128,22 @@ export class DatasetsController {
...
+    );
+    this.datasetCreationValidationRegex = this.configService.get<string>(
+      "datasetCreationValidationRegex",
+    );
+    this.datasetTypes = this.configService.get<string>("datasetTypes");
+  }
+  private accessGroups;
+  private datasetCreationValidationEnabled;
+  private datasetCreationValidationRegex;
+  private datasetTypes;
```
nitrosx, line 146: Why store them locally?

src/datasets/datasets.module.ts
```diff
@@ -5,19 +5,18 @@ import { DatasetsController } from "./datasets.controller";
...
 import { LogbooksModule } from "src/logbooks/logbooks.module";
 import { PoliciesService } from "src/policies/policies.service";
 import { PoliciesModule } from "src/policies/policies.module";
 import { ElasticSearchModule } from "src/elastic-search/elastic-search.module";
+import { DatasetsV4Controller } from "./datasets.v4.controller";
 
 @Module({
   imports: [
     AttachmentsModule,
-    ConfigModule,
```
nitrosx, line 20: Why we removed the config module from the imports?

src/datasets/datasets.service.ts
```diff
@@ -62,11 +79,9 @@ export class DatasetsService {
...
   }
 
-  async findAll(
-    filter: IFilters<DatasetDocument, IDatasetFields>,
-  ): Promise<DatasetClass[]> {
-    const whereFilter: FilterQuery<DatasetDocument> = filter.where ?? {};
-    const fieldsProjection: FilterQuery<DatasetDocument> = filter.fields ?? {};
+  async findAll(filter: FilterQuery<DatasetDocument>): Promise<DatasetClass[]> {
+    const whereFilter = filter.where ?? {};
+    const fieldsProjection = filter.fields ?? {};
```
nitrosx: why did we removed the typing?

src/datasets/pipes/pid-validation.pipe.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.datasetCreationValidationEnabled = this.configService.get<boolean>(
+      "datasetCreationValidationEnabled",
+    );
+    this.datasetCreationValidationRegex = this.configService.get<string>(
+      "datasetCreationValidationRegex",
+    );
+  }
+
+  private datasetCreationValidationEnabled;
+  private datasetCreationValidationRegex;
```
nitrosx, line 20: Why storing in local variables?

src/datasets/types/dataset-filter-content.ts
nitrosx, line 1: What is the purpose and functionality of this file?

src/datasets/types/dataset-lookup.ts
```diff
@@ -0,0 +1,73 @@
...
+    },
+  },
+  all: {
+    $lookup: {
+      from: "",
+      localField: "",
+      foreignField: "",
+      as: "",
+    },
+  },
```
nitrosx: Why the empty structure?

src/instruments/instruments.controller.ts
```diff
@@ -142,12 +139,10 @@ export class InstrumentsController {
...
       );
+
       return instrument;
-    } catch (e) {
-      throw new HttpException(
-        "Instrument with the same unique name already exists",
-        HttpStatus.BAD_REQUEST,
-      );
+    } catch (error) {
+      throw new InternalServerErrorException(error);
```
nitrosx: Does this exception expose the internal error to the user?
If so, I would like to change it so we log the internal error and we provide the same error as before.

.github/workflows/upload-sdk-artifact.yml
```diff
@@ -30,6 +34,7 @@ jobs:
         env:
           MONGODB_URI: "mongodb://localhost:27017/scicat"
           JWT_SECRET: thisIsTheJwtSecret
+          SDK_PACKAGE_SWAGGER_HELPERS_DISABLED: true
```
nitrosx, line 38: Same as above

src/app.module.ts
```diff
@@ -35,13 +35,15 @@ import { LoggerModule } from "./loggers/logger.module";
...
 @Module({
   imports: [
-    AttachmentsModule,
-    AuthModule,
-    CaslModule,
-    CommonModule,
     ConfigModule.forRoot({
       load: [configuration],
+      isGlobal: true,
+      cache: true,
```
nitrosx, line 42: Does this section of code makes the configuration available everywhere without importing it in the individual modules?

src/auth/auth.module.ts
```diff
@@ -7,7 +7,7 @@ import { AuthController } from "./auth.controller";
 import { JwtModule } from "@nestjs/jwt";
 import { JwtStrategy } from "./strategies/jwt.strategy";
 import { LdapStrategy } from "./strategies/ldap.strategy";
-import { ConfigModule, ConfigService } from "@nestjs/config";
+import { ConfigService } from "@nestjs/config";
```
nitrosx, line 10: Why removing `ConfigModule`?
Also why keeping `ConfigService` when has been declared global in `app.module.ts`?

src/casl/casl.module.ts
```diff
@@ -1,7 +1,9 @@
 import { Module } from "@nestjs/common";
+import { ConfigModule } from "@nestjs/config";
 import { CaslAbilityFactory } from "./casl-ability.factory";
 
 @Module({
+  imports: [ConfigModule],
```
nitrosx, line 6: Why it is necessary import `ConfigModule` here?

src/common/utils.ts
```diff
@@ -819,6 +840,23 @@ export const addUpdatedByField = <T>(
...
+    ) as DatasetLookupKeysEnum[];
+  }
+
+  datasetLookupFields?.forEach((field) => {
+    DATASET_LOOKUP_FIELDS[field].$lookup.as = field;
+
+    pipeline.push(DATASET_LOOKUP_FIELDS[field]);
+  });
+};
+
```
nitrosx: IS this function specific to datasets? If yes, why put it here in common?

src/datasets/datasets.service.ts
```diff
@@ -31,6 +41,8 @@ import {
   PartialUpdateDatasetWithHistoryDto,
   UpdateDatasetDto,
 } from "./dto/update-dataset.dto";
+import { isEmpty } from "lodash";
```
nitrosx, line 45: Was loadsh already in use in the project?
If it is new, I'm wondering if we need to add another library for this function or there is something already available in nest.js or other library that are already in use in the project?

src/datasets/datasets.v4.controller.ts
```diff
@@ -0,0 +1,865 @@
...
+    } else if (group == Action.DatasetAttachmentRead) {
+      canDoAction =
+        ability.can(Action.DatasetAttachmentReadAny, DatasetClass) ||
+        ability.can(Action.DatasetAttachmentReadOwner, datasetInstance) ||
+        ability.can(Action.DatasetAttachmentReadAccess, datasetInstance) ||
+        ability.can(Action.DatasetAttachmentReadPublic, datasetInstance);
+    } else if (group == Action.DatasetAttachmentCreate) {
+      canDoAction =
+        ability.can(Action.DatasetAttachmentCreateAny, DatasetClass) ||
+        ability.can(Action.DatasetAttachmentCreateOwner, datasetInstance);
```
nitrosx: WE are not allowing attachment creation anymore

src/datasets/datasets.v4.controller.ts
```diff
@@ -0,0 +1,865 @@
...
+    } else if (group == Action.DatasetAttachmentUpdate) {
+      canDoAction =
+        ability.can(Action.DatasetAttachmentUpdateAny, DatasetClass) ||
+        ability.can(Action.DatasetAttachmentUpdateOwner, datasetInstance);
+    } else if (group == Action.DatasetAttachmentDelete) {
+      canDoAction =
+        ability.can(Action.DatasetAttachmentDeleteAny, DatasetClass) ||
+        ability.can(Action.DatasetAttachmentDeleteOwner, datasetInstance);
```
nitrosx: No more attachments changes and delete

src/datasets/datasets.v4.controller.ts
```diff
@@ -0,0 +1,865 @@
...
+        ability.can(Action.DatasetOrigdatablockCreateAny, DatasetClass) ||
+        ability.can(Action.DatasetOrigdatablockCreateOwner, datasetInstance);
+    } else if (group == Action.DatasetOrigdatablockUpdate) {
+      canDoAction =
+        ability.can(Action.DatasetOrigdatablockUpdateAny, DatasetClass) ||
+        ability.can(Action.DatasetOrigdatablockUpdateOwner, datasetInstance);
+    } else if (group == Action.DatasetOrigdatablockDelete) {
+      canDoAction =
+        ability.can(Action.DatasetOrigdatablockDeleteAny, DatasetClass) ||
+        ability.can(Action.DatasetOrigdatablockDeleteOwner, datasetInstance);
```
nitrosx: No OrigDatablock create, update or delete

src/datasets/datasets.v4.controller.ts
```diff
@@ -0,0 +1,865 @@
...
+        ability.can(Action.DatasetDatablockCreateAny, DatasetClass) ||
+        ability.can(Action.DatasetDatablockCreateOwner, datasetInstance);
+    } else if (group == Action.DatasetDatablockUpdate) {
+      canDoAction =
+        ability.can(Action.DatasetDatablockUpdateAny, DatasetClass) ||
+        ability.can(Action.DatasetDatablockUpdateOwner, datasetInstance);
+    } else if (group == Action.DatasetDatablockDelete) {
+      canDoAction =
+        ability.can(Action.DatasetDatablockDeleteAny, DatasetClass) ||
+        ability.can(Action.DatasetDatablockDeleteOwner, datasetInstance);
```
nitrosx: No datablocks create, update and delete

src/datasets/datasets.v4.controller.ts
```diff
@@ -0,0 +1,865 @@
...
+    ability.can(Action.DatasetCreate, DatasetClass),
+  )
+  @UseInterceptors(
+    new UTCTimeInterceptor<DatasetClass>(["creationTime"]),
+    new UTCTimeInterceptor<DatasetClass>(["endTime"]),
+    new FormatPhysicalQuantitiesInterceptor<DatasetClass>("scientificMetadata"),
+  )
+  @Post()
+  @ApiOperation({
+    summary: "It creates a new dataset which can be a raw or derived one.",
```
nitrosx: Change it to `It creates a new dataset. Type should be raw, derived or any of the customized types available in your instance`

src/datasets/datasets.v4.controller.ts
```diff
@@ -0,0 +1,865 @@
...
+    const foundDataset =
+      await this.datasetsService.findOneComplete(mergedFilters);
+
+    if (!foundDataset) {
+      // TODO: Do we want to throw here if the dataset is not found!?
+      // something like: throw new NotFoundException(`Dataset with provided filters: ${queryFilter} was not found. Please check your filter and try again`);
+    }
+
+    return foundDataset;
+  }
```
nitrosx: On second thoughts, FindOne can be achieved with FindAll with limit set to one.
Thoughts?

nitrosx (changes requested):
Few change requests and lots of questions

.github/workflows/release-and-publish-sdk.yml
martin-trajanovski: I can leave a comment there. But it is only used when we generate the sdk in the github actions. This `content` property is supported there in the openapi generator and that's why I need to disable it. But otherwise it brings a lot of improvements in the documentation and this was the only way to go forward and still being able to generate the sdk successfully.

src/config/configuration.ts
martin-trajanovski: It is easier to read the configuration variables the right way (as nestjs recommend) when they are grouped. In some places we need to read more than one of these variables and this way we can say for example: `this.accessGroups = this.configService.get<AccessGroupsType>("accessGroups");` and after that access any of the groups individually in the controller. You can check the `datasets.controller.ts` for a good example of this. Otherwise we will need to read every single variable individually which adds a lot more redundant code.

src/datasets/datasets.controller.ts
martin-trajanovski: It is not storing them locally but just getting them here and accessing where needed in the controller. It is cleaner this way I think. throughout the controller we just access by for example `this.datasetTypes` instead of `this.configService.get<string>("datasetTypes")`

src/datasets/datasets.module.ts
martin-trajanovski: You can see in the `app.module.ts` that I made the ConfigModule global and is cached. So it is quicker to access by other modules.

src/datasets/datasets.service.ts
martin-trajanovski: I will push new code where I add the right types right now.

src/datasets/pipes/pid-validation.pipe.ts
martin-trajanovski: Just to make it more clean and clear

src/datasets/types/dataset-filter-content.ts
martin-trajanovski: There is a description in the file why do we need this one. But it is only for improving swagger documentation. It doesn't influence on any logic or anything else

src/datasets/types/dataset-lookup.ts
martin-trajanovski: Yes maybe I can remove this as it is overwritten anyway

src/app.module.ts
martin-trajanovski: Yes it does. But not in the custom factories where we use it more specifically. So in some special cases the import is needed again. But it reduces the imports in the regular Dependency Injection for example controllers, services, and so on...

src/auth/auth.module.ts
martin-trajanovski: We need to keep it here as it is used in some custom factories that are run before the Dependency injection does it's thing and import it.

src/common/utils.ts
martin-trajanovski: I was thinking to extend it later maybe if needed for the other controllers. But you are right for now we can move it to the dataset service and if needed make it general later.

src/casl/casl.module.ts
martin-trajanovski: Maybe it is not. I will try to remove it.

src/datasets/datasets.service.ts
martin-trajanovski: It was already in use

src/app.module.ts
nitrosx: Let's make sure that we document this both in code and in documentation. Regina should be aware of this.

src/auth/auth.module.ts
nitrosx: Let's make sure to document this. I see potential source of confusion, especially for new contributors or if it is a long time since last time we have touched the code

src/datasets/datasets.v4.controller.ts
nitrosx: Following our discussion, let's keep it and add the limits to the filter, although limits.limit will not be used as it is automatically set to 1

martin-trajanovski:
This issue https://github.com/SciCatProject/scicat-backend-next/issues/1561 should be fixed here

src/datasets/datasets.service.ts
```diff
@@ -50,8 +73,185 @@ export class DatasetsService {
...
+        const canViewPublic = ability.can(
+          Action.ProposalsReadManyPublic,
+          ProposalClass,
+        );
+
+        return { canViewAny, canViewOwner, canViewAccess, canViewPublic };
+      }
+      case DatasetLookupKeysEnum.origdatablocks: {
+        const ability =
+          this.caslAbilityFactory.origDatablockInstanceAccess(user);
```
Junjiequan: I feel like we are putting too many authorization logic codes into the service file, which supposed to store mainly for database interaction code.
Wouldn't it better to have a `datasets.access.ts` file or something similar to separate the logics?

src/datasets/datasets.service.ts
```diff
@@ -79,6 +278,41 @@ export class DatasetsService {
...
 
+  async findAllComplete(
+    filter: FilterQuery<DatasetDocument>,
+  ): Promise<OutputDatasetDto[]> {
+    const whereFilter: FilterQuery<DatasetDocument> = filter.where ?? {};
+    const fieldsProjection: string[] = filter.fields ?? {};
+    const limits: QueryOptions<DatasetDocument> = filter.limits ?? {
+      limit: 10,
+      skip: 0,
+      sort: undefined,
```
Junjiequan: It's probably more practical to have `createAt` as dscending by default instead of undefined

src/datasets/pipes/filter-validation.pipe.ts
```diff
@@ -0,0 +1,75 @@
...
+      );
+
+      if (!isInAllowedKeys) {
+        // TODO: Should we clean the filter or throw bad request error???!!!
+        // unset(inValueParsed, key);
+        throw new BadRequestException(
+          `Property ${key} should not exist in the filter object`,
+        );
+      }
+    });
```
Junjiequan, line 71: I'd argue that we don't need to know that if a user is sending request with wrong filter keys or not, but user should know that they are using wrong query

src/instruments/instruments.controller.ts
```diff
@@ -34,6 +35,7 @@ import {
   filterExample,
   replaceLikeOperator,
 } from "src/common/utils";
+import { logger } from "@user-office-software/duo-logger";
```
Junjiequan: We don't need to`import { logger } from "@user-office-software/duo-logger";` ; if InternalServerErrorException has already been imported, as the duo-logger has already been implemented in the logger module

src/instruments/instruments.controller.ts
```diff
@@ -57,14 +59,20 @@ export class InstrumentsController {
...
+        );
+      } else {
+        logger.logException(
+          "Something went wrong while creating instrument",
+          error,
+        );
+
+        throw new InternalServerErrorException(
+          "Something went wrong. Please try again later.",
+        );
```
Junjiequan: We don't need to seperate two logs here, all exceptions will be sent to the graylogs once enabled
Can we keep the original error object in InternalServerErrorException for debugging?

src/instruments/instruments.controller.ts
```diff
@@ -142,12 +150,23 @@ export class InstrumentsController {
...
+      } else {
+        logger.logException(
+          "Something went wrong while updating instrument",
+          error,
+        );
+
+        throw new InternalServerErrorException(
+          "Something went wrong. Please try again later.",
+        );
```
Junjiequan: Same for this block

src/datasets/pipes/filter-validation.pipe.ts
martin-trajanovski: That's maybe a valid argument but not sure how to inform the user that the query was wrong without throwing an error

Junjiequan: Either way it requires frontend change to inform users. We can create a new PR for that later. 
I was thinking to return a string instead of error, but I'm not sure if that's a good solution.

martin-trajanovski: Yes maybe we can do that but I guess on the frontend we rarely see these kind of errors as we are going to define the query filter by our own and do not let the user input them by themselves. The error can be seen only if you try to query the endpoints containing these filters by directly using the url and passing your own parameters or using swagger. I guess in these situations we would like to see errors because users might try to pass some weird or suspicious queries.

Junjiequan: I think you are correct. Frontend do not need to show errors as the query filter is defined by us. 
I agree that tracking suspicious query from direct API requests will be helpful for us if we can get IP address for the requests

nitrosx (approved):

src/datasets/datasets-public.v4.controller.ts
nitrosx, line 1: I'm wondering if we should add `@AllowAny` decorator to all the endpoints, just to make it clear that anybody can access them.
Thoughts?

nitrosx (commented):
Minor comment

src/datasets/pipes/include-validation.pipe.ts
```diff
@@ -0,0 +1,42 @@
...
+
+function isJsonString(str: string) {
+  try {
+    JSON.parse(str);
+  } catch {
+    return false;
+  }
+  return true;
+}
+
```
Junjiequan: This utility function is better to be placed in  `src/common/utils.ts` for future re-use

Junjiequan (approved):
looks fine to me

PR #1559, opened 2024-12-11
feat(metrics): add metrics logging and configuration support

Junjiequan:
## Description

This PR introduces a new Metrics Module to enable request tracking and logging in the application. The module includes a configurable Access Tracking Middleware, which intercepts HTTP requests based on user-defined routes and logs access details for monitoring purposes.

#### Access Tracking Middleware:
- Intercepts HTTP requests and logs key details such as:
  - User ID(from JWT)
  - Client IP
  - Endpoint
  - Query params
  - Response time
- Identifies bot requests based on User-Agent and excludes them from logging.
- Logs all requests to Graylog
- 
#### Configurable Route Interception:
- The middleware behavior is configured via `metricsConfig.json` (In the root directory):
  - include: List of routes to track.
  - exclude: List of routes to ignore.
- Example:
```
{
  "include": [
    { "path": "datasets/fullquery", "method": 0, "version": "3" },
    { "path": "datasets/:id", "method": 0, "version": "3" }
  ],
  "exclude": [
    { "path": "datasets/fullfacet", "method": 0, "version": "3" }
  ]
}
  ```
For further info check [official documentation ](https://docs.nestjs.com/middleware)

#### Integration with Conditional Module:
- The MetricsModule is conditionally loaded based on the environment variable `METRICS_ENABLED`.

#### Cache Management:
- Uses a request cache to prevent excessive logging of repeated requests within a short time.
- Automatically resets the cache every 10 minutes to avoid memory leaks.

## Motivation
For monitoring user-behavior 

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### CODE REVIEW

File: src/app.module.ts [hunk 1]
```diff
@@ -42,6 +43,10 @@ import { LoggerModule } from "./loggers/logger.module";
     ConfigModule.forRoot({
       load: [configuration],
     }),
+    ConditionalModule.registerWhen(
+      MetricsModule,
+      (env: NodeJS.ProcessEnv) => env.METRICS_ENABLED === "yes",
+    ),
```
nitrosx, line 52: Can we use the configuration module that is imported on line 43?

src/metrics/middlewares/accessTracking.middleware.ts
nitrosx, line 1: Why not  having this code in the previous file?

metricsConfig.example.json
nitrosx, line 1: Is the format mandated by the functions in the tracking?
I wonder why we cannot use `path="api/v3/Datasets"` or something similar?

src/metrics/middlewares/accessTracking.middleware.ts
Junjiequan: That is to seperate the implmentation logics. It's not a common practice to include implementation logic within the module file based on the official documentation

metricsConfig.example.json
Junjiequan: This is suggested format from the documentation. 
```
consumer
  .apply(LoggerMiddleware)
  .exclude(
    { path: 'cats', method: RequestMethod.GET },
    { path: 'cats', method: RequestMethod.POST },
    'cats/(.*)',
  )
  .forRoutes(CatsController);
```

The benefit of having version is that we can do version: [3,4] to support both api/v3 and api/v4

src/app.module.ts
Junjiequan: I have checked [official documentation](https://docs.nestjs.com/techniques/configuration#conditional-module-configuration) and some other articles. `ConditionalModule.registerWhen` method do not support access to the configService. 
The cleanest way is to use `ConditionalModule.registerWhen` and fetch env directly, however, there's hack to achieve same result with DyncmiModule with increased complexity.

Junjiequan: image

nitrosx (approved):

PR #1574, opened 2024-12-17
fix(createDatasetDto): datasetName field in the CreateDatasetDto(obsoletes) should be required

Junjiequan:
## Description
This PR explicitly overrides the datasetName field in CreateDatasetDto to ensure it is required.

## Motivation
Previously, datasetName was optional in `CreateRawDatasetObsoleteDto` and `CreateDerivedDatasetObsoleteDto`, due to its inheritance from `UpdateDatasetObsoleteDto`. 

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/dto/create-dataset.dto.ts
+++ b/src/datasets/dto/create-dataset.dto.ts
@@ -29,5 +29,5 @@ export class CreateDatasetDto extends UpdateDatasetDto {
       "A name for the dataset, given by the creator to carry some semantic meaning. Useful for display purposes e.g. instead of displaying the pid.",
   })
   @IsString()
-  declare datasetName: string;
+  declare readonly datasetName: string;
 }
--- a/src/datasets/dto/create-dataset.dto.ts
+++ b/src/datasets/dto/create-dataset.dto.ts
@@ -1,6 +1,7 @@
 import { IsOptional, IsString } from "class-validator";
 import { ApiProperty } from "@nestjs/swagger";
 import { UpdateDatasetDto } from "./update-dataset.dto";
+import { Prop } from "@nestjs/mongoose";
 
 export class CreateDatasetDto extends UpdateDatasetDto {
   @ApiProperty({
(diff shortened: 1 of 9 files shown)


src/datasets/dto/create-dataset.dto.ts
```diff
@@ -20,4 +20,13 @@ export class CreateDatasetDto extends UpdateDatasetDto {
   })
   @IsString()
   readonly type: string;
+
```
martin-trajanovski: @Junjiequan I think that the `UpdateDatasetDto` includes this already and the `CreateDatasetDto` is just an extension of the Update one.

Junjiequan: Sorry, I have mistyped. It should be update-dataset-obsolete.dto.ts

src/instruments/instruments.controller.ts
```diff
@@ -143,10 +143,13 @@ export class InstrumentsController {
         updateInstrumentDto,
       );
       return instrument;
-    } catch (e) {
+    } catch (error) {
       throw new HttpException(
         "Instrument with the same unique name already exists",
         HttpStatus.BAD_REQUEST,
+        {
```
martin-trajanovski: I have already done some improvements here in the error handling. Check this PR: https://github.com/SciCatProject/scicat-backend-next/pull/1541/files#diff-36c6f6d5764ae45a2bb3b57d024b0c83c9ad714d79735e13787fb8104be763d6

martin-trajanovski (commented):
I am not sure if these changes are really needed as some of them are already there

src/instruments/instruments.controller.ts
Junjiequan: I have reverted it, let's use what you have done there, I can leave a comment there.
The rest should be fine

nitrosx (approved):

PR #1575, opened 2024-12-17
Jobs RabbitMQ Action connection info should come from ConfigService

despadam:
## Description
Previously, all information needed to set up the Jobs RabbitMQ action would come from `jobConfig`.
A better approach is to split this information: connection config info should be in  `.env` file, and only the queue info in the `jobConfig` file.

## Motivation
Extract the info from `ConfigService` after the NestJS refactor.

## Changes
While NestJs offers a RabbitMQ microservice, unfortunately it does not cover all necessary options needed for Jobs. More specifically, we are configurating `exchange` and `key`, which is not offered by that library. So instead, we stick to using the `amqplib` library as before, but only request `queue`, `exchange` and `key` to be defined in `jobConfig`. Which means that the `.env` file should now look like this example for RabbitMQ Job Action to be set up properly:
```
RABBITMQ_ENABLED="yes"
RABBITMQ_HOSTNAME="rabbitmq"
RABBITMQ_PORT=5672
RABBITMQ_USERNAME="guest"
RABBITMQ_PASSWORD="guest"
```
And the job action definition in the `jobConfig` file, should look like this:
```
    - actionType: rabbitmq
      exchange: myexchange
      queue: myqueue
      key: mykey
```

Note: make sure you have a running RabbitMQ instance for testing this. For example, with `scicat-ci`: https://github.com/despadam/scicat-ci/commit/969959057ea13fa1fb8b01ba4c791f015ab3850f

---
To make sure that messages are being sent, you can connect to the RabbitMQ admin page (http://localhost:15672/ by default) with the username and password provided in the `.env` file. There, after a job creation or update, you will see that the queue has been created and messages are incoming:
image
You can also explore the queue contents, as seen here:
image

### DIFF
--- a/src/common/rabbitmq/rabbitmq.service.ts
+++ b/src/common/rabbitmq/rabbitmq.service.ts
@@ -1,0 +1,123 @@
+import amqp, { Connection, Channel } from "amqplib/callback_api";
+import { Injectable, Logger, OnModuleDestroy } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
+
+/**
+ * Service for publishing messages to a RabbitMQ queue.
+ */
+@Injectable()
+export class RabbitMQService implements OnModuleDestroy {
+  private connectionOptions: amqp.Options.Connect;
+  private connection: Connection;
+  private channel: Channel;
+
+  constructor(private configService: ConfigService) {
+    Logger.log("Initializing RabbitMQService.", "RabbitMQService");
+
+    const rabbitMqEnabled =
+      (this.configService.get<string>("rabbitMq.enabled") === "yes");
+
+    if (rabbitMqEnabled) {
+      const hostname = this.configService.get<string>("rabbitMq.hostname");
+      const port = this.configService.get<number>("rabbitMq.port");
+      const username = this.configService.get<string>("rabbitMq.username");
+      const password = this.configService.get<string>("rabbitMq.password");
+
+      if (!hostname || !port || !username || !password) {
+        Logger.error(
+          "RabbitMQ is enabled but missing one or more config variables.",
+          "RabbitMQService",
+        );
+      } else {
+        this.connectionOptions = {
+          protocol: "amqp",
+          hostname: hostname,
+          port: port,
+          username: username,
+          password: password,
+        };
+
+        amqp.connect(
+          this.connectionOptions,
+          (connectionError: Error, connection: Connection) => {
+            if (connectionError) {
+              Logger.error(
+                "Connection error in RabbitMQService: " +
+                  JSON.stringify(connectionError.message),
+                "RabbitMQService",
+              );
+              return;
+            }
+            this.connection = connection;
+    
+            this.connection.createChannel(
+              (channelError: Error, channel: Channel) => {
+                if (channelError) {
+                  Logger.error(
+                    "Channel error in RabbitMQService: " +
+                      JSON.stringify(channelError.message),
+                    "RabbitMQService",
+                  );
+                  return;
+                }
+                this.channel = channel;
+                Logger.log(this.channel);
+              }
+            );
+          },
+        );
+      }
+    } else {
+      Logger.error("RabbitMQ is not enabled.", "RabbitMQService");
+    }
+  }
+
+  connect(
+    queue: string,
+    exchange: string,
+    key: string,
+  ) {
+    try {
+      this.channel.assertQueue(queue, { durable: true });
+      this.channel.assertExchange(exchange, "topic", {
+        durable: true,
+      });
+      this.channel.bindQueue(queue, exchange, key);
+    } catch (error) {
+      Logger.error(
+        `Could not connect to RabbitMQ queue ${queue} with exchange ${exchange} and key ${key}.`,
+        "RabbitMQService",
+      );
+      Logger.error(error);
+    }
+  }
+
+  sendMessage(
+    queue: string,
+    message: string,
+  ) {
+    try {
+      this.channel.sendToQueue(queue, Buffer.from(message), {
+        persistent: true
+      });
+    } catch (error) {
+      Logger.error(
+        `Could not send message to RabbitMQ queue ${queue}.`,
+        "RabbitMQService",
+      );
+    }
+  }
+
+  onModuleDestroy() {
+    if (this.channel) {
+      this.channel.close(() => {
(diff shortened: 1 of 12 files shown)


src/config/job-config/actions/rabbitmqaction/rabbitmqaction.ts
```diff
@@ -6,37 +6,64 @@ import {
...
+          "RabbitMQJobAction",
+        );
+      } else {
+        this.connection = {
+          protocol: "amqp",
+          hostname: hostname,
+          port: port,
+          username: username,
+          password: password,
+        };
```
sbliven: In v3 this was provided as a connection string, like "amqp://USER:PASS@HOST:PORT". Is splitting it into 5 env variables necessary?

src/config/job-config/actions/rabbitmqaction/rabbitmqaction.ts
```diff
@@ -6,37 +6,64 @@ import {
...
-      password: options.password,
-    };
-    this.binding = {
-      exchange: options.exchange,
-      queue: options.queue,
-      key: options.key,
-    };
-  }
+    const rabbitMqEnabled =
+      this.configService.get<string>("rabbitMq.enabled") === "yes"
```
sbliven: Use `mathjs.parseBool` for this

sbliven (changes requested):
This will create a new messagebroker instance for each action. I think it would be better to wrap this into a nest module so it can be shared.

AppModule also constructs a RabbitMQMessageBroker, coming from [@user-office-software/duo-message-broker](https://github.com/UserOfficeProject/user-office-lib) rather than `amqplib` directly. It seems to be ESS-specific functionality for their proposal system. I was going to suggest using the same broker instance for that too, but after looking at the code I'm not sure it would work. We should discuss that code with the ESS people, but perhaps not in this PR.

src/config/job-config/actions/rabbitmqaction/rabbitmqaction.ts
despadam: Some of those variables are also used to configure the other RabbitMQ message broker in app.module. So for now I will leave this as is and later, when we have a discussion about that, we can bring this up.

src/config/job-config/actions/rabbitmqaction/rabbitmqaction.ts
despadam: I don't think that this exists, are you sure? So I just simplified the logic

src/config/job-config/actions/rabbitmqaction/rabbitmqaction.ts
```diff
@@ -1,91 +1,43 @@
...
-          );
-
-          channel.close(() => {
-            connection.close();
-          });
-        });
-      },
-    );
+    this.rabbitMQService.connect(this.queue, this.exchange, this.key);
+    this.rabbitMQService.sendMessage(this.queue, JSON.stringify(job));
```
sbliven, line 41: We should double check that it's thread-safe. I think it should be ok since they aren't async.

sbliven (approved):
Looks good, thanks for the update!

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -0,0 +1,116 @@
...
+      this.channel.assertExchange(exchange, "topic", {
+        durable: true,
+      });
+      this.channel.bindQueue(queue, exchange, key);
+    } catch (error) {
+      Logger.error(
+        `Could not connect to RabbitMQ queue ${queue} with exchange ${exchange} and key ${key}.`,
+        "RabbitMQService",
+      );
+      Logger.error(error);
```
sbliven, line 87: I would throw the error and log it higher up if needed.

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -0,0 +1,116 @@
...
+                }
+                this.channel = channel;
+                Logger.log(this.channel);
+              },
+            );
+          },
+        );
+      }
+    } else {
+      Logger.error("RabbitMQ is not enabled.", "RabbitMQService");
```
sbliven, line 71: What's the expected behavior if they don't have RabbitMQ enabled? I think the service gets instantiated either way. Maybe the constructor should succeed (without logging) but there should be an error if connect() is called or if a rabbitmqaction is created.

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -0,0 +1,116 @@
...
+    } catch (error) {
+      Logger.error(
+        `Could not connect to RabbitMQ queue ${queue} with exchange ${exchange} and key ${key}.`,
+        "RabbitMQService",
+      );
+      Logger.error(error);
+    }
+  }
+
+  sendMessage(queue: string, message: string) {
```
sbliven, line 91: What happens if a message is sent without first calling connect? If they are always called in pairs would it make more sense to merge them into one method?

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -0,0 +1,116 @@
...
+    try {
+      this.channel.sendToQueue(queue, Buffer.from(message), {
+        persistent: true,
+      });
+    } catch (error) {
+      Logger.error(
```
sbliven, line 97: As above, throw the error

src/config/job-config/actions/rabbitmqaction/rabbitmqaction.ts
```diff
@@ -1,91 +1,43 @@
...
-  constructor(options: RabbitMQJobActionOptions) {
+  getActionType(): string {
+    return actionType;
+  }
+
+  constructor(
+    private readonly rabbitMQService: RabbitMQService,
+    options: RabbitMQJobActionOptions,
+  ) {
     Logger.log(
```
sbliven, line 26: Remove logging

src/config/job-config/actions/rabbitmqaction/rabbitmqaction.ts
```diff
@@ -1,91 +1,43 @@
...
-    this.binding = {
-      exchange: options.exchange,
-      queue: options.queue,
-      key: options.key,
-    };
-  }
-
-  getActionType(): string {
-    return actionType;
+    this.queue = options.queue;
```
sbliven, line 30: should throw an error if rabbitMQ isn't enabled

PR #1576, opened 2024-12-18
feat: introduce public dataset endpoints

martin-trajanovski:
## Description
This PR aims to separate the public dataset endpoints and refactor the dataset relation fields access.

## Motivation
Querying the public datasets was part of the main endpoints which was a bit confusing as in swagger it shows that it needs authentication but if you are not logged in it returns the public datasets back.

## Changes:

* adding new dataset controller for the public datasets and new dataset-access.service.ts for the relational fields access control.

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/casl/casl-ability.factory.ts
+++ b/src/casl/casl-ability.factory.ts
@@ -98,17 +98,34 @@ export class CaslAbilityFactory {
       /*  unauthenticated users
       **/
 
+      can(Action.DatasetReadManyPublic, DatasetClass);
+      can(Action.DatasetReadOnePublic, DatasetClass, {
+        isPublished: true,
+      });
+      // -
+      can(Action.DatasetAttachmentReadPublic, DatasetClass, {
+        isPublished: true,
+      });
+      // -
+      can(Action.DatasetOrigdatablockReadPublic, DatasetClass, {
+        isPublished: true,
+      });
+      // -
+      can(Action.DatasetDatablockReadPublic, DatasetClass, {
+        isPublished: true,
+      });
+
       cannot(Action.DatasetCreate, DatasetClass);
-      can(Action.DatasetRead, DatasetClass);
+      cannot(Action.DatasetRead, DatasetClass);
       cannot(Action.DatasetUpdate, DatasetClass);
       // -
       cannot(Action.DatasetAttachmentCreate, DatasetClass);
-      can(Action.DatasetAttachmentRead, DatasetClass);
+      cannot(Action.DatasetAttachmentRead, DatasetClass);
       cannot(Action.DatasetAttachmentUpdate, DatasetClass);
       cannot(Action.DatasetAttachmentDelete, DatasetClass);
       // -
       cannot(Action.DatasetOrigdatablockCreate, DatasetClass);
-      can(Action.DatasetOrigdatablockRead, DatasetClass);
+      cannot(Action.DatasetOrigdatablockRead, DatasetClass);
       cannot(Action.DatasetOrigdatablockUpdate, DatasetClass);
       // -
       cannot(Action.DatasetDatablockCreate, DatasetClass);
--- a/src/datasets/datasets-public.v4.controller.ts
+++ b/src/datasets/datasets-public.v4.controller.ts
@@ -1,0 +1,245 @@
+import { Controller, Get, Param, Query, HttpStatus } from "@nestjs/common";
+import {
+  ApiExtraModels,
+  ApiOperation,
+  ApiParam,
+  ApiQuery,
+  ApiResponse,
+  ApiTags,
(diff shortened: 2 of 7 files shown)


src/casl/casl-ability.factory.ts
```diff
@@ -98,17 +98,34 @@ export class CaslAbilityFactory {
       /*  unauthenticated users
       **/
 
+      can(Action.DatasetReadManyPublic, DatasetClass);
+      can(Action.DatasetReadOnePublic, DatasetClass, {
+        isPublished: true,
+      });
```
Junjiequan, line 104: Do we need separated DatasetReadPublic permissions? One DatasetReadPublicAny feels enough for me. 
We can discuss with Max about that later for the final merge into master

src/datasets/datasets-public.v4.controller.ts
```diff
@@ -0,0 +1,245 @@
...
+import { CountApiResponse } from "src/common/types";
+import { DatasetLookupKeysEnum } from "./types/dataset-lookup";
+import { IncludeValidationPipe } from "./pipes/include-validation.pipe";
+import { FilterValidationPipe } from "./pipes/filter-validation.pipe";
+import { getSwaggerDatasetFilterContent } from "./types/dataset-filter-content";
+
+@ApiExtraModels(HistoryClass, TechniqueClass, RelationshipClass)
+@ApiTags("datasets public v4")
+@Controller({ path: "datasets/public", version: "4" })
+export class DatasetsPublicV4Controller {
```
Junjiequan, line 35: As we discussed during your demo, naming of v4 may not necessary. Users/Devs can choose image version s to accommodate their needs.
Again, this can be further discussed before merge into master with Max

src/datasets/datasets-public.v4.controller.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    this.addPublicFilter(parsedFilter);
+
+    const foundDataset =
+      await this.datasetsService.findOneComplete(parsedFilter);
+
+    if (!foundDataset) {
+      // TODO: Do we want to throw here if the dataset is not found!?
+      // something like: throw new NotFoundException(`Dataset with provided filters: ${queryFilter} was not found. Please check your filter and try again`);
+    }
```
Junjiequan: Let's keep what has been implemented on the datasets.controller for now. Even if we need to change it later it will be easier for us at that point.

src/datasets/datasets.service.ts
```diff
@@ -259,13 +259,24 @@ export class DatasetsService {
   ): Promise<OutputDatasetDto | null> {
     const whereFilter: FilterQuery<DatasetDocument> = filter.where ?? {};
     const fieldsProjection: string[] = filter.fields ?? {};
+    const limits: QueryOptions<DatasetDocument> = filter.limits ?? {
+      skip: 0,
+      sort: undefined,
```
Junjiequan: I suggest to sort createdAt latest by default instead of undefined

src/datasets/datasets.v4.controller.ts
```diff
@@ -396,6 +377,7 @@ export class DatasetsV4Controller {
 
   // GET /fullfacets
   @UseGuards(PoliciesGuard)
+  // TODO: Check if this should be closed or oppened endpoint for public!?
```
Junjiequan: Fullfact is currently used in two places, Filter UI dropdown and datasets total.
I'd say it'd be easier to maintain frontend by including fullfacet endpoint for public 
image
image

Junjiequan (commented):
Looks fine to me in general.
The only thing I'm not certain is that public dataset has no fullfacet endpoint in the controller. Maybe you can test it locally to see if the public fullfact is necessary or not

src/datasets/datasets-public.v4.controller.ts
martin-trajanovski: It's good to have it here

src/casl/casl-ability.factory.ts
martin-trajanovski: If you check the other permissions they all include both ReadMany and ReadOne. As the one is including the field `isPublished: true`.

src/datasets/datasets.v4.controller.ts
martin-trajanovski: I have added `fullfacet` to the public endpoints as well now. Let's see how it works later in the frontend all this together when we try to switch to the new v4 endpoints

src/datasets/datasets-public.v4.controller.ts
martin-trajanovski: This should be fixed in the last commit

Junjiequan (approved):
Looks good to me

Issue #154, opened 2022-09-28
Missing fallback to datasets default at dataset creation


minottic:
## Issue Name

Missing fallback to datasets default at dataset creation

### Summary

[This part](https://github.com/SciCatProject/backend/blob/master/common/models/dataset.js#L279-L444) hasn't been ported to the new backend

### Steps to Reproduce

Look in the code for defaults, e.g. `IN=medium,AV=low,CO=low`

### Expected Behaviour

Fallback to defaults should be implemented in the new be

minottic:
should be fixed by af31237855ca26ce612f273eda6f299294c3ecb1, b7ccf754ad3ae5ae885066f5336661319b2c0380, 997917fc06b2cf9e684262639be6325515044965

PR #1580, opened 2024-12-20
feat: add api tests for the new controllers

martin-trajanovski:
## Description
This is PR that adds unit and api tests for the dataset v4 controllers.

## Motivation
After adding the new controllers we need to test all this refactor and also the public endpoints and access to the data.

## Fixes

* https://jira.ess.eu/browse/SWAP-4366

## Changes:

* adding tests for dataset v4 controllers

## Tests included

- [x] Passing? 

## Documentation


martin-trajanovski:
@nitrosx or @Junjiequan maybe this one should be reviewed and merged before the https://github.com/SciCatProject/scicat-backend-next/pull/154 as it pushes changes into that branch first.

src/datasets/datasets.v4.controller.ts
```diff
@@ -320,11 +319,16 @@ export class DatasetsV4Controller {
...
     @Req() request: Request,
     @Body(PidValidationPipe)
-    createDatasetDto: CreateDatasetDto,
+    createDatasetDto: object,
   ) {
+    const createDatasetDtoInstance = plainToInstance(
+      CreateDatasetDto,
+      createDatasetDto,
+    );
+
```
nitrosx, line 328: Can you explain why this explicit conversion is needed?

test/DatasetV4.js
```diff
@@ -0,0 +1,985 @@
...
+          const [firstDataset] = res.body;
+
+          firstDataset.should.have.property("pid");
+          firstDataset.should.have.property("instruments");
+          firstDataset.should.have.property("proposals");
+          firstDataset.should.not.have.property("datablocks");
+        });
+    });
+
+    it("0205: should fetch all dataset relation fields if provided in the filter", async () => {
```
nitrosx: Should change `should fetch all dataset relation fields if provided in the filter` to `should fetch all datasets with related items when requested with all relations"

test/DatasetV4.js
```diff
@@ -0,0 +1,985 @@
...
+          res.body.should.be.a("object");
+
+          res.body.should.have.property("pid");
+          res.body.should.have.property("datasetName");
+          res.body.datasetName.should.be.eq(updatedDataset.datasetName);
+        });
+    });
+  });
+
+  describe("Datasets v4 update tests", () => {
```
nitrosx: There is already another section with `Datasets v4 update tests`. Why two?

nitrosx: Change to `Delete`. :)

test/DatasetV4Access.js
nitrosx, line 1: AS first version it looks good.

nitrosx (changes requested):

src/datasets/datasets.v4.controller.ts
martin-trajanovski: I think the main usage of this endpoint is to check if the input you provide is a valid dataset and return true/false based on that. If we put the `CreateDatasetDto` as a type for the input parameter the endpoint returns 400 (Bad request) all the time until we provide a valid input. And then once we provide a valid input it will always return `true` as a value. So my point is that in the previous case this endpoint will never return `false` as a value because of the input parameter validation. I think it makes more sense to accept any input as `object` in this endpoint and validate it inside as we are already doing and then return `true/false` based on the result. @nitrosx please let me know if this explains my change and if it is a reasonable one.

nitrosx (approved):

src/datasets/datasets.v4.controller.ts
nitrosx: @martin-trajanovski cristal clear!!!

PR #1581, opened 2024-12-20
RabbitMQ service: fix logging and error handling

despadam:
## Description
The NestJS `RabbitMQService` always gets initialized when the application starts. However, the RabbitMQ connection itself should not get initialized if `RABBITMQ_ENABLED=no` in the `.env` file. The following changes will initialize the connection only when `RABBITMQ_ENABLED=yes`, while also printing the relevant logs. Then, will also throw relevant errors in the case that we are unable to connect to the queue or send a message to it. 

On the other hand, if based on the `jobConfig` definition, we wish that a job will use the `RabbitMQJobAction`, then in the action's constructor we check that the service is enabled. If not, we throw an error.

## Motivation
Addressing @sbliven 's latest comments after merging PR https://github.com/SciCatProject/scicat-backend-next/pull/1575

### DIFF
--- a/src/common/rabbitmq/rabbitmq.service.ts
+++ b/src/common/rabbitmq/rabbitmq.service.ts
@@ -1,12 +1,17 @@
 import amqp, { Connection, Channel } from "amqplib/callback_api";
-import { Injectable, Logger, OnModuleDestroy, OnApplicationShutdown  } from "@nestjs/common";
+import {
+  Injectable,
+  Logger,
+  OnModuleDestroy,
+  OnApplicationShutdown
+} from "@nestjs/common";
 import { ConfigService } from "@nestjs/config";
 
 /**
  * Service for publishing messages to a RabbitMQ queue.
  */
 @Injectable()
-export class RabbitMQService implements OnModuleDestroy, OnApplicationShutdown  {
+export class RabbitMQService implements OnModuleDestroy, OnApplicationShutdown {
   private connectionOptions: amqp.Options.Connect;
   private connection: Connection;
   private channel: Channel;
@@ -92,10 +97,9 @@ export class RabbitMQService implements OnModuleDestroy, OnApplicationShutdown  
         persistent: true,
       });
     } catch (error) {
-      throw new Error(
-        `Could not send message to RabbitMQ queue ${queue}.`,
-        { cause: error },
-      );
+      throw new Error(`Could not send message to RabbitMQ queue ${queue}.`, {
+        cause: error
+      });
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -61,14 +67,11 @@ export class RabbitMQService implements OnModuleDestroy {
...
-                Logger.log(this.channel);
               },
             );
           },
         );
       }
-    } else {
-      Logger.error("RabbitMQ is not enabled.", "RabbitMQService");
     }
   }
```
sbliven, line 71: Should `connect` be private?

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -80,37 +83,44 @@ export class RabbitMQService implements OnModuleDestroy {
...
         `Could not connect to RabbitMQ queue ${queue} with exchange ${exchange} and key ${key}.`,
-        "RabbitMQService",
+        { cause: error },
       );
-      Logger.error(error);
     }
   }
 
-  sendMessage(queue: string, message: string) {
+  sendMessage(queue: string, exchange: string, key: string, message: string) {
```
sbliven, line 87: Should it throw an error if the service isn't enabled (maybe within this.connect)?

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -1,23 +1,29 @@
...
 /**
  * Service for publishing messages to a RabbitMQ queue.
  */
 @Injectable()
-export class RabbitMQService implements OnModuleDestroy {
+export class RabbitMQService implements OnModuleDestroy, OnApplicationShutdown {
   private connectionOptions: amqp.Options.Connect;
   private connection: Connection;
   private channel: Channel;
+  rabbitMqEnabled: boolean;
```
sbliven: Maybe this should be spelled simply `RabbitMQService.enabled`?

sbliven: should also be marked `readonly`

sbliven (approved):
It's a little weird (but not necessarily wrong) that RabbitMqService gets instantiated even if it's not enabled. Maybe you should add a comment documenting that users need to check `rabbitMQEnabled` before using the service (or else throw errors everywhere if it's disabled).

As we discussed, it's also worth noting that this implementation uses a single channel for the whole app. We might get better performance in the future by refactoring so that each action (or even each request?) gets its own channel. However this seems a bit premature at this point.

despadam:
Thanks for your comments @sbliven. After the latest changes, `RabbitMQService` is now only getting initialized when rabbitMQ is enabled in the `.env` file. Then, `RabbitMQJobAction`'s constructor will also check that rabbitMQ is enabled, otherwise will throw an error. This also protects us from mistakenly having rabbitMQ disabled in the `.env` file, but defining rabbitMQ actions in the `jobConfig` file.

sbliven (approved):
Looks good! Sorry for the delay reviewing this.

PR #1599, opened 2025-01-06
Release 6-1-2025 

Junjiequan:
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/.github/workflows/upload-sdk-artifact.yml
+++ b/.github/workflows/upload-sdk-artifact.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   NODE_VERSION: 20.x
-  SDK_VERSION: latest
+  SDK_VERSION: 0.0.1
 
 jobs:
   start-backend-and-upload-swagger-schema:
@@ -48,7 +48,7 @@ jobs:
       - start-backend-and-upload-swagger-schema
     strategy:
       matrix:
-        generator: [python, typescript-angular, python-pydantic-v1]
+        generator: [python, typescript-angular, typescript-fetch]
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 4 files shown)


.github/workflows/upload-sdk-artifact.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   NODE_VERSION: 20.x
-  SDK_VERSION: latest
+  SDK_VERSION: 0.0.1
```
sbliven, line 10: Does this need to be updated regularly in the future? If so that should be documented in the README.

sbliven (approved):
@Junjiequan Can you add more detail to these release PRs? I feel like they should include a draft of the release notes.

Since this release seems only CI changes for the SDK generator it looks ok.

nitrosx (approved):

.github/workflows/upload-sdk-artifact.yml
Junjiequan: This should be a one-time fix and will be used only for local development. The change is intended to address the issue with the Python SDK where pip install does not recognize tags labeled as “latest” and requires a numeric version instead.

Junjiequan:
> @Junjiequan Can you add more detail to these release PRs? I feel like they should include a draft of the release notes.
> 
> Since this release seems only CI changes for the SDK generator it looks ok.

@sbliven @nitrosx 
I agree that we should have a release description for breaking changes. For minor and patch updates though, I believe auto generated changelog should suffice, if not we should try to make them suffice by setting rules. 
The intention of creating release PR was not to collect and summarize a draft for all the changes, it was to automize the release process and easier roll-back. I do think there's room for automation process improvement and would be happy to hear suggetions

sbliven:
@Junjiequan Is is possible to see the auto-generated changelog before merging? If so maybe we could have a PR template for releases that would link this.

Issue #1590, opened 2024-12-30
Use openapi cli plugin to automate swagger api annotation


nitrosx:
## Summary
As highlighted by @emigun, maintenance and consistency of the swagger api documentation is expensive and prone to human error as we need to enter the same information in multiple location.

## Solution
We propose to adopt the [OpenAPI CLI plugin](https://docs.nestjs.com/openapi/cli-plugin), which reduce information duplication, code cluttering and inconsistencies, and allowing complete freedom in configuration where needed.

##Additional information
Below I included an edited version of the [post](https://github.com/SciCatProject/scicat-backend-next/pull/1579#issuecomment-2556686282) reported by @emigun in PR #1579

nitrosx:
Perhaps we should consider using https://docs.nestjs.com/openapi/cli-plugin instead of annotating manually?
It is already enabled.
I have tested it a bit and it seems to work nicely.

A fairly advanced example:
Let's change
```
  @ApiProperty({
    type: "array",
    items: { $ref: getSchemaPath(TechniqueClass) },
    required: false,
    default: [],
    description: "Stores the metadata information for techniques.",
  })
  @IsOptional()
  @IsArray()
  @ValidateNested({ each: true })
  @Type(() => CreateTechniqueDto)
  readonly techniques?: TechniqueClass[] = [];
```

to (remove ApiProperty)

```
  @IsOptional()
  @IsArray()
  @ValidateNested({ each: true })
  @Type(() => CreateTechniqueDto)
  readonly techniques?: TechniqueClass[] = [];
```

That would change the openapi json from:
```
"techniques": {
  "type": "array",
  "items": {
    "$ref": "#/components/schemas/TechniqueClass"
  },
  "default": [],
  "description": "Stores the metadata information for techniques."
},
```

to

```
"techniques": {
  "default": [],
  "type": "array",
  "items": {
    "$ref": "#/components/schemas/TechniqueClass"
  }
}
```

As you can see everything was generated (except description, see below)

I have tried with arrays, dates and ordinary string/number and it seems to work. It also reads the `?` for optional fields and correctly identifies the required ones.

We could change nest-cli.json to:
```{
  "collection": "@nestjs/schematics",
  "sourceRoot": "src",
  "compilerOptions": {
    "plugins": [{
      "name": "@nestjs/swagger",
      "options": {
        "introspectComments": true
      }
    }]
  }
}
```
in order to generate the field "description" based on a comment above. Or you could still use
```
[...]

PR #1629, opened 2025-01-20
refactor(proposal): use openapi plugin

emigun:
## Description
Use openapi plugin in proposal module

## Motivation
Less clutter in code and less risk for inconstensies between openapi spec and the actual behaviour.

## Fixes
Related to #1590 

## Changes:
Removed all `@ApiProperty` decorators in proposal module.

The first commit doesn't change the openapi spec. The second commit, which touches three fields, does make some changes which we can discuss.

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/nest-cli.json
+++ b/nest-cli.json
@@ -5,7 +5,12 @@
     "plugins": [{
       "name": "@nestjs/swagger",
       "options": {
-        "dtoFileNameSuffix": [".dto.ts", "sample.schema.ts"],
+        "dtoFileNameSuffix": [
+          ".dto.ts",
+          "sample.schema.ts",
+          "proposal.schema.ts",
+          "measurement-period.schema.ts"
+        ],
         "introspectComments": true
       }
     }]
--- a/src/proposals/dto/create-measurement-period.dto.ts
+++ b/src/proposals/dto/create-measurement-period.dto.ts
@@ -1,4 +1,3 @@
-import { ApiProperty } from "@nestjs/swagger";
 import { IsDateString, IsOptional, IsString } from "class-validator";
 
 export class CreateMeasurementPeriodDto {
--- a/src/proposals/dto/create-measurement-period.dto.ts
+++ b/src/proposals/dto/create-measurement-period.dto.ts
@@ -2,36 +2,27 @@ import { ApiProperty } from "@nestjs/swagger";
 import { IsDateString, IsOptional, IsString } from "class-validator";
 
 export class CreateMeasurementPeriodDto {
-  @ApiProperty({
-    type: String,
-    required: true,
-    description:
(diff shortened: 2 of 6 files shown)


emigun:
There are two commits. The first commit doesn't change anything in the swagger spec. The second one does, for proposal.schema.ts.

For field metadata:
Is it on purpose that the type in ApiProperty is MongooseSchema.Types.Mixed? Seems more reasonable to just have object , for the api users. Also added a default value.

For field parentProposalId :
Changed to string | null since it is nullable, I set the default value to null. Also, changed to optional field because it has a default anyway? What would the point be of having a default if it is required?

For field type:
Same here, changed to optional field since there is a default and I set the default. It is of course possible to only have the default in the ApiProperty, but probably not any real practical difference on setting it?

nitrosx (commented):
**For field metadata:**

> Is it on purpose that the type in ApiProperty is MongooseSchema.Types.Mixed? Seems more reasonable to just have object , for the api users. Also added a default value.

I do not know why we set it to this type. I agree that using _Object_ is easier. Could you please try switching it to _object_ and verify that all the test pass?

**For field parentProposalId :**

> Changed to string | null since it is nullable, I set the default value to null. Also, changed to optional field because it has a default anyway? What would the point be of having a default if it is required?

OK for the type. I would make required in the schema as you have the default value. It is optional in the dto.

**For field type:**

> Same here, changed to optional field since there is a default and I set the default. It is of course possible to only have the default in the ApiProperty, but probably not any real practical difference on setting it?

Same as _parentPRoposalId_, optional in the dto, but required in the schema set to the default value.

emigun:
> **For field metadata:**
> 
> > Is it on purpose that the type in ApiProperty is MongooseSchema.Types.Mixed? Seems more reasonable to just have object , for the api users. Also added a default value.
> 
> I do not know why we set it to this type. I agree that using _Object_ is easier. Could you please try switching it to _object_ and verify that all the test pass?

It is already object since I removed the ApiProperty decorator. The openapi plugin will read the Record type.

> 
> **For field parentProposalId :**
> 
> > Changed to string | null since it is nullable, I set the default value to null. Also, changed to optional field because it has a default anyway? What would the point be of having a default if it is required?
> 
> OK for the type. I would make required in the schema as you have the default value. It is optional in the dto.
> 

I have looked around a bit on the web and it seems the normal way is to not use `required` when you have a default.

`parentProposalId` and `type` are a bit different in that parentProposalId had `required: false` in the Prop but type had `required: true` in the Prop decorator. None of them had a `?` on the field. Perhaps just drop required on both of them? Without a `?` on the property it should be in the required list in the swagger.

nitrosx (approved):

PR #1691, opened 2025-02-07
fix: or filter in the proposals count endpoint to be able to get the related proposals easier

martin-trajanovski:
## Description
Fix the permissions for count, fullquery and fullfacet proposal endpoints and allow $or filter to work in the proposal count.

## Motivation
Permissions were not correct and the $or filter in count didn't work as expected

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/proposals/proposals.controller.ts
+++ b/src/proposals/proposals.controller.ts
@@ -233,12 +233,42 @@ export class ProposalsController {
           ProposalClass,
         );
         if (canViewAccess) {
-          mergedFilters.where["$or"] = [
-            { ownerGroup: { $in: user.currentGroups } },
-            { accessGroups: { $in: user.currentGroups } },
-          ];
+          if (!mergedFilters.where["$and"]) {
+            if (mergedFilters.where["$or"]) {
+              mergedFilters.where["$and"] = [
+                {
+                  $or: mergedFilters.where["$or"],
+                },
+                {
+                  $or: [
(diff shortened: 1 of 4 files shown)


src/proposals/proposals.controller.ts
```diff
@@ -390,49 +420,51 @@ export class ProposalsController {
       "Database filters to apply when retrieving count for proposals",
     required: false,
     type: ProposalCountFilters,
-    example: `{ fields: ${proposalsFullQueryExampleFields}}`,
+    example: `{ fields: ${proposalsFullQueryExampleFields}, filter: ${filterExample}}`,
```
Junjiequan: Below is the value of the filterExample. Is it correct example for the filter field?
```
export const filterExample =
  '{ "where": { "field": "value" }, "include": [ { "relation": "target" } ], "fields": ["field1", "field2"], "limits": {"limit": 1, "skip": 1, "order": "asc"}}';
```

Junjiequan (commented):
I've left one question, otherwise looks fine to me

src/proposals/proposals.controller.ts
martin-trajanovski: Yes you are right this should be fixed. I thought it is standard filter example but from that filter we use only the where filter actually.

Junjiequan (approved):

Issue #966, opened 2023-12-20
Validation of scientificMetadata


sbliven:
### Summary

Enable validation of scientificMetadata against a schema. The schema should be configurable for each site and default to no validation. In some cases multiple schema might be desirable, with different features available depending on the schema.

### Motivation

- Boundary conditions (eg matching a proposal system format, interoperability with OpenEM databases, etc)
- Frontend visualization (tabular or more advanced)
- Semantic tagging (eg [comment by @bpedersen2](https://github.com/SciCatProject/scicat-backend-next/issues/939#issuecomment-1860166149))
- Advanced search (eg ranges, searches with units)

### Current Behaviour

Validation must be performed out-of-band by the ingestor tools/libraries

### Expected Behaviour

- Configure a json schema for scientificMetadata
- `isValid` endpoints check this schema
- Creating and updating datasets fails if it doesn't match the schema
- Enable enhanced functionality if scientificMetadata matches a schema (eg matching a schema for quantities enables value+unit parsing by the frontend)

### Questions

1. Should users be able to specify (optionally) a more restrictive schema? For instance, a schema for EM data. Should this be another top-level property, or something like `scientificMetadata.@context` (JSON-LD style).
2. What schemas are implicitly supported by SciCat currently (eg with specific frontend behavior)?

sbliven:
Before implementing this we would have to agree on the schema technology and the spellings for the terms. I think that the schema language is not a major barrier to adoption, as there are very good tools for converting both schemas and data between different serializations (eg with [LinkML](https://linkml.io/linkml/generators/index.html)).

## JSON Schema

One option would be to add a json schema to the dataset model. Most people I've discussed this with seem to expect the schema to apply only to the scientificMetadata, not to the whole json.
```json
{
    "pid": "123...",
    "schema": "https://osc-em.github.com/oscem_schema.json",
    "scientificMetadata": {
        [validate this]
    }
}
```
If "schema" is omited we would default to no validation for backwards compatibility.

## JSON-LD

JSON-LD also has good support and is used in many metadata containers (eg RO-Crate). One nice feature is that the context (aka schema) can be either a shared schema included by URI or can be inlined into the data itself. We could adapt the dataset model so that it's valid JSON-LD by including a `@context` key linking a published minimal schema for scicat. This could even be overridden within `scientificMetadata` to specify a more restrictive schema. 

## LinkML

Would it make sense to validate LinkML directly?

sbliven:
[...]

PR #1694, opened 2025-02-10
feat: add scientific metadata validation

despadam:
A first implementation of scientific metadata validation based on #966 and #1308, using `JSON Schema`.

The schema that the scientific metadata should be validated against gets fetched using Angular's HttpClient.

If schema url is omitted then we default to no validation.

To be decided if the url for the schema should be provided inside scientificMetadata or directly in the dto.


sbliven:
I personally favor the second option. Should we include the schema type for future-proofing (eg if we want json-ld support as well?

```
{
    "pid": "123...",
    "schema": {
        "uri": "https://raw.githubusercontent.com/osc-em/OSCEM_Schemas/refs/heads/main/project/spa/jsonschema/oscem_schemas_spa.schema.json"
        "type": "json-schema"
    },
    "scientificMetadata": {
        [validate this]
    }
}
```

emigun:
Perhaps it not obvious to everyone that the schema is for scientificMetadata?

Is `schemaScientificMetadata` or `scientificMetadataSchema` too long?

nitrosx:
Couple of questions:
- this PR implements validation in the sense that you will not create the dataset if the validation is not successful. Am I correct?
- what if I update the dataset later and the metadata do not pass the validation?
- what if I want to create the dataset independently if the validation is successful?

despadam:
> 1. this PR implements validation in the sense that you will not create the dataset if the validation is not successful. Am I correct?

Yes, that is correct

> 2. what if I update the dataset later and the metadata do not pass the validation?

The same should apply, meaning that the update will not be successful in the case of non successful metadata validation. This has not been implemented yet, but it would not require significant code changes, since the validation is in the form of a Pipe.

> 3. what if I want to create the dataset independently if the validation is successful?

Validation will only be performed when a schema is passed. All users still have the ability to create/update datasets as usual and can skip the validation part by omitting the metadata schema.

---
Question related to 2.:
- Currently the validation Pipe is only being used in `POST /api/v3/datasets`. There has been recent activity that updates the Dataset dtos, schema, controller etc to v4. Should validation be applied to both versions? Should we wait until migration is complete?

nitrosx:
Just to make sure that I understand correctly...
The validation is performed only if you provide a schema in the dataset dto when creating with a POST.
Am I correct?

Regarding your question:
I would say let's wait. Let's implement it for dataset v3 first. Once the feature has been merged and dataset v4 is production ready, we can implement it there also.

despadam:
> Just to make sure that I understand correctly... The validation is performed only if you provide a schema in the dataset dto when creating with a POST. Am I correct?

Yes, that is correct

> Regarding your question: I would say let's wait. Let's implement it for dataset v3 first. Once the feature has been merged and dataset v4 is production ready, we can implement it there also.

Ok, sounds good

despadam:
After discussing about https://github.com/SciCatProject/scicat-backend-next/issues/1772, the current implementation is as follows.

Only applied to POST `api/v3/datasets` for now, just for testing purposes.
Adds the field `scientificMetadataSchema` in the relevant DTOs and the `DatasetClass`. The fields is an optional string and should contain the link to the schema that the `scientificMetadata` should be validated against.

No longer using the `FormatPhysicalQuantitiesInterceptor`, which edits the `scientificMetadata` by adding `unitSI/valueSI` wherever `unit/value` exist. We now use `ExtractPhysicalQuantitiesInterceptor`, which is using the same logic as the previous interceptor to automatically calculate the SI fields, but does not directly add them to `scientificMetadata`. Instead these fields will now be stored **separately**, as an additional field **inside** `scientificMetadata` (called `scientificMetadataSI`). `scientificMetadataSI` is a flat object that only contains the SI fields, to avoid duplicating the entire `scientificMetadata` object.

**Example:**
Assuming the given `scientificMetadata` is:
```
(A)

"scientificMetadata": {
    "Person": {
        "Name": "John",
        "Surname": "Doe"
    },
    "Height": {
        "value": 180,
        "unit": "cm"
    }
}
```
then `FormatPhysicalQuantitiesInterceptor` would **directly** edit the field and store it in the db like so:
```
(B)

"scientificMetadata": {
    "Person": {
        "Name": "John",
        "Surname": "Doe"
    },
    "Height": {
        "value": 180,
        "unit": "cm",
        "valueSI": 1.8,
        "unitSI": "m"
    }
}
```
Instead the `ExtractPhysicalQuantitiesInterceptor` will store this:
```
(C)

"scientificMetadata": {
    "Person": {
        "Name": "John",
        "Surname": "Doe"
    },
    "Height": {
        "value": 180,
        "unit": "cm"
    },
    "scientificMetadataSI": {
        "Height.valueSI": 1.8,
        "Height.unitSI": "m"
    }
}
```

Approach (C) allows us to separate the original `scientificMetadata` from the SI fields, while also doing the calculation once and storing this information in the db. So `scientificMetadata` can remain untouched and easily be returned as is: this would only require one edit to omit the `scientificMetadataSI` field before returning the retrieved dataset.

On the other hand, to return the `scientificMetadata` combined with the SI fields, like (B) in the example above, we introduce another interceptor called `ReturnWithPhysicalQuantitiesInterceptor`:

`ReturnWithPhysicalQuantitiesInterceptor` will only run at the very end of an endpoint's logic so that it will edit the result to be returned. By using this interceptor we are able to combine the original `scientificMetadata` with the SI fileds, by traversing the flattened object stored in `scientificMetadataSI`, as seen in (C), to create the object that contains both, like (B).

---
**A note on NestJS Interceptors and Pipes:**
As mentioned also in https://github.com/SciCatProject/scicat-backend-next/issues/966, the correct approach is that validation is performed with a Pipe. However, Interceptors will always run first, before Pipes. This means that the SI fields are still calculated before the validation:
- requires that users provide a schema that already includes the `unitSI/valueSI` fields
- validation should be performed on an appended `scientificMetadata` object like (B)

Generally, Interceptors are the first and last thing to run when calling an endpoint. This means that when a user for example does a POST `/api/v3/datasets`, all Interceptors of that endpoint will be called twice: once at the beginning and once at the end.
Even though called twice, it is possible to separate an interceptors logic and execute something only at the end, like `ReturnWithPhysicalQuantitiesInterceptor` does. To achieve this the Interceptor subscribes to new Observable, waiting for the endpoint's return value to be calculated, at which point it will execute some logic that will edit that value before it actually gets returned.

sbliven:
`ExtractPhysicalQuantitiesInterceptor` is a good proof of concept. To use it we would have to update the search functionality so that searching for quantities redirects them to `scientificMetadataSI`.

There are a couple design decisions I'm not sure about (unless they help implementing search).
- Why flatten the object instead of keeping it nested?
```
"scientificMetadataSI": {
  "Height": {
    "valueSI": 1.8,
    "unitSI": "m"
  }
}
```
- I would use a special character in the name, eg `__scientificMetadataSI`
- Should `scientificMetadataSI` be stored in the dataset? The cleanest solution would be to add a `DatasetSI` mongo document class which stores this info independently (but would have to be added to the search operations).

despadam:
@sbliven 
No specific reason why the object is flattened to be honest, just personal preference, so no problem to change it back to a normal one. Also, agreed to rename to `__scientificMetadataSI`.

The decision to store this in the dataset document was that this way will require less changes in the search functionality. Creating a new mongo collection could also be a good solution, especially if it can be reused for multiple entities that get SI fields and not only datasets. We can discuss the pros and cons during the meeting.

sbliven:
Some notes from today's meeting:
- Making search/filter work with `scientificMetadataSI` will still require quite a lot of work. It also requires a database migration.
- There was no consensus about whether to keep the *SI fields. There was a principled position from a librarian that scientificMetadata shouldn't be modified from the user's input. However, modifying the scientificMetadata even has some benefits, for instance if you want to enforce min/max bounds from a schema. I believe the decision was to shelve this issue for now, although we may return to it later.
- `__scientificMetadataSI` should not be within `scientificMetadata`. It should either be a Dataset field or a separate collection.
- There was a preference for nested structures over the flattened keys (unless there is a strong reason to flatten them when implementing search)

despadam:
### Latest changes / ready for review:
- `ScientificMetadataValidationPipe` used for POST, PUT, PATCH `v4/datasets`
- Optional field `scientificMetadataSchema` in the DTO
- Optional fields `scientificMetadataSchema` and `scientificMetadataValid` in `DatasetClass`
- Validation is **optional**:
  - If schema is not provided, we skip validation
  - If validation not successful, we set `scientificMetadataValid = false`
- `scientificMetadataSchema` can be a valid url pointing directly to the json schema or a w3id that redirects to the schema. In any case it should return a valid JSON object.

src/datasets/schemas/dataset.schema.ts
nitrosx, line 1: Are we storing the SI units structure within the `ScientificMetadata` field?
If so, is that changing the value provided by the users?

I would add an highlevel field named `scientificMetadataSI` or `scientificMetadataInternal` or, as @sbliven suggested `_ScientificMetadata` where we can store SI units and anything else that is system controlled

nitrosx (changes requested):

src/datasets/schemas/dataset.schema.ts
despadam: We had decided to separate the two issues, during our discussion on 23 Apr 2025 regarding scientific metadata validation. See also here: https://github.com/SciCatProject/scicat-backend-next/issues/1772

The current behavior is that the `scientificMetadata` field gets edited - more specifically appended with `unitSI` and `valueSI` fields, by the `FormatPhysicalQuantitiesInterceptor`. So the original values provided by the users do not change, we just also calculate the SI ones. 

The decision was to not edit that interceptor, nor edit where those new fields get stored, within the scope of this PR. That is because the search functionality would have to change significantly.

despadam:
The latest changes make this feature as less disruptive as possible: if someting fails then instead of throwing an error, we only log it and then set `scientificMetadataValid = false`

sofyalaski (approved):
Thanks for this change to still accept the request even if fetching the schema fails, I believe it will simplify the ingestion

HayenNico:
Hi,
question regarding the validation process: How does the Validator instance from jsonschema work exactly? Does it strictly enforce that all top-level keys in scientificMetadata match the schema description, or can scientificMetadata contain fields unspecified in the schema without failing validation?

A few use cases in DAPHNE4NFDI would want to check on ingestion if a dataset's metadata fields include all necessary parts to comply with community standards without having to provide a full schema/standard for the entirety of scientificMetadata. Is that possible with the current version, or is there maybe a configuration option that could be added to switch to this behavior if not?

nitrosx (approved):

despadam:
Hey @HayenNico,

By default, the Validator does not fail if the object being validated contains extra fields that are not defined in the schema. This means that unspecified fields are allowed unless you explicitly tell the schema to reject them.

I believe this should cover the use case you are mentioning, but since it is also dependent on the schema definition, you need to make sure that the schema itself allows additional properties.

PR #1714, opened 2025-02-18
feat: add support for dyanmic redirect url for oidc login and improve session management

omkar-ethz:
## Description

This PR enables support for frontends to specify a successURL and returnURL during OIDC login. This is required to support custom frontends (e.g. scilog at PSI), but also improves security and user experience compared to the existing approach.

## Motivation

We clarify some potentially confusing terminology:
- `successURL`: the URL of the frontend to pass the access-token (and userId etc) to (e.g. `https://discovery.development.psi.ch/auth-callback`, `http://localhost/login`)
- `returnURL`: the url, or `route` that the frontend should redirect the user to after successful login. (e.g. `/datasets` or `/datasets/pid123/datafiles`)

Previously, in case of OIDC login, the successURL was statically set in the config (`OIDC_SUCCESS_URL`), OR if the config was unset, was extracted from the `Referer` header of the requester of /oidc.
The `Referer` approach is inherently unsafe, as blindly trusting this header can result in passing the access-token to an arbitrary origin.
This PR addresses this by allowing multiple successURLs to be defined in config. For each `${CLIENT}`, one is able to define `OIDC_${CLIENT}_SUCCESS_URL`. In turn, the client parameter can be passed as a query param to /oidc. To preserve backwards compatibility, if the client parameter is not passed, it defaults to "scicat" and uses the `OIDC_SUCCESS_URL` value as before.
Moreover, we keep support for the Referer approach for backwards compatibility, but recommend deprecating it and using this new client specific config instead.

Previously, In SciCat frontend, after successful OIDC login, the user was always redirected back to /datasets page (or any statically configured path set in `OIDC_RETURN_URL`. A much better user experience is: if a user clicks on Login from `/datasets/pid123/datafiles` page, she is redirected to the same page on successful login. To support this, we enable passing `returnUrl` as query param in /oidc. This is then saved in the session, and /oidc/callback controller adds it to the query params when redirecting to the client specific SUCCESS_URL (e.g. https://discovery.development.psi.ch/auth-callback?returnUrl=/datasets/pid123/datafiles), so that the frontend can use it to redirect to returnUrl. 
A corresponding change in scicat frontend is made to add the `returnUrl` query param in the login component: https://github.com/SciCatProject/frontend/pull/1753

## Fixes

* Bug fixed (#X)

## Changes:

- Add `client` specific success and return URLs to the config
- Add `client` and `returnURL` as query parameters to /oidc controller
- Update OidcAuthGuard to use express-session in the recommended way (https://docs.nestjs.com/techniques/session), instead of custom Map as before
- Update /oidc/callback controller to use values from the session to support dynamic success/returnUrl

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
 | `ACCESS_GROUPS_STATIC_VALUES` | string | Yes | Comma-separated list of access groups automatically assigned to all users. Example: "scicat, user". | |
 | `ACCESS_GROUPS_OIDCPAYLOAD_ENABLED` | string | Yes | Flag to enable/disable fetching access groups directly from OIDC response. Requires specifying a field via `OIDC_ACCESS_GROUPS_PROPERTY` to extract access groups. | false |
 | `DOI_PREFIX` | string | | The facility DOI prefix, with trailing slash. | |
-| `EXPRESS_SESSION_SECRET` | string | No | Secret used to set up express session. | |
+| `EXPRESS_SESSION_SECRET` | string | No | Secret used to set up express session. Required if using OIDC authentication | |
 | `HTTP_MAX_REDIRECTS` | number | Yes | Max redirects for HTTP requests. | 5 |
 | `HTTP_TIMEOUT` | number | Yes | Timeout for HTTP requests in ms. | 5000 |
 | `JWT_SECRET` | string | | The secret for your JWT token, used for authorization. | |
@@ -140,7 +140,7 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
 | `OIDC_CLIENT_SECRET` | string | Yes | Secret to provide to the OIDC service to obtain the user token. Example: Aa1JIw3kv3mQlGFWhRrE3gOdkH6xreAwro. | |
 | `OIDC_CALLBACK_URL` | string | Yes | SciCat callback URL to redirect to after a successful login. Example: http://myscicat/api/v3/oidc/callback. | |
 | `OIDC_SCOPE` | string | Yes | Space-separated list of info returned by the OIDC service. Example: "openid profile email". | |
-| `OIDC_SUCCESS_URL` | string | Yes | SciCat Frontend auth-callback URL. Required to pass user credentials to SciCat Frontend after OIDC login. Example: https://myscicatfrontend/auth-callback. | |
+| `OIDC_SUCCESS_URL` | string | Yes | SciCat Frontend auth-callback URL. Required to pass user credentials to SciCat Frontend after OIDC login. Example: https://myscicatfrontend/auth-callback. Must be `frontend-base-url/auth-callback` or `frontend-base-url/login` for the official SciCat frontend. | |
 | `OIDC_RETURN_URL` | string | Yes | The path segment within the SciCat Frontend to redirect to, passed as query param in `OIDC_SUCCESS_URL` and handled by frontend. Example: /datasets. | |
 | `OIDC_FRONTEND_CLIENTS` | string | Yes | Comma separated list of additional frontend OIDC clients for this backend. Example: scilog,maxiv. Their success and return URLs can be configured by setting `OIDC_${CLIENT}_SUCCESS_URL` (E.g. `OIDC_SCILOG_SUCCESS_URL`) and `OIDC_${CLIENT}_RETURN_URL` | |
 | `OIDC_ACCESS_GROUPS` | string | Yes | Functionality is still unclear. | |
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
 | `ACCESS_GROUPS_STATIC_VALUES` | string | Yes | Comma-separated list of access groups automatically assigned to all users. Example: "scicat, user". | |
 | `ACCESS_GROUPS_OIDCPAYLOAD_ENABLED` | string | Yes | Flag to enable/disable fetching access groups directly from OIDC response. Requires specifying a field via `OIDC_ACCESS_GROUPS_PROPERTY` to extract access groups. | false |
 | `DOI_PREFIX` | string | | The facility DOI prefix, with trailing slash. | |
-| `EXPRESS_SESSION_SECRET` | string | Yes | Secret used to set up express session. | |
+| `EXPRESS_SESSION_SECRET` | string | No | Secret used to set up express session. | |
 | `HTTP_MAX_REDIRECTS` | number | Yes | Max redirects for HTTP requests. | 5 |
 | `HTTP_TIMEOUT` | number | Yes | Timeout for HTTP requests in ms. | 5000 |
 | `JWT_SECRET` | string | | The secret for your JWT token, used for authorization. | |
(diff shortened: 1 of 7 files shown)


src/auth/auth.controller.ts
```diff
@@ -89,16 +102,25 @@ export class AuthController {
...
-        res.req.headers["referer"] ||
+      this.configService.get<OidcConfig>("oidc")?.clientConfig[
+        res.req.session.client!
+      ].successURL ||
+        res.req.session.successURL || // only for MAXIV. The value is HTTP Referer of /oidc request. Recommended deprecating.
         "",
     );
     url.searchParams.append("access-token", token.access_token as string);
     url.searchParams.append("user-id", token.userId as string);
     url.searchParams.append(
```
minottic: from the guard implementation [here](https://github.com/SciCatProject/frontend/blob/master/src/app/app-routing/auth.guard.ts#L38), isn't the returnUrl param only used when the successUrl is /login? I think in all other cases that is overridden by the guard (say we set the successUrl to /users and returnUrl=/datasets, the callback will return 302 fe/users?returnUrl=/datasets. But the guard will intercept the /users and set the returnUrl to /users, thus not redirecting to the correct returnUrl

src/auth/auth.controller.spec.ts
minottic, line 1: can we add the test checking for no client in the request and returning the default scicat one?

src/config/configuration.ts
minottic, line 1: maybe test configuration at least the clientConfig as it has non trivial logic

src/auth/auth.controller.spec.ts
minottic: I think we also miss the test where we check that no client is provided, we have a referer and the SUCCESS_URL is unset. this should redirect to the referer

src/auth/auth.controller.ts
omkar-ethz: The guard implementation does the following: if the user is logged in, proceed with the route. If not, redirect the user to /login with returnUrl in query parameter. This is only relevant for _local_ login, because the app does not reload. However, this is useless for OIDC login, as the `this.returnUrl` was not passed to the backend, and would be lost after the backend redirects the user back to login (because it is stored in an angular variable, and in case of OIDC login the app restarts from scratch). I have therefore removed this logic from the AuthGuard FE implementation, and also using RouteTrackerService for local login (Updated PR: https://github.com/SciCatProject/frontend/pull/1753). Now, the RouteTrackerService handles this case for both routes protected by AuthGuard and without. 

Example: Open an incognito window and go to http://localhost/user?qp=abc. Because of AuthGuard, You are redirected to the login page (no query params, but RouteTrackerService has saved /user?qp=abc as returnUrl). After logging in (local / oidc), you return to  http://localhost/user?qp=abc.

You are absolutely right, after the backend redirects with the access-token, only /login, AND /auth-callback are able to handle it. Hence, OIDC_SUCCESS_URL must be fe/login or fe/auth-callback.

src/auth/auth.controller.spec.ts
omkar-ethz: Thanks, added these tests in oidc.guard.spec.ts (as adding it at the controller level would need an e2e test [something like instantiate the nest app with some config, call /oidc with client/returnUrl query param, call /oidc/callback and verify redirected url] -- do you think it is worth writing these?)

Also modified the logic to only use referer when no SUCCESS_URL is configured for the client.

src/config/configuration.ts
omkar-ethz: Agree it has quite some logic, added `configuration.spec.ts` to test it.

src/auth/auth.controller.ts
omkar-ethz: Added this note in the README for OIDC_SUCCESS_URL

src/auth/guards/oidc.guard.ts
```diff
@@ -1,23 +1,51 @@
...
+    if (client && typeof client === "string") {
+      if (!oidcConfig?.frontendClients.includes(client)) {
+        throw new HttpException("Unauthorized client", HttpStatus.UNAUTHORIZED);
+      }
     } else {
-      request.headers["referer"] = this.referer[cookie];
-      delete this.referer[cookie];
+      client = "scicat";
+    }
+    request.session.client = client;
```
omkar-ethz: This is probably a bug. When the request is intercepted for /oidc/callback, the query parameter will be absent and hence request.session.client will be overwritten to "scicat"

src/config/configuration.ts
```diff
@@ -74,6 +74,41 @@ const configuration = () => {
...
+            isDefault
+              ? "OIDC_SUCCESS_URL"
+              : `OIDC_${client.toUpperCase()}_SUCCESS_URL`
+          ],
+        returnURL:
+          process.env[
+            isDefault
+              ? "OIDC_RETURN_URL"
+              : `OIDC_${client.toUpperCase()}_RETURN_URL`
+          ],
```
Junjiequan: For a better debug experience I'd suggest to give a warning or error when `OIDC_${client.toUpperCase()}_SUCCESS_URL` or `OIDC_${client.toUpperCase()}_RETURN_URL` is not defined or mismatching.

src/auth/guards/oidc.guard.ts
omkar-ethz: Fixed by treating the /oidc and /oidc/callback cases differently in the guard. No modification to the request is made in /oidc/callback. The session is populated during /oidc, and /oidc/callback simply uses values from the session. As the guard validates client on /oidc, the successURL in session can be considered trusted in /oidc/callback

src/config/configuration.ts
omkar-ethz: Thanks for the suggestion, throwing error on missing `OIDC_${client.toUpperCase()}_SUCCESS_URL` (as if this is missing, `/oidc/callback` will crash due to empty string in new URL(...)) and a logging a warning on missing `OIDC_${client.toUpperCase()}_RETURN_URL` (as there are defaults for this and can be dynamically set and does not cause complete failure in `/oidc/callback`)

README.md
```diff
@@ -140,7 +140,9 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
 | `OIDC_CLIENT_SECRET` | string | Yes | Secret to provide to the OIDC service to obtain the user token. Example: Aa1JIw3kv3mQlGFWhRrE3gOdkH6xreAwro. | |
 | `OIDC_CALLBACK_URL` | string | Yes | SciCat callback URL to redirect to after a successful login. Example: http://myscicat/api/v3/oidc/callback. | |
 | `OIDC_SCOPE` | string | Yes | Space-separated list of info returned by the OIDC service. Example: "openid profile email". | |
-| `OIDC_SUCCESS_URL` | string | Yes | SciCat Frontend auth-callback URL. Required to pass user credentials to SciCat Frontend after OIDC login. Example: https://myscicatfrontend/auth-callback. | |
+| `OIDC_SUCCESS_URL` | string | Yes | SciCat Frontend auth-callback URL. Required to pass user credentials to SciCat Frontend after OIDC login. Example: https://myscicatfrontend/auth-callback. Must be `frontend-base-url/auth-callback` or `frontend-base-url/login` for the official SciCat frontend. | |
```
minottic, line 143: minor comment: maybe this can be part of the config validation? I.e. when setting client = scicat (or empty), and OIDC_SUCCESS_URL must be either /login or /auth-callback

minottic (approved):

src/auth/auth.controller.ts
minottic: minor comment: maybe this can be part of the config validation? I.e. when setting client = scicat (or empty), and OIDC_SUCCESS_URL must be either /login or /auth-callback

Junjiequan (approved):
Looks good to me

README.md
omkar-ethz: Agree it's better to make it explicit -- added this validation in the config (and a corresponding unit test)

PR #1715, opened 2025-02-18
chore: allow unauthenticated users to access dataset origdatablock endpoint

Junjiequan:
## Description
This PR fixes an issue where unauthenticated users could not read public originaldatablocks from the `dataset/:id/originaldatablock` endpoint.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/casl/casl-ability.factory.ts
+++ b/src/casl/casl-ability.factory.ts
@@ -442,74 +442,85 @@ export class CaslAbilityFactory {
     const { can, cannot, build } = new AbilityBuilder(
       createMongoAbility<PossibleAbilities, Conditions>,
     );
-    if (user.currentGroups.some((g) => this.accessGroups?.delete.includes(g))) {
-      /*
+    if (!user) {
+      /**
+      /*  unauthenticated users
+      **/
+      can(Action.OrigdatablockRead, OrigDatablock);
+    } else {
+      if (
+        user.currentGroups.some((g) => this.accessGroups?.delete.includes(g))
+      ) {
+        /*
       / user that belongs to any of the group listed in DELETE_GROUPS
       */
 
-      can(Action.OrigdatablockDelete, OrigDatablock);
-    } else {
-      /*
+        can(Action.OrigdatablockDelete, OrigDatablock);
+      } else {
+        /*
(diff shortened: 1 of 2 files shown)


src/casl/casl-ability.factory.ts
```diff
@@ -125,7 +125,7 @@ export class CaslAbilityFactory {
       cannot(Action.DatasetAttachmentDelete, DatasetClass);
       // -
       cannot(Action.DatasetOrigdatablockCreate, DatasetClass);
-      cannot(Action.DatasetOrigdatablockRead, DatasetClass);
+      can(Action.DatasetOrigdatablockRead, DatasetClass);
```
martin-trajanovski, line 128: I am a bit unsure about this one because all the endpoints that allow unauthenticated users are using the Public keyword. For example if we take a look at any of the `can` blocks above: 

can(Action.DatasetDatablockReadPublic, DatasetClass, {
        isPublished: true,
      });

Not sure if we should use something similar here as well. For example `DatasetOrigdatablockReadPublic`

martin-trajanovski: But as I see we already have that in the code above but not sure why it is not useful. I guess it is because those Origdatablocks are not public and the user is not able to access them. But this way I guess we are allowing unauthenticated users to access all the Origdatablocks if I am not wrong.

src/casl/casl-ability.factory.ts
```diff
@@ -1192,11 +1203,11 @@ export class CaslAbilityFactory {
...
       **/
 
-      can(Action.OrigdatablockReadManyPublic, SampleClass);
-      can(Action.OrigdatablockReadOnePublic, SampleClass, {
+      can(Action.OrigdatablockReadManyPublic, OrigDatablock);
+      can(Action.OrigdatablockReadOnePublic, OrigDatablock, {
         isPublished: true,
       });
-      can(Action.DatasetOrigdatablockReadPublic, SampleClass, {
+      can(Action.DatasetOrigdatablockReadPublic, OrigDatablock, {
```
martin-trajanovski, line 1210: Good catch

martin-trajanovski (approved):

PR #1728, opened 2025-02-25
remove datasetsValidation fromJobClass

sofyalaski:
## Description
Removes `datasetsValidation` from JobClass
## Motivation
Storing unnecessary data in the database

## Changes
Instead of providing this property to Casl in form of a class member, the variable `datasetsNoAccess` is preset to zero. After evaluating `numberOfDatasetsWithAccess`, `datasetsNoAccess` are recalculated (same as before). 
On casl side, only JobCreateConfiguration authorization instance required the `datasetsVallidation` property. To compensate for that, in  the controller a logical and on this casl term and `datasetsNoAccess==0` is evaluated. This way two previous casl instances still can evaluate to true  and JobCreateConfiguration should only be evaluated in combination with validated datasets.
This way we don't need to change DTO either.

## Tests included

- [ ] Passing? 

## Documentation


src/jobs/jobs.controller.ts
```diff
@@ -508,9 +509,8 @@ export class JobsController {
       }
       const numberOfDatasetsWithAccess =
         await this.datasetsService.count(datasetsWhere);
-      const datasetsNoAccess =
-        datasetIds.length - numberOfDatasetsWithAccess.count;
-      jobInstance.datasetsValidation = datasetsNoAccess == 0;
+      datasetsNoAccess = datasetIds.length - numberOfDatasetsWithAccess.count;
+      // jobInstance.datasetsValidation = datasetsNoAccess == 0;
```
despadam: Awesome, thanks @sofyalaski! Let's remove this comment as well and then feel free to merge!

despadam (approved):

Issue #1668, opened 2025-01-28
bug[release-jobs]: Circular Dependency in ValidateJobAction


sbliven:
## Summary

I'm currently struggling with a circular dependency in the validateaction. This issue is half plea for help and half documenting my attempts. I expect the issue to generalize to other JobActions in the future.

## Steps to Reproduce

The issue relates to my [validateaction-circular](https://github.com/sbliven/scicat-backend-next/tree/validateaction-circular) branch. Here's the dependency graph, where errors indicate an `imports` relation between NestJS modules:

```mermaid
flowchart LR

DatasetsModule --> CaslModule
CaslModule --> JobConfigModule
JobConfigModule --> CoreJobActionCreators
CoreJobActionCreators --> ValidateJobActionModule
ValidateJobActionModule -->|forwardRef| DatasetsModule
```

CaslService needs to get authorization info from the JobConfigModule. ValidateJobActionModule needs to access the database to retrieve dataset information.

NestJS provides a doc page on [circular dependencies](https://docs.nestjs.com/fundamentals/circular-dependency#module-forward-reference) which suggests breaking the circle using a `forwardRef`. This somewhat works in the above branch (the app starts up), but numerous tests fail due to remaining circular dependencies. Adding `forwardRef` to every import of CaslModule fixes the circular dependency errors, but gives further errors about the database connections not being set up.

Another suggestion from the docs is to use `ModuleRef` to load the module dynamically. I tried this in the [validateaction](https://github.com/sbliven/scicat-backend-next/tree/validateaction) branch. Tests pass, but the app doesn't start because the DatasetService is not available at runtime in the correct context.

sbliven:
@fpotier @despadam Thanks for your suggestions on this.

sbliven:
This is now fixed in #1473. ModuleRef was the correct path, with the DatasetService getting resolved at runtime rather than during dependency injection since DatasetService has REQUEST scope.

sbliven:
Fixed thanks to @sofyalaski

Issue #1438, opened 2024-10-14
Appropriate default for JOB_CONFIGURATION_FILE


sbliven:
### Summary
SciCat should have an appropriate default job configuration.

### Current Behaviour
The job configuration file is read from the JOB_CONFIGURATION_FILE. It currently defaults to (src/config/configuration.ts:47):

```
  const jobConfigurationFile =
    process.env.JOB_CONFIGURATION_FILE ||
    ("src/jobs/config/jobConfig.example.json" as string);
```

This was convenient for development, but should now be set to a secure final value.

### Expected Behaviour
I would say that the default behavior when JOB_CONFIGURATION_FILE is unset should be to look for `"jobConfig.json"` in the current directory. This mirrors the location for `functionalAccounts.json`. If the file is not found it should be treated as an empty config, eg
```
{
  "configVersion": "",
  "jobs": []
}
```

If #1120 is implemented then Scicat checks for `./jobConfig.{json,yaml,yml}` and uses the first one found.

If JOB_CONFIGURATION_FILE is set then it is loaded. An error should be thrown if the file is not found.

sbliven:
This was resolved by #1539, but with slightly different behavior. If JOB_CONFIGURATION_FILE is omitted then it doesn't configure any jobs. Thus any request to `POST /jobs` will lead to an `unknown job type` error.

If you want to configure the job system, it is required to set `JOB_CONFIGURATION_FILE=jobConfig.yaml` manually. The file extension is ignored, with either JSON or YAML content supported.

Issue #1435, opened 2024-10-03
Jobs: add unit and integration tests for actions


despadam:
As seen also here: https://github.com/SciCatProject/scicat-backend-next/issues/1123

despadam:
mocha test with URL action and archiver mock in scicatlive

despadam:
#1644 has been merged

PR #1741, opened 2025-03-04
feat: job configuration subsystem

sbliven:
## Description
This culminates the jobs release, a major change to how Jobs work in scicat. This allows flexible interaction with other facility infrastructure, including archival and publishing services. Rather than a few hard-coded job types, jobs can now be configured using the `JOB_CONFIGURATION_FILE` env variable, which points to a yaml file allowing customization how scicat handles new and updated jobs.

Full documentation is provide in https://github.com/SciCatProject/documentation/pull/56. Please read that for information about authorization, configuration, and options.

The following actions are included in the initial release, with additional actions planned to follow:
- url
- rabbitmq
- validate
- email
- log

## Fixes

- #1668
- #1439
- #1438
- #1435
- #1424
- #1411
- #1388
- #1341
- #1294
- #1265
- #1154
- #1152
- #1124
- #1123
- #1121
- #1119
- #1114
- #1113
- #841
- #840
- #839
- #838
- #740
- #626
- #625
- #624
- #623
- #622
- #621
- #620
- #1754 

## Tests included

- [x] Passing? 

## Documentation

https://github.com/SciCatProject/documentation/pull/56


src/casl/casl-ability.factory.ts
emigun, line 1: I think this file is kind of growing out of control. It would be nice to refactor it so that at least each endpoint has its own file.

Junjiequan: I agree. I've found the code hard to read, and I've also heard complaints from others. 
I will find a time to refactor this.

emigun: Lovely!

src/config/job-config/actions/corejobactioncreators.module.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+@Module({
+  imports: [
+    EmailJobActionModule,
+    LogJobActionModule,
+    ValidateJobActionModule,
+    URLJobActionModule,
+    RabbitMQJobActionModule,
+  ],
+  providers: [
```
Junjiequan, line 39: ~~I can see that each module in the array will be initialized two times. One for `CREAT_JOB_ACTION` and the other for `UPDATE_JOB_ACTION`.~~  <-- Correcting my initial comment - NestJS modules are singletons and their providers are only initialized once.
If I want to use only RabbitMQJobAction and do not need the others, how can I disable them from initializing to reduce unnecessary initialization overhead

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -0,0 +1,120 @@
...
+    const hostname = this.configService.get<string>("rabbitMq.hostname");
+    const port = this.configService.get<number>("rabbitMq.port");
+    const username = this.configService.get<string>("rabbitMq.username");
+    const password = this.configService.get<string>("rabbitMq.password");
+
+    if (!hostname || !port || !username || !password) {
+      Logger.error(
+        "RabbitMQ is enabled but missing one or more config variables.",
+        "RabbitMQService",
+      );
```
Junjiequan, line 31: `onApplicationBootstrap()` under the [app.module.ts ](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/app.module.ts) tries to connect to the rabbitMQ using the same variables, so there will be two instances of rabbitMQ connections once enabled.
That's some legacy code from the ESS that is no longer used but not removed. Would you prefer me to create new PR to remove that or just delete them from this PR?

src/common/rabbitmq/rabbitmq.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from "@nestjs/common";
+import { ConfigModule } from "@nestjs/config";
+import { RabbitMQService } from "./rabbitmq.service";
+
+@Module({
+  imports: [ConfigModule],
+  providers: [RabbitMQService],
+  exports: [RabbitMQService],
+})
+export class RabbitMQModule {}
```
Junjiequan, line 10: RabbitMQModule is created but is not imported anywhere. Is it intentional?

src/common/common.module.ts
```diff
@@ -1,8 +1,24 @@
...
+        const isEnabled =
+          configService.get<string>("rabbitMq.enabled") === "yes";
+        if (isEnabled) {
+          return new RabbitMQService(configService);
+        }
+        return null;
+      },
+      inject: [ConfigService],
+    },
+  ],
```
Junjiequan: Wouldn't it be better to import `RabbitMQmodule` here instead of setting `RabbitMQService` as a provider? Considering rabbitmq module is created but not used anywhere

src/config/job-config/README.md
```diff
@@ -0,0 +1,51 @@
...
+instance. Job types are arbitrary and facility-specific, but `archive` and `retrieve`
+are traditional for interacting with the data storage system. Authorization can be
+configured for each job type for *create* and *update* requests, and then a list of
+actions are provided which should run after the request.
+
+## Implementing an Action
+
+Implementing an Action requires four (short) files:
+
+1. `action.ts` contains a class implementing `JobAction`. It's constructor can take any
```
emigun:
```suggestion
1. `action.ts` contains a class implementing `JobAction`. Its constructor can take any
```

emigun: Or perhaps `The constructor...`

src/config/job-config/actions/urlaction/urlaction.ts
```diff
@@ -0,0 +1,83 @@
...
+      throw new HttpException(
+        {
+          status: response.status,
+          message: `Got response: ${await response.text()}`,
+        },
+        response.status,
+      );
+    }
+
+    // TODO do something with the response?
```
emigun: Forgotten?

src/jobs/dto/create-job.dto.ts
```diff
@@ -1,30 +1,52 @@
...
+    description: "Valid job type as defined in configuration.",
+  })
   @IsString()
   readonly type: string;
 
+  @ApiProperty({
+    type: Object,
+    required: true,
+    description: "Job's parameters as defined by job template in configuration",
+  })
```
emigun: Did you consider using the openapi cli plugin instead for ApiProperty? I think it is more maintainable.

src/jobs/dto/update-job.dto.ts
```diff
@@ -1,18 +1,30 @@
...
 
+@ApiTags("jobs")
 export class UpdateJobDto {
-  @IsDateString()
-  @IsOptional()
-  readonly executionTime?: Date;
+  @ApiProperty({
+    type: String,
+    required: true,
+    description: "Updated job status code for the current status.",
```
emigun: Would be nice with the automatic swagger from openapi cli plugin here too.

src/jobs/schemas/job.schema.ts
```diff
@@ -2,100 +2,134 @@ import { Prop, Schema, SchemaFactory } from "@nestjs/mongoose";
...
   },
 })
-export class JobClass {
+export class JobClass extends OwnableClass {
   @ApiProperty({
     type: String,
+    default: () => uuidv4(),
     description: "Globally unique identifier of a job.",
-    readOnly: true,
   })
```
emigun: Same for this file (openapi cli plugin)

test/DatasetFilter.js
```diff
@@ -120,7 +120,7 @@ describe("0400: DatasetFilter: Test retrieving datasets using filtering capabili
     accessTokenArchiveManager = await utils.getToken(appUrl, {
       username: "archiveManager",
       password: TestData.Accounts["archiveManager"]["password"],
-    });
+    });
```
emigun: Just a whitespace here

test/DatasetLifecycle.js
```diff
@@ -29,7 +29,7 @@ describe("0500: DatasetLifecycle: Test facet and filter queries", () => {
     });
   });
 
-
+
```
emigun: And whitespace here. At least that's what I think it means :)

test/DerivedDataset.js
```diff
@@ -21,6 +21,7 @@ describe("0700: DerivedDataset: Derived Datasets", () => {
     accessTokenAdminIngestor = await utils.getToken(appUrl, {
       username: "adminIngestor",
       password: TestData.Accounts["adminIngestor"]["password"],
+
```
emigun: Remove this line?

emigun:
Wow! :star_struck: Honestly, this looks amazing! :star: 

Well done!! :1st_place_medal: :1st_place_medal: :1st_place_medal:

src/common/rabbitmq/rabbitmq.module.ts
nitrosx: Same question as @Junjiequan

src/common/rabbitmq/rabbitmq.service.ts
nitrosx: @Junjiequan please create a PR to this PR

src/common/utils.ts
```diff
@@ -819,6 +819,24 @@ export const addApiVersionField = <T extends object>(
...
+  jobResultObject: Record<string, unknown> | undefined;
+} => {
+  return {
+    ...obj,
+    statusCode: statusCode,
+    statusMessage: statusMessage,
+    jobResultObject: jobResultObject,
+  };
+};
+
```
nitrosx: I wonder if this is the right place for this function given that it is specific for the job subsystem?
Not a blocking issue

src/config/job-config/actions/corejobactioncreators.module.ts
nitrosx: Valid point!!!

src/jobs/dto/create-job.dto.ts
nitrosx: I agree with @emigun , although it can be done in a new PR

src/users/users.service.ts
```diff
@@ -230,6 +236,22 @@ export class UsersService implements OnModuleInit {
...
+      return {
+        _id: userProfile.id,
+        username: userProfile.username,
+        email: userProfile.email,
+        currentGroups: userProfile.accessGroups,
+      } as JWTUser;
+    }
+    return null;
+  }
+
```
nitrosx, line 254: This function is already present in master.

nitrosx (commented):
Just few minor comments.
Good work.

src/users/users.service.ts
sofyalaski: Indeed, there's a very similar function that finds a user by ID.
I introduced this new function at a time because of the use case where a user in admin groups creates a job for another user. This other user is specified in the newly introduced optional ownerUser parameter in the job DTO.  It would expect an input as a username as of now. 
We can reuse the `findById2JWTUser` and then job DTO needs to provide an actual ID of the user for whom the job will be created. Should I create a PR for this? 

Now when I think about this use case, I think there is another check that is missing. If both ownerUser and ownerGroup are provided in DTO, we don’t check upon job creation, if the ownerUser is part of that ownerGroup. I will create a PR with that check. 

In most cases, ownerUser is optional and is set during job creation based on logged in user.

sofyalaski: @nitrosx, Despina and I decided to keep the `findByUsername2JWTUser` function. As I explained, it is used to find a user provided in `ownerUser` in DTO. We thought that a userID could be used for that field, but concluded that it would be confusing since the created job would also contain just a userID. In a real case scenario, if the admin would want to look at the created job and see some userID in the ownerUser field, it will require a call to User backend to understand who this job was created for. 
And username is probably a unique identifier too.

src/config/job-config/actions/corejobactioncreators.module.ts
sbliven: You could disable them by providing different arrays to nestjs for the `CREATE_JOB_ACTION_CREATORS` and `UPDATE_JOB_ACTION_CREATORS` tokens. Currently `app.module.ts` imports `CoreJobActionCreators` statically. However, if we implement a plugin mechanism in the future (#862) then I would expect RabbitMQJobAction to be moved from core to a plugin, and `AppModule` would use some kind of dynamic import mechanism. It was decided this was too complicated for now, so all actions are included in the core.

src/common/rabbitmq/rabbitmq.module.ts
sbliven: It's imported by `CoreJobActionCreators`

src/common/common.module.ts
sbliven: CommonModule is a bit of a weird module. It is needed for services which are configured globally in app.module but which need to be imported by other modules.

That said, I don't really get why RabbitMQService needs to be created here rather than in its own module. @despadam do you remember the reason?

src/common/rabbitmq/rabbitmq.module.ts
Junjiequan: Searching through the source code it seems that [CoreJobActionCreators](https://github.com/SciCatProject/scicat-backend-next/blob/954516bc707eb2f941de22118e921b3f3487b93a/src/config/job-config/actions/corejobactioncreators.module.ts) imports `RabbitMQJobActionModule`
and [RabbitMQJobActionModule](https://github.com/SciCatProject/scicat-backend-next/blob/954516bc707eb2f941de22118e921b3f3487b93a/src/config/job-config/actions/rabbitmqaction/rabbitmqaction.module.ts) imports `CommonModule`. 
My guess is, create job action for RabbitMQ works since commonModule imports RabbitMQ service as a provider.

src/config/job-config/actions/corejobactioncreators.module.ts
Junjiequan: If it takes much more time to implement then it's better to skip it for now and add it in the future.
NestJS recently introduced new concept of [Conditional module configuration](https://docs.nestjs.com/techniques/configuration#conditional-module-configuration) which might be useful for future reference

src/jobs/dto/create-job.dto.ts
sbliven: I agree, this should be done to bring release-jobs into alignment with @emigun's changes in main.

test/DatasetFilter.js
sbliven: We should add js files to eslint. That will lead to a huge diff though, so let's do it after merging

src/common/utils.ts
sbliven: IMO it fits well with `addCreatedByFields` and `addUpdatedByField` in the same file, but I see your point.

sbliven:
TODO before merging
- [x] Resolve schema concerns from #1742 and merge it to this branch
- [x] Replace `@ApiProperty` annotations with the cli-plugin following #1590 and #1611 (PR: #1770)
- [x] Try [Conditional module configuration](https://docs.nestjs.com/techniques/configuration#conditional-module-configuration) for the RabbitMqModule (PR: #1771)
- [x] Minor lint and comment changes
- [x] Merge in all the bump PRs committed despite the code freeze

despadam:
PRs:
- Replace `ApiProperty` annotations with the openapi cli-plugin: https://github.com/SciCatProject/scicat-backend-next/pull/1770
- Conditional RabbitMQModule: https://github.com/SciCatProject/scicat-backend-next/pull/1771

nitrosx (approved):

PR #1742, opened 2025-03-05
feat(jobs): Ensure backwards compatibility

despadam:
## Description

With this PR we make sure that `api/v4/Jobs/` is backwards compatible with `v3`.

## Motivation

These changes allow us to deploy the new Jobs subsystem without major changes in the FE, while also accommodating the facilities that run `v3`.

## Fixes

Alongside the following changes, all Jobs tests have been updated to use `api/v4/Jobs/`.

## Changes:

NestJS annotations allow us to indicate which methods are compatible with which version(s).
- `GET/fullfacet` and `DELETE` are now compatible with both versions - no further changes were required.
- For `POST`, `PATCH`, `GET`, `GET/:id` and `GET/fullquery` we need two methods for each, one for `v3` and one for `v4`. For that, the following changes were introduced:
  - Brought back the `JobClassV3` schema
  - In the controller, added the method `mapJobClassV4toV3()` that transforms a `v4` Job to a `JobClassV3` instance
  - `GET`, `GET/:id` and `GET/fullquery` for `v3` are now using the `mapJobClassV4toV3()` method so that `JobClassV3` instances are returned and FE compatibility is ensured
  - Added `CreateJobV3MappingInterceptor` which is used before a Job creation with `v3`, to transform the dto to `v4`
  - Added `UpdateJobV3MappingInterceptor` which is used before a Job update with `v3`, to transform the dto to `v4`

## Tests included

- [x] Passing? 

## Documentation

https://github.com/sbliven/scicat-documentation/pull/3


src/jobs/jobs.controller.ts
```diff
@@ -766,6 +935,7 @@ export class JobsController {
     ability.can(Action.JobRead, JobClass),
   )
   @Get("/fullfacet")
+  @Version(["3", "4"])
```
sbliven, line 993: Should we use `VERSION_NEUTRAL` instead of listing the versions explicitly?

src/jobs/jobs.controller.ts
```diff
@@ -567,13 +570,64 @@ export class JobsController {
...
+   */
+  private mapJobClassV4toV3(job: JobClass): JobClassV3 {
+    const jobV3 = new JobClassV3();
+    // Map fields from v4 to v3
+    jobV3._id = job._id;
+    jobV3.id = job.id;
+    jobV3.emailJobInitiator = job.createdBy;
+    jobV3.type = job.type;
+    jobV3.creationTime = job.createdAt;
+    jobV3.jobStatusMessage = job.statusMessage;
```
sbliven: I wonder if `jobV3.jobStatusMessage = job.statusCode` would better reflect how they were actually used. I think v3 statusMessage fields were single words (eg `jobStarted`).

sbliven (approved):
This looks great. Using interceptors for the conversion is a nice design.

src/jobs/jobs.controller.ts
despadam: Let me test that just to be sure

src/jobs/jobs.controller.ts
despadam: Sure, I will use `statusCode` instead

despadam:
Thanks for the review @sbliven - I will also update the documentation accordingly

src/jobs/jobs.controller.ts
sbliven: If it works we should make it the default for all controllers IMO

src/jobs/jobs.controller.ts
```diff
@@ -588,53 +642,50 @@ export class JobsController {
     type: JobClass,
     description: "Created job",
   })
-  async create(
+  async createV3(
     @Req() request: Request,
     @Body() createJobDto: CreateJobDto,
```
sbliven, line 701: Testing this more I noticed a bug. The v3 endpoint is using the v4 DTO (datasets in the jobParams).

sbliven (changes requested):

src/jobs/jobs.controller.ts
```diff
@@ -567,13 +570,64 @@ export class JobsController {
...
+   * Transform a v4 job instance so that is compatible with v3
+   * @param job: a JobClass instance (v4)
+   * @returns a JobClassV3 instance
+   */
+  private mapJobClassV4toV3(job: JobClass): JobClassV3 {
+    const jobV3 = new JobClassV3();
+    // Map fields from v4 to v3
+    jobV3._id = job._id;
+    jobV3.id = job.id;
+    jobV3.emailJobInitiator = job.createdBy;
```
sbliven: Should be contactEmail

src/jobs/jobs.controller.ts
nitrosx: I agree

despadam: I believe that the way we handle versioning falls into a special category seen in the screenshot below from https://docs.nestjs.com/techniques/versioning#multiple-versions

This makes the endpoints with unspecified or neutral version fail, so for now I will keep using `@Version(["3", "4"])` for `/fullfacet` and `/delete` as I am unable to find a workaround 

image

src/jobs/jobs.controller.ts
```diff
@@ -567,13 +570,64 @@ export class JobsController {
...
+    jobV3.type = job.type;
+    jobV3.creationTime = job.createdAt;
+    jobV3.jobStatusMessage = job.statusMessage;
+    jobV3.jobResultObject = job.jobResultObject;
+    jobV3.ownerGroup = job.ownerGroup;
+    // Remove datasetList from jobParams and assign it to jobV3.datasetList
+    const { datasetList, ...jobParams } = job.jobParams;
+    jobV3.datasetList = datasetList as DatasetListDto[];
+    // Store v4 configVersion in v3 jobParams
+    const newJobParams = { ...jobParams, configVersion: job.configVersion };
```
sbliven: v3 didn't have a config version. I suggest we omit it rather than squashing it into jobParams

src/jobs/jobs.controller.ts
sbliven: Ah, that's definitely not what we wanted. Maybe instead
```
app.enableVersioning({
  defaultVersion: ['3', '4']
});
```

The downside is that the swagger explorer will list all endpoints twice.

despadam: I also came across this but was thinking that perhaps it should not be included in this PR as it will affect all endpoints.

sbliven: A separate PR is fine. I just thought about here since you wouldn't need this line if it were already a default.

sbliven:
POST /api/v3/jobs returns a v4 DTO

despadam:
> POST /api/v3/jobs returns a v4 DTO

Good catch @sbliven , I updated the return types for v3

sbliven:
Sorry to keep dribbling in change requests, but I was looking over the [real v3 schema](https://dacat.psi.ch/explorer/#!/Job/Job_findById) and there are still some differences.

| JobClass (v4)                          | dacat (v3)                              | JobClassV3                         | CreateJobDtoV3                      | UpdateJobDtoV3                      |
|------------------------------------|-------------------------------------|------------------------------------|-------------------------------------|-------------------------------------|
|                                    |                                     | _id (string)                       |                                     |                                     |
| id (string)                        | id (string)                         | id (string)                        |                                     |                                     |
| contactEmail (string, optional)    | emailJobInitiator (string)          | emailJobInitiator (string)         | emailJobInitiator (string)          |                                     |
| type (string)                      | type (string)                       | type (string)                      | type (string)                       |                                     |
|                                    | creationTime (string, optional)     | creationTime (string($date-time))  |                                     |                                     |
|                                    | executionTime (string, optional)    | executionTime (string($date-time)) | executionTime (string($date-time))  | executionTime (string($date-time))  |
| jobParams (object, optional)       | jobParams (object, optional)        | jobParams (object)                 | jobParams (object, optional)        |                                     |
| statusCode (string, optional)      | jobStatusMessage (string, optional) | jobStatusMessage (string)          | jobStatusMessage (string, optional) | jobStatusMessage (string, optional) |
|                                    | datasetList (object, optional)      | datasetList (array, optional)      | datasetList (array)                 |                                     |
| jobResultObject (object, optional) | jobResultObject (object, optional)  | jobResultObject (object)           | jobResultObject (object, optional)  | jobResultObject (object, optional)  |
| createdBy (string)                 | createdBy (string, optional)        |                                    |                                     |                                     |
| updatedBy (string)                 | updatedBy (string, optional)        |                                    |                                     |                                     |
| createdAt (string($date-time))     | createdAt (string, optional)        |                                    |                                     |                                     |
| updatedAt (string($date-time))     | updatedAt (string, optional)        |                                    |                                     |                                     |
| ownerGroup (string)                |                                     | ownerGroup (string)                |                                     |                                     |
| accessGroups (array)               |                                     |                                    |                                     |                                     |
| ownerUser (string)                 |                                     |                                    |                                     |                                     |
| instrumentGroup (string, optional) |                                     |                                    |                                     |                                     |
| isPublished (boolean)              |                                     |                                    |                                     |                                     |
| statusMessage (string, optional)   |                                     |                                    |                                     |                                     |
| configVersion (object)             |                                     |                                    |                                     |                                     |



- The optional->Required changes in the schema are fine. 
- The `ownerGroup` addition to JobClassV3 is a significant change. Is this needed for authorization?
- createdAt seem redundant to creationTime, but they do differ slightly in dacat jobs
- executionTime doesn't seem to be stored or set.

src/jobs/interceptors/update-job-v3-mapping.interceptor.ts
```diff
@@ -13,14 +15,33 @@ export class UpdateJobV3MappingInterceptor implements NestInterceptor {
   intercept(context: ExecutionContext, next: CallHandler): Observable<unknown> {
     const request = context.switchToHttp().getRequest();
     const dtoV3 = request.body as UpdateJobDtoV3;
+
+    if (!dtoV3.jobStatusMessage) {
+      throw new HttpException(
```
sbliven: I'm concerned about throwing an exception in the interceptor. I think we should change the DtoV3 properties to be `required: false` (even if this isn't strictly backwards compatible) rather than saying that some jobs can only be fetched over the v4 api.

sbliven: (also `emailJobInitiator` above)

nitrosx: I am also concern about testing for a specific version

despadam: - The concern makes sense regarding `create-job-v3-mapping.interceptor.ts`, I can remove the check for `emailJobInitiator` and keep the implementation as close as possible to `contactEmail`'s in v4.
- Regarding `update-job-v3-mapping.interceptor.ts` the `jobStatusMessage` imitates v4, because there cannot be an update if `statusCode` is not provived.
- I don't see how there are now jobs that cannot be fetched over both APIs

despadam: I have removed the exceptions from the interceptors, updated the v3 dtos regarding optional/required fields and I am making the tests more detailed.

src/jobs/schemas/job.v3.schema.ts
nitrosx, line 1: Why do we need schema v3?
Shouldn't we save every job with schema v4?

src/jobs/jobs.service.ts
nitrosx, line 1: Why do we need updated V3?

src/common/utils.ts
```diff
@@ -837,6 +837,18 @@ export const addStatusFields = <T>(
...
+  jobParams: Record<string, unknown> | undefined,
+): T & {
+  jobParams: Record<string, unknown> | undefined;
+} => {
+  return {
+    ...obj,
+    jobParams: jobParams,
+  };
+};
+
```
nitrosx: I wonder if this function should be in the job controller or service as it is specific for the jobs?

nitrosx: I wonder if this function should be in the job controller or service as it is specific for the jobs?

nitrosx (changes requested):

src/jobs/schemas/job.v3.schema.ts
despadam: All jobs are saved in the db with v4 schema. We brought back the v3 schema and dtos to enforce typing and ensure the mapping between the different versions is correct

src/jobs/jobs.service.ts
despadam: To preserve the `executionTime` field from v3, we store it in `jobParams` in v4. This function makes sure that this field will be updated correctly by using `addUpdatedJobParamsField()` that was introduced in `utils.ts`. I could split this logic differently so that most of this happens in the controller and `jobs.service` and `utils` have as less changes possible. Thanks for the comments

src/jobs/schemas/job.v3.schema.ts
nitrosx: Apologies for being a pain.... 
Why do you need the schema v3 to validate? Wouldn't you use the dto's?

Having schema v3 in code will be confusing!!!! More so, if it is not used!!!

src/jobs/jobs.service.ts
nitrosx: I would leave this function in `utils` if it is or could be used somewhere else.
If the function is specific to jobs, I would move the it within the job subsystem.
I trust your judgement

src/jobs/schemas/job.v3.schema.ts
despadam: No worries, I will see if I can make it work just by using the dtos

src/jobs/jobs.service.ts
despadam: It is specific to jobs, so I will move it :)

despadam:
To map the v3/v4 create/update DTOs `CreateJobV3MappingInterceptor` and `UpdateJobV3MappingInterceptor` have been introduced.

### `CreateJobV3MappingInterceptor`:

- Move `datasetList` to `jobParams`
- Move `executionTime` to `jobParams`
- If no `emailJobInitiator` is provided then assign `request.user.email` to `contactEmail` to accommodate for **anonymous** jobs
- Since v4 is using `ownerUser` and `ownerGroup` for authorization, utilize `username` that is included in the `jobParams` coming from the FE: extract `ownerUser` and `ownerGroup` of user with `findByUsername2JWTUser()`

#### Issues depending on case - for logged in users:

1. Anonymous jobs can be created only by admins:

    - Can be created if no `emailJobInitiator` is provided. Then admin will automatically be assigned to `contactEmail`. `ownerUser` and `ownerGroup` will not exist.
    - Can be created if `emailJobInitiator` is provided but `username` was not provided in `jobParams`. Then `emailJobInitiator` will automatically be assigned to `contactEmail`. `ownerUser` and `ownerGroup` will not exist. The user whose email was provided in `emailJobInitiator` will **not** have access to the job.

3. To differentiate anonymous from user-owned jobs, we take advantage of the `username` that is included in `jobParams` from the FE. Then normal users can only create jobs for themselves:

    - If no `username` is provided in `jobParams` then normal users cannnot create jobs for themselves: `ownerGroup` is required and the controller checks that the user belongs in the specified `ownerGroup`
    - Users can belong in multiple groups but the v4 dto only asks for one group: it will always be impossible to find the correct group required for authorization when using `findByUsername2JWTUser()`
    - `JobClass` in v4 inherits from `Ownable` which has `ownerGroup` as a mandatory field: makes sense for dataset authorization but why is this used for jobs?

### `UpdateJobV3MappingInterceptor`:

- Both `statusCode` and `statusMessage` get the value of `jobStatusMessage`
- To be able to match the v4 dto, if `executionTime` is provided then store it temporarily in `jobResultObject`. When performing the update, remove it from there and store it in `jobParams`.
- Removed `addStatusFields()` from `utils.ts` (it was only used by jobs) and replaced it with helper functions `addStatusFields()` and `updateJobFields()` inside `jobs.service`.  `updateJobFields()` makes sure to also update `jobResultObject` and `jobParams` accordingly in the case that `executionTime` was provided for v3

src/jobs/schemas/job.v3.schema.ts
sbliven: This is a good point @nitrosx. The `JobClassV3` is needed as a return type for the `createV3` and `updateV3` endpoints. However, I think you're right that it shouldn't have the mongoose `@Prop` annotations. It's never stored in the database, so calling it a DTO instead of a schema makes more sense.

sbliven:
Thanks for the summary, @despadam.
1. I'm worried that some facilities might want anonymous users to submit 'public' jobs for moving data. In v3 this was possible by adding `emailJobInitiator` without logging in. In v4 both  `ownerUser` and `ownerGroup` are required in the schema; do you set these to the empty string for anonymous cases?
2. Does it make sense for `JobClass` to extend `OwnableClass`? I'm tempted to suggest extending `QueryableClass` instead, which would eliminate the `ownerGroup`, `accessGroups`, `instrumentGroup`, and `isPublished` fields. This could be a fairly big change request (eg in casl-ability). However, I'm not sure that the 'ownable' model makes sense for jobs, which are more user-focused. Plus, removing those 4 properties would make the schemas much more comparable:

| JobClass (v4)                      | dacat (v3)                          |
| ---------------------------------- | ----------------------------------- |
| id (string)                        | **id (string)**                     |
| contactEmail (string, optional)    | **emailJobInitiator (string)**      |
| type (string)                      | **type (string)**                   |
| *(createdAt)*                      | creationTime (string, optional)     |
| jobParams.executionTime (optional)           | executionTime (string, optional)    |
| jobParams (object, optional)       | jobParams (object, optional)        |
| statusCode (string, optional)      | jobStatusMessage (string, optional) |
| statusMessage (string, optional)   |                                     |
| jobParams.datasetList (optional)      | datasetList (object, optional)      |
| jobResultObject (object, optional) | jobResultObject (object, optional)  |
| createdBy (string)                 | createdBy (string, optional)        |
| updatedBy (string)                 | updatedBy (string, optional)        |
| createdAt (string($date-time))     | createdAt (string, optional)        |
| updatedAt (string($date-time))     | updatedAt (string, optional)        |
| configVersion (string, filled by scicat)   |                                     |
| ownerUser (string)                 |                                     |
| ownerGroup (string)                |                                     |
| accessGroups (array)               |                                     |
| instrumentGroup (string, optional) |                                     |
| isPublished (boolean)              |                                     |

src/jobs/schemas/job.v3.schema.ts
despadam: The latest commit (https://github.com/SciCatProject/scicat-backend-next/pull/1742/commits/c2b375bc57847b95e1b3243745a9c16e8e6ac24a) fixes `JobClassV3`: removed `@Schema` and `@Prop` annotations for mongoose and we are now only using `@ApiProperty` for swagger.
- Tried to use just `CreateJobDtoV3` extended with `id`, but was not possible because `CreateJobDtoV3` has readonly values but we need to be able to update the fields of the instance.
- Tried using the `openapi cli-plugin` instead of `ApiProperty` but was not able to make it work in this case. Perhaps because this class does not correspond to a db schema? Not sure, so comments and suggestions are welcome

despadam:
@sbliven 
1. is now covered: Anonymous users can create anonymous jobs with the possibility to add an `emailJobInitiator`. Then these jobs can only be accessed by admins. The user whose email was provided as `emailJobInitiator` cannot access the job. In general `ownerUser` and `ownerGroup` will not exist at all for anonymous cases, something that actually puzzles me since they are required in the schema.

despadam:
@nitrosx @sbliven @sofyalaski 
Based on our discussion yesterday:
- renamed `JobClassV3` to `OutputJobV3Dto`
- added an empty string as `ownerGroup`, unless `username (=ownerUser)` is provided, in which case there is a check in v4 that the user should be part of the provided `ownerGroup`:
  - to make this work the solution is to keep using `findByUsername2JWTUser` and pick the last of the groups on the list. This means that when a user wants to create a job for himself in v3, the job will be automatically shared with his group (or one of them in case he belongs in multiple groups)
- Created issue #1775 to de discussed later

nitrosx (approved):
Awesome work

sofyalaski (approved):
looks good, thank you!

sbliven:
> to make this work the solution is to keep using findByUsername2JWTUser and pick the last of the groups on the list. This means that when a user wants to create a job for himself in v3, the job will be automatically shared with his group (or one of them in case he belongs in multiple groups)

Can you explain why it doesn't work to leave `userGroup=""` in this case? I find it unexpected that jobs get automatically shared with an arbitrary group.

src/jobs/jobs.controller.ts
```diff
@@ -362,20 +362,45 @@ export class JobsController {
...
-          );
+        let jobUser: JWTUser | null = null;
+        if (jobCreateDto.ownerUser) {
+          if (user.username != jobCreateDto.ownerUser) {
+            jobUser = await this.usersService.findByUsername2JWTUser(
+              jobCreateDto.ownerUser,
+            );
+            if (jobUser === null) {
+              Logger.log(
+                "Owner user was not found, using current user instead.",
```
sbliven: Should we throw an httpexception instead of just logging it?

sofyalaski:
@sbliven it's because in the current controller, it is expected that the ownerGroup field includes a group and a user creating the request is part of this group for non-admin users: 
https://github.com/SciCatProject/scicat-backend-next/blob/f4a41de34480b50463460d40b5b5188aea20b43b/src/jobs/jobs.controller.ts#L399-L421

src/jobs/dto/create-job.v3.dto.ts
```diff
@@ -12,7 +12,8 @@ import { UpdateJobDtoV3 } from "./update-job.v3.dto";
 
 export class CreateJobDtoV3 extends UpdateJobDtoV3 {
```
sbliven, line 13: Fields lack descriptions

src/jobs/dto/output-job-v3.dto.ts
```diff
@@ -0,0 +1,76 @@
...
+    description: "Globally unique identifier of a job.",
+  })
+  id: string;
+
+  @ApiProperty({
+    type: String,
+    required: false,
+    description: "The email of the person initiating the job request.",
+  })
+  emailJobInitiator?: string;
```
sbliven, line 20: It looks like `@Optional()` is needed in order for 'optional' to show up in the swagger docs (although the '*' is correctly omitted based just on the typescript `?` type). I'm not sure if this matters.

despadam:
Thanks @sofyalaski,
Yes exactly, cannot have an ownerUser without an ownerGroup, because of this check. A group is required and the user should belong to that group

despadam:
The other solution, instead of adding arbitrary group, is assigning an empty string also to ownerUser but then that user will not have access to the job he just created

src/jobs/dto/update-job.v3.dto.ts
```diff
@@ -7,8 +7,7 @@ export class UpdateJobDtoV3 {
   readonly executionTime?: Date;
 
   @IsString()
```
sbliven, line 15: Also lack descriptions

sbliven (approved):
I am satisfied with all the DTOs now, which was the most important issue to decide here.

src/jobs/jobs.controller.ts
despadam: Let's address this one later. See relevant issue https://github.com/SciCatProject/scicat-backend-next/issues/1776

despadam:
Created issue https://github.com/SciCatProject/scicat-backend-next/issues/1776 to address the remaining topics after merging this PR, specifically how we want to handle authorization with `ownerUser`/`ownerGroup`

sbliven:
Looks great. This was unexpectedly arduous. Thanks for the hard work, @despadam!

PR #1743, opened 2025-03-05
feat: metadata quantity_range valueSI conversion support and test

martin-trajanovski:
## Description
After adding new scientific metadata types like number_range and quantity_range on the frontend this is just the improvement of the valueSI and unitSI conversion for the quantity types. Also added test case where it checks if the conversion is done properly.

## Motivation
Addition of the new scientific metadata types

## Fixes
https://jira.ess.eu/browse/SWAP-4564

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [x] Passing? 

## Documentation


test/TestData.js
```diff
@@ -176,6 +176,11 @@ const TestData = {
         value: 8500,
         unit: "",
       },
+      approx_distance_range: {
+        value: [1, 2],
+        unit: "cm",
+        type: "quantity_range",
```
sbliven, line 182: Does the type get used by mathjs anywhere? It looks like this code would apply to all arrays.

sbliven (commented):
Applying a single unit to the whole array is a bit limiting. Maybe it would be better to encode quantity_range as a tuple of quantities?

```yaml
approx_distance_range:
  - value: 1
    unit: cm
  - value: 2
    unit: cm
```

This would fit the current code without modifications.

test/TestData.js
martin-trajanovski: Your comment seems valid one and here is what I think:
The type doesn't get used at the moment but the condition is set there from before. The unit conversion is called if:
`value !== undefined && unit && unit.length > 0`
So in this case if there is an array value and we have provided a unit we assume that it is a quantity_range type.
I am adding an extra check to prevent using arrays with more than 2 values for now.

nitrosx: @sbliven I agree with you, but that needs more effort that we have available at the moment. 
I see this as the first implementation that will be improved as soon as we have effort available and we gather more feedback from users once we merge the current version

nitrosx (approved):

test/TestData.js
sbliven: Isn't my suggestion less effort, as it works with the current code? But I'm fine with this implementation, at least unless we can find an alternative way to store unitSI.

PR #1755, opened 2025-03-12
fix: improve validate action validation

sbliven:
## Description
Improve the error message when you misconfigure a 'validate' action in a create job.

## Motivation
We ran into this error with a buggy jobConfig.yaml and had trouble tracking it back. This improves the error message.

It also uses `isValidateCreateJobActionOptions` correctly in create jobs. This doesn't actually have an impact, since we don't currently throw errors for extra fields in action configuration (maybe this is a bug?). This means that `isValidateCreateJobActionOptions` and `isValidateJobActionOptions` accept the same objects (all added fields are optional), which is why this "bug" wasn't picked up in testing.

## Fixes
(no issue filed)

## Tests included

- [x] Passing? 

## Documentation
- [ ] swagger documentation updated (required for API changes) - N/A

### DIFF
--- a/src/config/job-config/actions/validateaction/validateaction.service.ts
+++ b/src/config/job-config/actions/validateaction/validateaction.service.ts
@@ -5,7 +5,10 @@ import {
   JobDto,
 } from "../../jobconfig.interface";
 import { ValidateCreateJobAction, ValidateJobAction } from "./validateaction";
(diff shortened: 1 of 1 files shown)


sbliven:
I'm going to merge this minimal PR as is. Extra parameters in actions continue to be silently ignored.

PR #1758, opened 2025-03-13
feat(jobs): fix ownerUser, contactEmail logic and datasetList validation

despadam:
## Description
- `ownerUser`, `ownerGroup`, `contactEmail` in `CreateJobDto` were not typescript optional fields even thought they had the `@IsOptional()` annotation
- `instanceAuthorizationJobCreate` logic not covering all `contactEmail` cases
- `datasetList` was not deeply validated

## Motivation
Topics that came up while working on #1742 

## Fixes
- `ownerUser`, `ownerGroup`, `contactEmail` in `CreateJobDto` are now also typescript optional fields
- fixed `datasetList` validation to ensure that `pid` is a string and `files` is an array 

## Changes:
 `instanceAuthorizationJobCreate` logic
- in `admin` mode:
  - when `ownerUser` is provided and is the same as admin then the owner is the admin, otherwise if a different user is provided then we search for this user and assign the retrieved username as owner, unless it cannot be found in which case we set the admin as owner
  - for anonymous jobs, meaning when `ownerUser` and `ownerGroup` are not provided, then `contactEmail` is required to be provided via the dto
  - in general we prioritize `contactEmail` when that is provided via the dto. If it is not provided then we fall back to the `ownerUser` email and if `ownerUser` does not exist then the `contactEmail` is set to admin
- in `user` mode:
  - logic remains mainly identical, but we ensure that `contactEmail` is prioritized when provided

All tests where updated accordingly.

## Tests included

- [x] Passing? 

## Documentation

https://github.com/sbliven/scicat-documentation/pull/2

### DIFF
--- a/src/jobs/dto/create-job.dto.ts
+++ b/src/jobs/dto/create-job.dto.ts
@@ -27,7 +27,7 @@ export class CreateJobDto {
   })
   @IsString()
   @IsOptional()
-  readonly ownerUser: string;
+  readonly ownerUser?: string;
 
   @ApiProperty({
     type: String,
@@ -37,7 +37,7 @@ export class CreateJobDto {
   })
   @IsString()
   @IsOptional()
-  readonly ownerGroup: string;
+  readonly ownerGroup?: string;
 
   @ApiProperty({
     type: String,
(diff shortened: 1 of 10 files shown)


sofyalaski (approved):
Thank you Despina, that looks good

PR #1761, opened 2025-03-14
feat(attachments): attachmentsController v4

Junjiequan:
## Description

This PR introduces a new version of the attachment controller (v4) along with several important changes:

### Direct CRUD Operations:
Users can now directly perform CRUD actions on attachments via the /attachments endpoint. This eliminates the need to rely on nested endpoints such as `/datasets/:id/attachments` or `/samples/:id/attachments`.

### Enhanced Querying with findAllComplete:

The new /attachments endpoint leverages the `findAllComplete` service. This service enables users to use allowed query operators—including major MongoDB query operators—to perform flexible and powerful queries on attachment records.

### Schema Update & Migration:

- A new attachment schema has been created which introduces an additional relationships field.
- Obsolete fields (datasetId, sampleId, and proposalId) have been removed from the new schema.
- A migration script is provided to convert existing documents into the new format.
- For backward compatibility, a fallback conversion function is implemented for users who prefer to continue using the old DTO structure

### CASL Factory Refactoring for attachments:

- A refactored CASL factory has been implemented for both attachment endpoint access and instance access
- The attachment action enum has been narrowed down to 8 actions:
  - **Endpoint Actions:** AttachmentCreateEndpoint, AttachmentReadEndpoint, AttachmentUpdateEndpoint, AttachmentDeleteEndpoint
  - **Instance Actions:** AttachmentCreateInstance, AttachmentReadInstance, AttachmentUpdateInstance, AttachmentDeleteInstance
- The `cannot` logic from the permission-building process has been removed for attachment. Now, if a `can` action isn’t explicitly provided, the application ability will return false by default.

### Centralized Attachment Management
The new Attachment V4 controller consolidates all attachment CRUD operations into a single, centralized controller. Previously, the attachment service was directly invoked from multiple controllers (datasets, samples, and proposals), which led to duplicated logic and increased maintenance overhead.

### CASL Factory Refactoring Initiative
This PR includes the first step in a broader effort to refactor the CASL factory. The current CASL ability logic is complex and hard to understand, so we are narrowing down the actions for attachments and removing redundant “cannot” logic. This initial draft for the attachment refactoring will pave the way for similar improvements across the entire CASL factory in future work.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation


src/casl/casl-ability.factory.ts
```diff
@@ -368,6 +369,46 @@ export class CaslAbilityFactory {
     });
   }
 
+  attachmentEndpointAccess(user: JWTUser) {
+    const { can, cannot, build } = new AbilityBuilder(
```
martin-trajanovski: If cannot is not used we should not extract it at all.

src/casl/casl-ability.factory.ts
```diff
@@ -125,7 +126,7 @@ export class CaslAbilityFactory {
       cannot(Action.DatasetUpdate, DatasetClass);
       // -
       cannot(Action.DatasetAttachmentCreate, DatasetClass);
-      cannot(Action.DatasetAttachmentRead, DatasetClass);
+      can(Action.DatasetAttachmentRead, DatasetClass);
```
martin-trajanovski, line 129: Are we allowing access for non-authenticated users to the `DatasetAttachmentRead`?

src/casl/casl-ability.factory.ts
```diff
@@ -368,6 +369,46 @@ export class CaslAbilityFactory {
...
   }
 
+  attachmentEndpointAccess(user: JWTUser) {
+    const { can, cannot, build } = new AbilityBuilder(
+      createMongoAbility<PossibleAbilities, Conditions>,
+    );
+    /*
+     * default allow anyone to read attachments
+     */
+    can(Action.AttachmentReadEndpoint, Attachment);
```
martin-trajanovski, line 379: Do we really need to allow everyone to read attachments? I guess this is only on endpoint level and not on the instance afterwards. At least the comment should be updated maybe to clarify that we are doing instance check for attachments read in the logic layer itself. When you read this comments it might seem a bit strange that we allow anyone to read any of the attachments.

src/casl/casl-ability.factory.ts
```diff
@@ -368,6 +369,46 @@ export class CaslAbilityFactory {
...
+      ) {
+        /**
+         * authenticated users belonging to any of the group listed in ADMIN_GROUPS
+         */
+
+        can(Action.AttachmentCreateEndpoint, Attachment);
+        can(Action.AttachmentUpdateEndpoint, Attachment);
+        can(Action.AttachmentDeleteEndpoint, Attachment);
+
+        can(Action.Admin, Attachment);
```
martin-trajanovski: I am not sure I understand what is the point of having this `Admin` action here for the attachment specifically.

src/casl/casl-ability.factory.ts
```diff
@@ -1813,4 +1854,109 @@ export class CaslAbilityFactory {
         item.constructor as ExtractSubjectType<Subjects>,
     });
   }
+
+  attachmentInstanceAccess(user: JWTUser) {
+    const { can, cannot, build } = new AbilityBuilder(
```
martin-trajanovski: `cannot` is not used here as well and should be removed as eslint suggests.

src/casl/casl-ability.factory.ts
```diff
@@ -1813,4 +1854,109 @@ export class CaslAbilityFactory {
...
+        // -------------------------------------
+        // users belonging to any of the group listed in ADMIN_GROUPS
+        // -------------------------------------
+
+        can(Action.AttachmentReadInstance, Attachment);
+        can(Action.AttachmentCreateInstance, Attachment);
+        can(Action.AttachmentUpdateInstance, Attachment);
+        can(Action.AttachmentDeleteInstance, Attachment);
+
+        can(Action.Admin, Attachment);
```
martin-trajanovski: For me this is mix of a role based authorization and instance (or attribute) based one. Not sure if this is the right place to put the `Admin` action for the instance.

src/config/configuration.ts
```diff
@@ -26,7 +26,11 @@ const configuration = () => {
 
   const proposalGroups = process.env.PROPOSAL_GROUPS || "";
   const sampleGroups = process.env.SAMPLE_GROUPS || "#all";
-  const samplePrivilegedGroups = process.env.SAMPLE_PRIVILEGED_GROUPS || "";
+  const samplePrivilegedGroups =
+    process.env.SAMPLE_PRIVILEGED_GROUPS || ("" as string);
+  const attachmentGroups = process.env.ATTACHMENT_GROUPS || "#all";
+  const attachmentPrivilegedGroups =
+    process.env.ATTACHMENT_PRIVILEGED_GROUPS || ("" as string);
```
martin-trajanovski, line 33: If these groups are really needed they should be maybe listed in the readme file as well.

migrations/20250318101300-migrate-attachment-id-field-to-aid.js
```diff
@@ -0,0 +1,169 @@
...
+                },
+                {
+                  $concatArrays: [
+                    // Add dataset relationship if exists
+                    {
+                      $cond: [
+                        { $ifNull: ["$datasetId", false] },
+                        [
+                          {
+                            targetIds: ["$datasetId"],
```
nitrosx: I would code each relationship as the relationship between one target and the current attachment.
If the attachment is linked to multiple datasets, you should create multiple relationships.

migrations/20250318101300-migrate-attachment-id-field-to-aid.js
```diff
@@ -0,0 +1,169 @@
...
+                        [],
+                      ],
+                    },
+                    // Add sample relationship if exists
+                    {
+                      $cond: [
+                        { $ifNull: ["$sampleId", false] },
+                        [
+                          {
+                            targetIds: ["$sampleId"],
```
nitrosx: Same as datasets

migrations/20250318101300-migrate-attachment-id-field-to-aid.js
```diff
@@ -0,0 +1,169 @@
...
+                        [],
+                      ],
+                    },
+                    // Add proposal relationship if exists
+                    {
+                      $cond: [
+                        { $ifNull: ["$proposalId", false] },
+                        [
+                          {
+                            targetIds: ["$proposalId"],
```
nitrosx: Same as datasets

src/attachments/attachments.service.ts
```diff
@@ -16,44 +25,183 @@ export class AttachmentsService {
...
+      .findOne(convertedFilter, projection ?? {})
+      .exec();
   }
 
   async findOneAndUpdate(
     filter: FilterQuery<AttachmentDocument>,
-    updateAttachmentDto: PartialUpdateAttachmentDto,
+    updateAttachmentDto:
+      | PartialUpdateAttachmentObsoleteDto
+      | PartialUpdateAttachmentDto,
```
nitrosx: I would use just the current DTO here and transfer the conversion in the controller

src/attachments/attachments.service.ts
```diff
@@ -16,44 +25,183 @@ export class AttachmentsService {
...
   ): Promise<Attachment | null> {
     const username = (this.request?.user as JWTUser).username;
-    return this.attachmentModel
+    const convertedFilter =
+      this.convertObsoleteWhereFilterToCurrentSchema(filter);
+    const convertedDto =
+      this.convertObsoleteDtoToCurrentSchema(updateAttachmentDto);
```
nitrosx, line 120: See previous comment

src/attachments/schemas/relationship.schema.ts
```diff
@@ -0,0 +1,40 @@
...
+  })
+  targetType: AttachmentRelationTargetType;
+
+  @ApiProperty({
+    type: String,
+    description: "Relationship type (defaults to 'is attached to').",
+    default: "is attached to",
+  })
+  @Prop({ type: String, default: "is attached to" })
+  relationType?: string;
```
martin-trajanovski, line 35: I am wondering what other `relationTypes` do we have than `is attached to`

migrations/20250318101300-migrate-attachment-id-field-to-aid.js
```diff
@@ -0,0 +1,169 @@
+module.exports = {
+  async up(db, client) {
+    await db.collection("Attachment").updateMany(
```
martin-trajanovski, line 3: Have you tested this with some `testing/develop` database `dump` to verify that the `up` and `down` migrations work?

src/attachments/dto-obsolete/create-attachment-obsolete.dto.ts
nitrosx, line 1: Can you switch the folder name from dto-obsolete to dto-v3, or better move obsolete dtos to dto folder and  add v3 in the file name

src/attachments/dto-obsolete/create-sub-attachment-obsolete.dto.ts
nitrosx, line 1: Same as previous comment

src/attachments/dto-obsolete/output-attachment-obsolete.dto.ts
nitrosx, line 1: Same as previous comment

src/attachments/dto-obsolete/update-attachment-obsolete.dto.ts
nitrosx, line 1: Same as previous comment

src/attachments/dto/output-attachment.dto.ts
nitrosx, line 1: Is this following the datasets example?

src/attachments/pipes/attachment-filter-validation.pipe.ts
nitrosx, line 1: Do we have the same function for datasets v4? If so, can we generalize the code and have only one instance of the functionality?

src/attachments/schemas/relationship.schema.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Schema({ _id: false })
+export class AttachmentRelationshipClass {
+  @ApiProperty({
+    type: [String],
+    description: "Array of entity target IDs.",
+    default: [],
+  })
+  @Prop({ type: [String], default: [] })
+  targetIds: string[];
```
nitrosx: I will suggest one target one relationship

src/attachments/types/relationship-filter.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum AttachmentRelationTargetType {
+  DATASET = "dataset",
```
nitrosx, line 2: dataset -> is attachment of dataset
proposal -> is attachment of proposal
sample -> is attachment of sample

src/datasets/datasets.controller.ts
nitrosx, line 1: Change `CreateAttachmentObsoleteDto` to `Create AttachmentV3Dto`, and so foe

src/proposals/proposals.controller.ts
nitrosx, line 1: Same as datasets

src/samples/samples.controller.ts
nitrosx, line 1: Same as datasets

nitrosx (changes requested):

src/attachments/schemas/attachment.schema.ts
```diff
@@ -42,17 +50,15 @@ export class Attachment extends OwnableClass {
...
-  sampleId: string;
+  @ApiPropertyOptional({
+    type: Array,
+    items: { $ref: getSchemaPath(AttachmentRelationshipClass) },
+    default: [],
+    description:
+      "Array of relationships with other entities. It contains relationship type, target ids and target entity type.",
+  })
+  @Prop({ type: [AttachmentRelationshipSchema], default: [] })
+  relationships?: AttachmentRelationshipClass[];
```
martin-trajanovski, line 61: Why do we need the AttachmentRelationshipClass?(https://www.mongodb.com/docs/manual/applications/data-models-relationships/) I cannot understand the whole point behind it. And why not using the same way as before where we just have the document ids directly on the attachment document itself.
Doesn't this add extra complexity and maintainability problems? How can we do lookup on these fields?

martin-trajanovski (commented):
It is just a partial review. Maybe we can have a discussion about my thoughts here together with @nitrosx and see how to move forward with this.

migrations/20250318101300-migrate-attachment-id-field-to-aid.js
Junjiequan: The old attachment schema/dto expects datasetId to be a string and does not allow to be attached to multiple dataset. For the backward compatibility I think it's better to rollback it was before?

src/attachments/attachments.service.ts
Junjiequan: Then we need to force users who use the old endpoint `// PATCH /datasets/:id/attachments/:fk` to provide format of non-obsoleteDto. That's probably not a good way of handling backward compatibility

src/attachments/pipes/attachment-filter-validation.pipe.ts
Junjiequan: Yes, I was also thinking that. But I don't think it should be included in this PR

src/attachments/schemas/relationship.schema.ts
Junjiequan: Could you give me a scenario where the suggested change would be beneficial? I'm assuming you are suggesting to target: {ids:[], type:""}
My stance here is, fewer nested fields are easier to understand and less code is needed

src/attachments/types/relationship-filter.enum.ts
Junjiequan: Are you suggesting we should have AttachmentRelationType? The AttachmentRelationTargetType is used for targetType 
like:

```suggestion
export enum AttachmentRelationType{
  DATASET = "is attachment of dataset",
```

src/casl/casl-ability.factory.ts
Junjiequan: The `DatasetAttachmentRead` is used for endpoint access and non-authenticated user should be able to access/view published attachment

src/casl/casl-ability.factory.ts
Junjiequan: The comment needs to be updated.
And yes we need to allow anyone to be able to access the endpoint and check permission only for instance layer

src/casl/casl-ability.factory.ts
Junjiequan: The Admin action here is to replace individual viewAny action. The action is intended to be used on some special cases where checking viewAny are needed for filter transformation.
For example, we are mainly using viewAny for filter modifications like in [here](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/datasets.controller.ts#L189)

src/attachments/schemas/relationship.schema.ts
Junjiequan: For now, we will use only `is attached to`. But in the future we plan to expand it more, like having `is relation to` or something else

migrations/20250318101300-migrate-attachment-id-field-to-aid.js
Junjiequan: Yes, I tested it with staging database ( from backed up file )

src/attachments/schemas/attachment.schema.ts
Junjiequan: It will bring extra complexity, not necessary maintainability as the structures are unified. on the flip side it enhances flexibility greatly, as they are unified in one form and we can easily add new targetTypes like instrument, or job or anything else.
To name one benefit other than the flexibility: this structure allows us to display attachments in different formats, layouts, or colors based on the relationType

src/attachments/dto/output-attachment.dto.ts
Junjiequan: yes

src/attachments/attachments.service.ts
nitrosx: If users use API v4, they need to use the new Dto and the service should be converted to the new DTO.
Dto v3 are only available only in API v3, so I put the conversion in the related endpoints.

src/attachments/schemas/relationship.schema.ts
nitrosx: Apologies for not being clear.
One entry in the relationships can describe only a relationship between one attachment and one dataset or one sample or one instrument.
If the attachment is attached to multiple datasets, you should have as many relationships as the number of datasets. One per datasets.
Let me know if I was able to explain it better

src/attachments/types/relationship-filter.enum.ts
nitrosx: I see. 
I think it would be more clear, although it is a nice to have at the moment

src/casl/casl-ability.factory.ts
nitrosx: I would suggest to implement access as it is done in Datasets v4.
- endpoint `datasets` can only be accessed by authenticated users
- endpoint `datasets\public` can be accessed by unauthenticated users

src/attachments/dto/create-attachment.dto.ts
nitrosx, line 1: Should we call it `create-attachment.v4.dto.ts`?

src/attachments/dto/output-attachment.dto.ts
nitrosx, line 1: Should we call it `output-attachment.v4.dto.ts`?

src/attachments/dto/update-attachment.dto.ts
nitrosx, line 1: Should we name it `update-attachment-v4.dto.ts`?

src/attachments/pipes/attachment-filter-validation.pipe.ts
nitrosx, line 1: I wonder if we can share some code/functionalities with the dataset one?

src/attachments/attachments.v4.controller.ts
nitrosx, line 1: I'm confused about how you are managing the unauthenticated users?

nitrosx (changes requested):
Please check the comment on controller v4.
Renaming dto files to v4.dto is optional, but a nice to have

src/attachments/pipes/attachment-filter-validation.pipe.ts
Junjiequan: yes, we can put the function in a utility folder that can be shared across different modules.
I think it's better to create a new PR for that change

src/attachments/dto/create-attachment.v4.dto.ts
```diff
@@ -0,0 +1,8 @@
+import { IsOptional, IsString } from "class-validator";
+import { UpdateAttachmentDto } from "./update-attachment.v4.dto";
```
nitrosx: Should we rename `UpdateAttachmentDto` to `UpdateAttachmentV4Dto`?

src/attachments/dto/create-attachment.v4.dto.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export class CreateAttachmentDto extends UpdateAttachmentDto {
```
nitrosx: Should we rename `CreateAttachmentDto` to `CreateAttachmentV4Dto`?

src/attachments/dto/output-attachment.v4.dto.ts
nitrosx, line 1: Should we rename `OutputAttachmentDto` to `OutputAttachmentV4Dto`?

src/attachments/dto/update-attachment.v4.dto.ts
nitrosx, line 1: Should we rename `UpdateAttachmentDto` to `UpdateAttachmentV4Dto`?

nitrosx (commented):
Check my comments and see if they make sense.
Besides my naming suggestions, it looks good

nitrosx (approved):

PR #1770, opened 2025-03-18
feat(jobs): replace ApiProperty with openapi cli-plugin

despadam:
## Description

~~It seems that by applying these changes the v3, v4 dtos and schemas coexist in swagger, so one needs to be mindful of the api version in use.~~ --> Incorrect: just an error on my side due to a faulty container. There are no changes in swagger

## Motivation
https://github.com/SciCatProject/scicat-backend-next/pull/1741#issuecomment-2731026311

## Changes:

They are no longer using `ApiProperty`:
- `CreateJobDto`
- `UpdateJobDto`
- `JobClass`

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/jobs/dto/update-job.dto.ts
+++ b/src/jobs/dto/update-job.dto.ts
@@ -1,29 +1,23 @@
-import { ApiProperty, ApiTags } from "@nestjs/swagger";
+import { ApiTags } from "@nestjs/swagger";
 import { IsOptional, IsString, IsObject } from "class-validator";
 
 @ApiTags("jobs")
 export class UpdateJobDto {
-  @ApiProperty({
-    type: String,
-    required: true,
-    description: "Updated job status code for the current status.",
-  })
+  /**
+   * Updated job status code for the current status.
+   */
   @IsString()
   readonly statusCode: string;
 
-  @ApiProperty({
-    type: String,
-    required: false,
-    description: "Additional message about the current job status.",
-  })
+  /**
+   * Additional message about the current job status.
+   */
   @IsString()
   readonly statusMessage: string;
 
-  @ApiProperty({
-    type: Object,
-    required: false,
-    description: "Dataset archiving results.",
-  })
+  /**
+   * Additional information, i.e. dataset archiving results.
+   */
   @IsObject()
   @IsOptional()
   readonly jobResultObject?: Record<string, unknown>;
--- a/src/jobs/schemas/job.schema.ts
+++ b/src/jobs/schemas/job.schema.ts
@@ -32,10 +32,9 @@ export class JobClass extends OwnableClass {
   })
   id: string;
 
-
   /**
-  * Defines the user that owns this job.
-  */
+   * Defines the user that owns this job.
+   */
   @Prop({
     type: String,
     index: true,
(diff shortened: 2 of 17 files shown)


src/jobs/dto/update-job.dto.ts
```diff
@@ -1,29 +1,23 @@
...
-  @ApiProperty({
-    type: String,
-    required: false,
-    description: "Additional message about the current job status.",
-  })
+  /**
+   * Additional message about the current job status.
+   */
   @IsString()
   readonly statusMessage: string;
```
emigun, line 16: Should this be required? It was not before.

emigun:
"It seems that by applying these changes the v3, v4 dtos and schemas coexist in swagger, so one needs to be mindful of the api version in use."

Can you elaborate on this? The changes you made shouldn't change the swagger.

despadam:
> "It seems that by applying these changes the v3, v4 dtos and schemas coexist in swagger, so one needs to be mindful of the api version in use."
> 
> Can you elaborate on this? The changes you made shouldn't change the swagger.

@emigun yes, sorry about that! Please ignore it. It was an error on my side due to a misconfigured container because I was previously working on backwards compatibility. I restarted everything and there are no changes in swagger now

src/jobs/dto/update-job.dto.ts
despadam: Currently when users try to perform a job update without providing `statusMessage` in the Dto, they will get an error saying `"statusMessage must be a string"`. So since this check already exists, for this reason I decided to have the field as required in the dto.

sbliven: I think the intent was that both `statusCode` (a short machine-readable string) and `statusMessage` (a human-readable description) would be required. The JobClassV3 maps both to the old `statusMessage` field.

src/jobs/schemas/job.schema.ts
```diff
@@ -28,107 +32,80 @@ export class JobClass extends OwnableClass {
...
+  /**
+   * Contains the dataset archiving results. Initially empty, then provided during update.
+   */
   @Prop({
     type: Object,
     required: false,
+    default: {},
   })
-  jobResultObject: Record<string, unknown>;
+  jobResultObject?: Record<string, unknown>;
```
sbliven: Should this be required (but with the {} default)?

despadam: Done, makes more sense like this

sbliven (approved):
I ran a diff of the openapi changes from this PR and all of them look fine.

Can you go further and delete all the type annotations too (`@Optional() @IsString()` etc)? @emigun left these from #1611 but I did a quick test and the plugin seems to infer them from the typescript type. On the other hand, #1590 remains open and it's probably better to update all DTOs and schemas consistently across scicat after this PR is merged.

emigun:
> Can you go further and delete all the type annotations too (`@Optional() @IsString()` etc)? @emigun left these from #1611 but I did a quick test and the plugin seems to infer them from the typescript type. On the other hand, #1590 remains open and it's probably better to update all DTOs and schemas consistently across scicat after this PR is merged.

That was on purpose to leave them as is. They are used for validation and they are not related to the swagger json.

emigun (approved):
:star:

PR #1774, opened 2025-03-20
fix: permission for ownerUser to read jobs

sofyalaski:
## Description
As the DTO enables user to provide both ownerUser and OwnerGroup, anyone who falls under the specified fields should be able to read the job.
## Motivation
Owner user was not getting access, only the ownerGroup.
## Fixes
Casl was fixed and explicitly allows for a user specified in ownerUser to access the job and for a group specified in ownerGroup to get access to the job.

## Changes:
Tests are added to check this rules.
Test utils introduced new function to get an email of the user, which is used to compare in the contactEmail field in certain tests. 
## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/casl/casl-ability.factory.ts
+++ b/src/casl/casl-ability.factory.ts
@@ -1454,7 +1454,7 @@ export class CaslAbilityFactory {
 
           can(Action.JobCreateOwner, JobClass, {
             ownerGroup: { $in: user.currentGroups },
-          });       
+          });
         } else {
           /**
            * authenticated users not belonging to any special group
@@ -1470,7 +1470,6 @@ export class CaslAbilityFactory {
               String(v).includes("#dataset"),
             ),
           ];
-   
 
           if (
             jobCreateInstanceAuthorizationValues.some(
--- a/src/casl/casl-ability.factory.ts
+++ b/src/casl/casl-ability.factory.ts
@@ -1437,6 +1437,12 @@ export class CaslAbilityFactory {
           ...user.currentGroups.map((g) => "@" + g),
           user.username,
         ];
+        can(Action.JobReadAccess, JobClass, {
+          ownerGroup: { $in: user.currentGroups },
+        });
+        can(Action.JobReadAccess, JobClass, {
+          ownerUser: user.username,
+        });
         if (
           user.currentGroups.some((g) =>
             this.accessGroups?.createJob.includes(g),
@@ -1448,10 +1454,7 @@ export class CaslAbilityFactory {
 
           can(Action.JobCreateOwner, JobClass, {
             ownerGroup: { $in: user.currentGroups },
-          });
-          can(Action.JobReadAccess, JobClass, {
-            ownerGroup: { $in: user.currentGroups },
-          });
+          });       
         } else {
           /**
            * authenticated users not belonging to any special group
@@ -1467,11 +1470,8 @@ export class CaslAbilityFactory {
               String(v).includes("#dataset"),
             ),
           ];
+   
 
-          can(Action.JobReadAccess, JobClass, {
-            ownerGroup: { $in: user.currentGroups },
-            ownerUser: user.username,
-          });
           if (
             jobCreateInstanceAuthorizationValues.some(
               (a) => jobConfiguration.create.auth === a,
--- a/test/JobsDatasetOwner.js
+++ b/test/JobsDatasetOwner.js
@@ -25,12 +25,11 @@ let datasetPid1 = null,
   encodedJobOwnedByGroup5 = null,
   jobId6 = null,
   encodedJobOwnedByAnonym = null;
-  encodedJobOwnedByGroup13 = null;
-  encodedJobOwnedByGroup12 = null;
-  encodedJobOwnedByGroup11 = null;
+  jobId7 = null,
+  jobId8 = null,
+  jobId12 = null;
+  jobId21 = null;
   jobId31 = null;
-  jobId32 = null;
-  jobId33 = null;
 
 
 const dataset1 = {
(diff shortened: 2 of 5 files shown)


src/casl/casl-ability.factory.ts
```diff
@@ -1437,6 +1437,12 @@ export class CaslAbilityFactory {
           ...user.currentGroups.map((g) => "@" + g),
           user.username,
         ];
+        can(Action.JobReadAccess, JobClass, {
+          ownerGroup: { $in: user.currentGroups },
+        });
+        can(Action.JobReadAccess, JobClass, {
+          ownerUser: user.username,
+        });
```
nitrosx, line 1445: are they equivalent as a functional `or`?

nitrosx, line 1445: are they equivalent as a functional `or`?

test/LoginUtils.js
```diff
@@ -46,3 +46,20 @@ exports.getTokenAD = function (appUrl, user, cb) {
...
+      .set("Accept", "application/json")
+      .end((err, res) => {
+        if (err) {
+          reject(err);
+        } else {
+          resolve(res.body.user.email);
+        }
+      });
+  });
+};
```
nitrosx: Do we need to make a POST call to retrieve the email?
Would it be possible to refactor it to just use the user service?

src/casl/casl-ability.factory.ts
sofyalaski: Yes, this is what the Casl documentation says to use multiple can/cannot rules for logical operators: https://casl.js.org/v6/en/guide/conditions-in-depth#why-logical-query-operators-are-not-included

test/LoginUtils.js
despadam: I would suggest one function that returns both token and email, instead of two separate ones for the same user

nitrosx: My bad and my apologies!!!
I did not see that this function was in test.
Please leave it as it is and ignore my comment.
Marking this issue resolved.

nitrosx (approved):

src/casl/casl-ability.factory.ts
sbliven, line 1445: I guess it would be easy to add read access to `accessGroups` here as well (in a future PR).

sbliven: (And isPublished)

despadam (approved):
Looks good! My only comment is whether we can have one function that returns both token and email for the admin, instead of two separate ones

despadam:
Also I think we should merge this one first and afterwards doublecheck that #1742 agrees with these changes

test/LoginUtils.js
sbliven: I also would have expected this to use `/api/v3/users/my/self` with an already-authorized user. But like @nitrosx said, it doesn't really matter since it's in test.

test/JobsDatasetOwner.js
```diff
@@ -617,4 +724,41 @@ describe("1150: Jobs: Test New Job Model Authorization for #datasetOwner jobs co
...
+  it("0320: Access jobs as a User1 ", async () => {
+    return request(appUrl)
+      .get(`/api/v3/Jobs/`)
+      .send({})
+      .set("Accept", "application/json")
+      .set({ Authorization: `Bearer ${accessTokenUser1}` })
+      .expect(TestData.SuccessfulGetStatusCode)
+      .expect("Content-Type", /json/)
+      .then((res) => {
+        res.body.should.be.an("array").to.have.lengthOf(7);
```
sbliven, line 735: Maybe this should check the actual jobs ids returned? That might be easier to maintain than just checking the length, which tends to change as new tests get added.

test/LoginUtils.js
```diff
@@ -46,3 +46,20 @@ exports.getTokenAD = function (appUrl, user, cb) {
       }
     });
 };
+
+
+exports.getEmail = function (appUrl, user) {
+  return new Promise((resolve, reject) => {
+    request(appUrl)
+      .post("/api/v3/auth/Login?include=user")
```
sbliven, line 53: I think `include=user` is ignored. At least it's not documented in swagger.

sbliven (approved):

src/casl/casl-ability.factory.ts
sofyalaski: I agree with accessGroup but what is a published job?

nitrosx: Is a job that was made public, so anybody can view it!!!
I do not have a use case in mind at the moment... I'm sure it will come to me.
We can change the logic in a future PR

test/JobsDatasetOwner.js
nitrosx: On second thought, I agree with @sbliven.
Can we make the change in a later PR?

test/LoginUtils.js
nitrosx: Good catch!!!

test/JobsDatasetOwner.js
sofyalaski: In general, I agree with this. What I proposed was to add specific ids for this testing file, but maybe leave it a count for the main file (JobsAll)?

nitrosx: Checking ids will be awesome, but in a later PR!!!

nitrosx: Open an issue, so we know that we have to work on.

nitrosx (commented):
Ready to be merged

test/JobsDatasetOwner.js
sofyalaski: https://github.com/SciCatProject/scicat-backend-next/issues/1778

Issue #1688, opened 2025-02-05
Dataset size property reset to 0 after partial update


HayenNico:
## Summary
Datasets size property gets reset to 0 after an update via PATCH /api/v3/datasets/{pid}, despite associated origdatablocks being unchanged. Leads to wrong information displayed in the FE dataset table. Reproduced on latest BE/FE versions on main branch.

## Steps to Reproduce
Create a dataset and at least one origdatablock with size > 0. The size property on the dataset will correctly update after datablock creation.
Run a partial update on any field via PATCH /api/v3/datasets/{pid}.

## Current Behaviour
Dataset size is calculated and managed by the backend, and will be updated upon changes to the origdatablocks and datablocks. An update to the dataset itself resets size to 0 (if not specified) or allows one to set any value for the size property irrespective of existing datablocks. This is inconsistent with the rules enforced for datablock updates.

## Expected Behaviour
The size property should reflect the sum of the datablock sizes at all times and should be recalculated after an update to the dataset. Arguably, size should not be a valid field for dataset creation/update DTOs and instead be completely managed by the backend.

## Details

![Image](https://github.com/user-attachments/assets/fd67c091-e776-4dd8-8e2c-261c8f6e092f)
![Image](https://github.com/user-attachments/assets/3ceddabd-20c8-4769-b132-a21982bd3950)
Example dataset with size 2GB

![Image](https://github.com/user-attachments/assets/03f36e51-dc4b-4c15-9762-18b58149baa8)
Test update query (could be anything) -> 200 OK

![Image](https://github.com/user-attachments/assets/c983ad57-dcf0-4f24-9e7c-556ca12a72ad)
![Image](https://github.com/user-attachments/assets/8242d92f-3fa5-4438-b9bd-7ce4bcf0aaf6)
Size reset to 0 B on dataset

![Image](https://github.com/user-attachments/assets/e913f07c-8e88-418e-9687-8a4d7de6bd5c)
Origdatablocks are there and intact

martin-trajanovski:
@nitrosx should we close this one or we should also add the fix for the affected datasets that is in the following PR https://github.com/SciCatProject/scicat-backend-next/pull/1854? Also if we want to fully fix this issue we need both I guess. The one that is already merged and the migration script to be run.

PR #1802, opened 2025-03-28
fix: remove default values in DTOs that cause value overwrite on update

martin-trajanovski:
## Description
This PR aims to fix the `size` and `numberOfFiles` fields overwrite in the partial update of a dataset.

## Motivation
`size` and `numberOfFiles` fields were overwritten every time when a partial update on a dataset was performed.

## Fixes

* https://github.com/SciCatProject/scicat-backend-next/issues/1688

## Changes:

* changes made

## Tests included

- [x] Passing? 

## Documentation


HayenNico:
Small addition: This will also fix #1688 as it's caused by the same default values. I think we should still consider removing `size` as a DTO field for Dataset, since it's mostly managed/updated by the backend when changing OrigDatablocks.

nitrosx:
@HayenNico should I bring it up in the collaboration meeting and ask opinion there?

That said, I think that this field is important given that one dataset can have multiple origDatablock and multiple files associated with it

HayenNico:
@nitrosx I did not mean to remove `size` from Dataset schema, only the DTOs. In the code, when an OrigDatablock is created/updated, the size of the associated Dataset is also recalculated. As such, I'd call that a field that's managed by the backend, and it should not be available to edit in the Dataset DTO itself.
Right now, one could in principle overwrite `size` in a Dataset via the API, such that it does not have any relation to the data files anymore. See also #1688 for more details

martin-trajanovski:
> @nitrosx I did not mean to remove `size` from Dataset schema, only the DTOs. In the code, when an OrigDatablock is created/updated, the size of the associated Dataset is also recalculated. As such, I'd call that a field that's managed by the backend, and it should not be available to edit in the Dataset DTO itself. Right now, one could in principle overwrite `size` in a Dataset via the API, such that it does not have any relation to the data files anymore. See also #1688 for more details

Yes I agree with this. I think as we are calculating this on the backend side it is a good idea to remove it from the update and create DTOs and just leave it in the database schema. This removes the possibility to update those fields by mistake and have the backend calculations as the only source of truth when it comes to total size and number of files. I am not sure if I should do that as part of this PR or it can be done separately in another one. @nitrosx let me know what do you think

nitrosx:
Remove `size` from the create and update DTO and manage it internally.
It should stay in the output DTO.

I would say create an additional PR

nitrosx (approved):

PR #1804, opened 2025-03-31
feat: add mongodb session store

### LINKED ISSUE #1803: Enable multiple backend replicas
## Summary
Currently, it is only possible to have a single backend replica due to the login session being stored in the backend.

## Steps to Reproduce
Scale up the backend

## Current Behaviour
Scicat only works correctly with one backend.

## Expected Behaviour
In order to enable high availability and load balancing one should be able to use more than one backend.

## Details
Sessions should be stored either on the client side or on the backend using a session store. Since there is already a mongodb, this can be used to store the user sessions

belfhi:
resolves #1803


src/main.ts
```diff
@@ -95,21 +96,28 @@ async function bootstrap() {
...
   if (expressSessionSecret) {
     app.use(
       session({
         secret: expressSessionSecret,
         resave: false,
         saveUninitialized: true,
+        store: MongoStore.create({
+          mongoUrl, // MongoDB connection string
+          collectionName: "sessions", // Collection name for storing sessions
+          ttl: 24 * 60 * 60, // Session TTL (24 hours)
```
minottic, line 110: should this equal [this](https://github.com/SciCatProject/scicat-backend-next/blob/master/CI/E2E/backend.env#L5)?

src/main.ts
```diff
@@ -95,21 +96,28 @@ async function bootstrap() {
...
+
   if (expressSessionSecret) {
     app.use(
       session({
         secret: expressSessionSecret,
         resave: false,
         saveUninitialized: true,
+        store: MongoStore.create({
+          mongoUrl, // MongoDB connection string
+          collectionName: "sessions", // Collection name for storing sessions
```
minottic, line 109: I think "sessions" is the default, so maybe we can avoid specifying it here?

src/main.ts
```diff
@@ -95,21 +96,28 @@ async function bootstrap() {
...
   );
+  const mongoUrl = configService.get<string>("mongodbUri");
+
   if (expressSessionSecret) {
     app.use(
       session({
         secret: expressSessionSecret,
         resave: false,
         saveUninitialized: true,
+        store: MongoStore.create({
```
minottic, line 107: shall we leave space for choosing if to use mongoStore or in memory one?

minottic (commented):
Thanks for the changes! I have left a couple of comments, maybe we could also add a test here.

src/main.ts
```diff
@@ -1,4 +1,5 @@
 import session from "express-session";
+import MongoStore from "connect-mongo";
```
nitrosx, line 2: I wonder if we need to have both mongoose and connect-mongo?
Can we achieve the same functionality just with mongoose?

src/main.ts
```diff
@@ -95,21 +96,28 @@ async function bootstrap() {
...
         resave: false,
         saveUninitialized: true,
+        store: MongoStore.create({
+          mongoUrl, // MongoDB connection string
+          collectionName: "sessions", // Collection name for storing sessions
+          ttl: 24 * 60 * 60, // Session TTL (24 hours)
+        }),
+        cookie: {
+          secure: true,
+        },
```
nitrosx, line 114: According to this post, it is possible to have only one connection to the database.
https://stackoverflow.com/questions/66731669/mongoose-and-connect-mongo

Can you look into it and see if we can reduce duplication?

nitrosx (changes requested):

Junjiequan:
I'm second to nitrosx's comment, to avoid 2 db connections. 
Although I'm more inclined utilizing Redis similar solutions to deal with temporary data like sessions. However, in this case that might just be a overkill

src/main.ts
martin-trajanovski: There is this mongoose example in the connect-mongo package (https://github.com/jdesboeufs/connect-mongo/blob/master/example/mongoose.js). But as we use the wrapper for mongoose @nestjs/mongoose then there might be some way to get the connection in a "nestjs" way. So I guess the idea is to reuse the existing mongoose connection if possible here instead of creating a new connection for the connect-mongo.

src/main.ts
nitrosx: +1
I would add a property in configuration to select the current behavior or session in mongodb.

nitrosx:
@belfhi do you have any update on this PR? Thank you so much

Junjiequan:
Shall we close this PR as the feature #1884  has been merged into the master branch

belfhi:
yes, sorry I closed the issue but forgot avout this PR 👍

PR #1815, opened 2025-04-03
fix: oidc returnUrl value was being discarded

GBirkel:
## Description

This is a small set of changes that enforces consistency with "returnUrl" and "successUrl".

Previously, "returnURL" and "returnUrl" were being mixed in the code, and this resulted in some confusion between values carried in URI parameters and values stored in the user session while OIDC authorization was taking place.

(To make the "returnUrl" process work fully, a front-end fix is needed as well, which will come along shortly.  The front-end error is unrelated to this one.)

## Motivation / Fixes

Making the names consistent has fixed an issue where the "returnUrl" value was not surviving the trip from front end -&gt; back end -&gt; front end again.  The back-end tests were already passing but in live operation "returnUrl" was being discarded in favor or "/datasets".

## Changes:

"returnURL" has been changed to "returnUrl".  "successURL" exists alongside "returnURL" and was changed to "successUrl" for consistency.

## Tests included

All tests pass as before.

## Documentation

No changes needed.


nitrosx (approved):

nitrosx:
@GBirkel could you please merge into your PR the latest master?
Thanks

nitrosx:
@GBirkel would you mind fix the title and make sure that your PR isd ready to be merge?
I would like to merge it ASAP.
Thank you so much

omkar-ethz:
Hello, I was the author of the original feature: https://github.com/SciCatProject/scicat-backend-next/pull/1714 https://github.com/SciCatProject/frontend/pull/1753
And although confusing, the feature was well tested and worked as expected. `returnURL` parameter was a query parameter to /auth/oidc _request_. And `returnUrl` was a query parameter in /auth/callback _response_ (i.e. the redirected URL as a result of /auth/callback). Perhaps the reason you could not get returnUrl to persist is misconfigured session (maybe EXPRESS_SESSION_SECRET not set in config)?

However, this change breaks the feature: as the frontend still relies on `returnURL`: https://github.com/SciCatProject/frontend/blame/335a3cd5e32aff6c39996ade28efb4ca72deb916/src/app/_layout/app-header/app-header.component.ts#L52. Should I create a frontend PR to change it to `returnUrl` / or should we revert this PR instead?

omkar-ethz:
Also, this creates a breaking change in the API (clients who were relying on `returnURL` will all break, e.g. I will need to change it to `returnUrl` also [here](https://github.com/paulscherrerinstitute/scilog/blob/main/scilog/src/app/core/config/server-settings.service.ts#L19) in a different project) -- although I imagine the feature was recent enough that not many clients relied on it.

omkar-ethz:
Apologies for the confusion. Although the issue wasn't due to parameter names (the feature worked fine when OIDC_SUCCESS_URL is /auth-callback), the frontend PR: https://github.com/SciCatProject/frontend/pull/1807 fixes the case when OIDC_SUCCESS_URL is /login. As the frontend PR uses the consistent naming introduced in this PR, I'm no longer asking for this PR to be reverted. I will update my other client code to use the updated `returnUrl`. Thanks!

nitrosx:
Thank you for clarifying and helping out.

PR #1830, opened 2025-04-08
feat: make frontend config and theme file locations configurable

HayenNico:
## Description
Add backend configuration options to customize the file locations of `frontend.config.json` and `frontend.theme.json` via new environment variables `FRONTEND_CONFIG_FILE` and `FRONTEND_THEME_FILE`. The default location has been changed from /src/config to project root for consistency with other customizable configuration files (up for discussion).

## Motivation
The backend will serve a configuration and theme to a connected frontend via the `/api/v3/admin/config` and `/api/v3/admin/theme` endpoints. The data sources for these endpoints are currently hardcoded to `/src/config/frontend.config.json` and `/src/config/frontend.config.json`. This is inconsistent with other configuration files which are located at the project root and can be configured in `.env`.

This PR introduces environment variables `FRONTEND_CONFIG_FILE` and `FRONTEND_THEME_FILE` for customizing the location of the json files. The default location of both files has been moved to the project root directory.

## Changes:

* added `FRONTEND_CONFIG_FILE` and `FRONTEND_THEME_FILE` environment variables
* admin.service gets config and theme from ConfigService instead of static location
* moved `frontend.config.json` and `frontend.theme.json` from `/src/config` to project root directory
* null returns from admin controller/service now indicate missing frontend configuration or theme
* admin.servoce.spec unit tests use mock objects instead of static json

## Tests included

- [x] Passing? 

## Documentation


HayenNico:
Will leave conflicts between master and this branch alone until some points are discussed. I'll transfer the most recent frontend.config.json to the new file once ready to merge.

- Is the change in default location for config and theme sensible?
- Do downstream dependencies (SciCatLive, local installations) benefit from this or is this a hindrance?

nitrosx (approved):

nitrosx:
@HayenNico I think that the changes is sensible, although I would move all the configuration files under a folder called `config`, so we do not struggle to find them in between all the other files.

It definitely impact existing deployment, unless they specify the relative env variables to the current location that they are using now. Maybe we should explicitly say that in the documentation.

HayenNico:
@nitrosx Okay, in that case how about this plan:
For this PR I'll revert the default location for FE config back to the original spot, so things will continue to work for any deployment with the added new .env options in case someone wants to move the files.
For reorganizing all different configuration files in one place, I'll open a new PR. Do you think it would make sense to add functionality for multiple fallback cases for where to find a file? For eample: If the location in .env does not work, should the config service then look in a few standard places (config folder, project root) if it can find something?

nitrosx (approved):

GBirkel:
A note for future searchers: I have Helm manifests that map custom theme and config files into my Kubernetes deployment at e.g `/home/node/app/dist/config/frontend.config.json` .

This change broke my deployment in a mysterious way because the call to `/api/v3/admin/theme` from the front end would return NO response (not even a 404) when it didn't find a theme file, and the front end would try and read properties out of it, and crash.

I changed my Helm manifests to map the files to `/home/node/app/src/config/frontend.config.json` instead, and the problem was fixed.

It might be a good idea to tweak this code in the future so it returns something like a 404 if it can't find a config file, rather than ..... nothing.

patrickmadela:
I ran into the exact same issue on my side. The frontend was stuck on `Loading...` because the `/api/v3/admin/theme` and `/api/v3/admin/config` endpoints returned empty responses after upgrading to 4.17.x.

I also had custom config and theme files mapped to `/home/node/app/src/config/frontend.config.json` and `/home/node/app/src/config/frontend.theme.json`, to fix the problem.

Thanks for sharing the workaround — it helped me resolve it too!

I agree that returning a clearer error (like a 404) when the config/theme files are missing would make debugging much easier.

Issue #1405, opened 2024-09-02
Support multiple storage locations for datasets


sbliven:
### Summary

We would like to add the ability to store data at different locations. This is primarily driven by our plan to allow users from multiple institutes to share a SciCat instance. The physical location in which data is stored will changed based on the affiliation of the dataset owner. While data transfer is handled outside of SciCat, Scicat needs to track the storage location so that the archive system can retrieve data from the correct storage. I could also imagine this feature being used for single facilities that wanted to support multiple archive systems (eg for disk and tape storage).

### Suggested changes

1. Add `storageLocation` as a top-level field in Dataset. All DataBlocks associated with that dataset are assumed to be in the same storage location.
2. Allow Job actions to be filtered by storage location. This can be implemented as a FilterAction with no changes to the job data model
3. Ensure jobs involve datasets from only a single storageLocation. This can probably be implemented on the backend as a ValidateAction without code changes, but may require some changes in the frontend.
4. Permission model might need to be updated, depending on use cases

bpedersen2:
Would also be welcome at MLZ, where likely different storage systems will be in operation.

dylanmcreynolds:
We are very interested in this as well.

sbliven:
There were a couple good points discussed at the meeting.

1. Moving data between different storage at one facility is an interesting use case. For instance, moving datasets between S3-accessible storage and parallel filesystems. Datasets might be in multiple places; this could be represented by several DataBlocks with the same files but different storageLocations.

1. (Martin) SciCat is not a data management system. Managing multiple storage locations, tiering, redundancy, etc. is better managed by a dedicated storage management system. Such systems abstract the physical storage location from the file identifier. A SciCat instance with such a system would store the file identifiers in DataBlock. Changing the physical storage location could then be done without updating the scicat database.

   - I agree with this in theory, but replacing legacy archive systems with a completely new architecture might be impractical. Pragmatic solutions based off the existing architecture should also be considered.

2. (@minottic) Since the storageLocation is associated with the archived version, it might make sense to add it to `DataBlock` rather that `Dataset`. The desired storageLocation would first be added to the archive job (in `jobParams`) and then the archive system would be responsible for setting the `storageLocation` when the DataBlocks are created after archiving. Retrieve jobs would then request the datablocks from the relevant location.

[...]

PR #1843, opened 2025-04-15
feat: add `datasetlifecycle.storageLocation` property

sbliven:
## Description

Adds `datasetlifecycle.storageLocation` to the dataset schema.

## Motivation

Most sites have a single storage location for all data, so they don't need this feature. However with this it is possible to configure storage at several different sites. 

Sites that use this should also add validation to their job configuration to ensure that archive and retrieve jobs only apply to a single storageLocation.

## Fixes

* Support multiple storage locations for datasets (#1405)

## Changes:

* add `datasetlifecycle.storageLocation` property

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/schemas/lifecycle.schema.ts
+++ b/src/datasets/schemas/lifecycle.schema.ts
@@ -159,6 +159,15 @@ export class LifecycleClass {
   })
   @Prop({ type: Boolean, default: false, required: false })
   retrieveIntegrityCheck?: boolean;
+
+  @ApiProperty({
+    type: String,
+    required: false,
+    description:
+      "Location of the data on the archive system. Allowed values are facility-specific and may be validated when submitting scicat jobs relating to this dataset. Facilities with a single storage location can leave this field empty.",
+  })
+  @Prop({ type: String, required: false })
+  storageLocation?: string;
 }
 
 export const LifecycleSchema = SchemaFactory.createForClass(LifecycleClass);


despadam (approved):

nitrosx (commented):
No problem to add the field. 
I just wonder if the dataset life cycle is the best location where to add it?
Wouldn't be better to add it to Datablock, given that they should be the custodian of the archived files?

sbliven:
@nitrosx We discussed quite a bit whether this should be in `Dataset`, `OrigDatablock`, `Datablock`, or just within the `jobParams`. We settled on the principal that datasets should be stored in one and only one location. That corresponds best to `Dataset`.

If you really wanted to implement Datablock-level storageLocation then it leads to a whole new level of complexity:
- When fetching Datablocks for a Dataset you need to group them by storageLocation; files could be duplicated over several Datablocks at different locations
- You either need to ensure that all files are accessible from all locations, or else add file-level retrieval mechanisms to figure out which datablocks need to be retrieved to cover all files
- Ingest jobs need to include the storageLocation(s) where the data should be stored. Likewise retreive jobs need to specify which location to retrieve from.
- Both the frontend and ingestors would need to be updated to make multiple storageLocations work

We don't currently have a need for this, so I'd prefer keeping it in Dataset. And since `datasetlifecycle` already has a lot of stuff relating to the archiving system, that seemed preferable to a top-level field.

nitrosx (approved):

Issue #1775, opened 2025-03-20
Should JobClass extend OwnableClass?


despadam:
## Summary

After reviewing the code for #1741 and #1742 we have been wondering whether `JobClass` should actually extend `OwnableClass`. We use `ownerGroup` coming from `OwnableClass` in many cases during authorization when normal users want to create jobs from themselves.

## Current Behaviour

Because we extend `OwnableClass` we see that `JobClass` also inherits `accessGroups` and `isPublished` which are actually never used during job authorization and not needed to be included in the job schema.

## Expected Behaviour

`accessGroups` and `isPublished` should not be included in `JobClass` since they are never used.

## Details

Should `JobClass` really extend `OwnableClass`? Perhaps it could just extend `QueryableClass` and we could just  include the `ownerGroup` field directly in the job schema.
In general, in which cases is  `OwnableClass` actually needed?

PR #1845, opened 2025-04-17
feat: introduce jobAdmin configuration option

sofyalaski:
## Description
Introduces a new type of configuration `jobAdmin`. For `#datasetX` checks the corresponding access to the dataset for ownerUser and ownerGroup for whom job is being created.

## Motivation
There was no possibility to configure a job such that only admin(s) can create it. This PR introduces the feature. 
Additionally, checks added on access to the dataset even if elevated users create job for someone. 

## Changes:
* new configuration `jobAdmin` is added. It allows to create jobs only for users who are either part of ADMIN_GROUPS or CREATE_JOB_PRIVILEGED_GROUPS. The same configuration is added to the update method, which will allow to update jobs to users of ADMIN_GROUPS or UPDATE_JOB_PRIVILEGED_GROUPS only.
* ADMIN_GROUPS is still considered to be more elevated, compared to CREATE_JOB_PRIVILEGED_GROUPS. As such, when user of ADMIN_GROUPS create a job _for someone_ in `#datasetPublic`, `#datasetAcces` and `#datasetOwner`, it isn'r checked if this someone has an appropriate access to dataset. On the contrary, the same configuration for user of 
CREATE_JOB_PRIVILEGED_GROUPS performs checks to see if the user for whom job is created has appropriate access. 
* Since it's possible to create a job for another user and another group, where user is not part of that group, it is checked that both have an access to the dataset

## Tests included

- [x] Passing? 

## Documentation
- [x] official documentation updated (https://github.com/SciCatProject/documentation/pull/69)


sofyalaski:
What I was thinking about while working on this PR, is that for configurations of no `#datasetX` type, we do no checks on dataset access for other configurations, even if the job concerns only datasets. 
To implement it properly, we would need to extend the list of possible configuration options, since we have different levels of dataset permissions. 
Instead, maybe we should configure the `validate` action for these operations? Also, I think it is worth pointing out this fact explicitly in the documentation that no dataset checks are done when these options of configuration are used. 

Also, when the job here is created by jobAdmin for one user and another group (user 2 + group3, for instance) we will be checking access for both group3 and `currentGroups` of user2. As I understand, groups get updated with the user logging in, so there might be an issue with the job if user hasn't logged in in a long time.

sbliven:
How does read access work with `auth: "#jobAdmin"`?

It would be possible to validate certain values of `ownerUser` or `ownerGroup` using a `validate` action. However it's not (currently) possible to validate against the actual logged in user, or to check group ownership.

sofyalaski:
> How does read access work with `auth: "#jobAdmin"`?

I added a get test for this. Generally, reading the job is never related to the configuration of the job, but to the dto on request. So as long as you provide correct recipient in either `ownerUser` or `ownerGroup`, it will be accessible to read.

src/jobs/jobs.controller.ts
```diff
@@ -500,7 +540,37 @@ export class JobsController {
...
+
+        if (
+          (!jobUser && jobInstance.ownerGroup) ||
+          (jobUser && user.username === jobUser.username) ||
+          (jobUser && jobUser.currentGroups.includes(jobInstance.ownerGroup))
+        ) {
+          datasetsWhere["where"]["ownerGroup"] = {
+            $eq: jobInstance.ownerGroup,
+          };
+        }
```
despadam: As I understand it, ownership in datasets is defined by groups and not users, meaning that if for example `user1` who belongs to `group1` creates a dataset then everyone in `group1` is essentially an owner. Is that correct?
If yes, then based on that I would say that this code makes sense. However the third condition I think is missing one use case: If `jobUser` exists (which happens when the logged-in user is an admin/jobAdmin and `ownerUser` was provided in the Dto and corresponds to a valid user) then it is not guaranteed that `jobInstance.ownerGroup` exists. `jobInstance.ownerGroup` will only get populated if such a value exists in the Dto but never from the `jobUser`'s group(s), I am assuming because a user can belong in multiple groups and here we would not know which one to choose from - I understand this is a tricky problem to solve and probably out of scope here. 

In any case I would suggest that `#datasetOwner` perhaps should be renamed to something that indicates we are searching based on groups and not users, otherwise this name might be misleading.

src/jobs/jobs.controller.ts
```diff
@@ -500,7 +540,37 @@ export class JobsController {
...
+                { ownerGroup: { $in: jobUser.currentGroups } },
+              ],
+            },
+          ];
+        } else if (jobUser && !jobInstance.ownerGroup) {
+          // job for user with no ownerGroup specified
+          datasetsWhere["where"]["ownerGroup"] = { $in: jobUser.currentGroups };
+        } else {
+          // job for anonymous user is always faulty, because job id cannot be empty
+          datasetsWhere["where"]["$or"] = [{ _id: { $in: [] } }];
```
despadam, line 578: Is this needed? Does it not get caught by the exception thrown above?

src/jobs/jobs.controller.ts
```diff
@@ -524,10 +594,13 @@ export class JobsController {
...
     const canCreate =
-      ability.can(Action.JobCreateAny, JobClass) ||
+      (ability.can(Action.JobCreateAny, JobClass) &&
+        user.currentGroups.includes("admin")) ||
+      (ability.can(Action.JobCreateAny, JobClass) && datasetsNoAccess == 0) ||
       ability.can(Action.JobCreateOwner, jobInstance) ||
       (ability.can(Action.JobCreateConfiguration, jobInstance) &&
-        datasetsNoAccess == 0);
+        datasetsNoAccess == 0 &&
+        jobConfiguration.create.auth != CreateJobAuth.JobAdmin);
```
despadam, line 609: I am not sure I understand this part. Why `jobConfiguration.create.auth =! CreateJobAuth.JobAdmin`?

And similarly for the update, same question.

despadam (commented):
Overall I think this is great work.

I also agree with the idea of performing additional checks using the validate action. It allows for more flexibility when defining what a jobAdmin can do, which is something that may vary from facility to facility.

src/jobs/jobs.controller.ts
sofyalaski: Because `Action.JobCreateConfiguration` (and similarly update) check permissions for non-elevated users. With this condition, I am saying not to check them at all if we are in the `#jobAdmin` configuration

src/jobs/jobs.controller.ts
sofyalaski: yes, Ownable class defines the ownership per group.
Thanks for pointing it out! I think this use case is written out in lines 567-569, however, it wasn't reachable and should be the first else if clause. I will change it. 

To get back to the discussion about inheriting from Ownable, in this use case, the job instance actually doesn't have an OwnerGroup at all (while it's not optional in OwnableClass )
 I am not sure renaming it would help, because `#datasetAccess` is also group-based. Maybe we should just write in the documentation that both these terms relate to group access and not user access, as this is not supported for datasets?

src/jobs/jobs.controller.ts
sofyalaski: I can extend the above exception like this to check that dto  is empty for `ownerUser/Group` and current user is not in ADMIN_GROUPS
```
if (
          !user ||
          (!user.currentGroups.some((g) =>
            this.accessGroups?.admin.includes(g),
          ) &&
            !jobCreateDto.ownerGroup &&
            !jobCreateDto.ownerUser)
        ) 
```

src/jobs/jobs.controller.ts
despadam: I agree to both: this if clause should be moved further up and also instead of renaming let's make the documentation more clear. I leave the discussion about OwnableClass for another time when we are ready to tackle https://github.com/SciCatProject/scicat-backend-next/issues/1775

src/jobs/jobs.controller.ts
despadam: Yes I think throwing an exception would be better in this case

despadam (dismissed):

src/jobs/jobs.controller.ts
```diff
@@ -481,17 +484,61 @@ export class JobsController {
...
+          ];
+        } else if (jobUser && !jobInstance.ownerGroup) {
+          // job for user with no ownerGroup specified
           datasetsWhere["where"]["$or"] = [
-            { ownerGroup: { $in: user.currentGroups } },
-            { accessGroups: { $in: user.currentGroups } },
+            { ownerGroup: { $in: jobUser.currentGroups } },
+            { accessGroups: { $in: jobUser.currentGroups } },
             { isPublished: true },
           ];
```
despadam: Actually, now that I am thinking about it shouldn't this one also be moved above? To be before ```
else if (
          jobUser &&
          !jobUser.currentGroups.includes(jobInstance.ownerGroup)
        )```

despadam (approved):

PR #1846, opened 2025-04-17
fix(security): don't log mongodb URI

sbliven:
## Description

The Mongodb URI can contain secrets and should not be logged.

## Fixes

* Remove potentially sensitive database information from the log

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -106,10 +106,6 @@ async function bootstrap() {
   }
 
   const port = configService.get<number>("port") ?? 3000;
-  Logger.log(
-    "MongoDB URI : " + configService.get<string>("mongodbUri"),
-    "Main",
-  );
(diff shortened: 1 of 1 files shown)


dylanmcreynolds (commented):
If the url isn't set, we would see that in the old code. What would you think of this?

```
const mongoUri = configService.get<string>("mongodbUri");
Logger.log(
  `MongoDB URI is ${mongoUri?.trim() ? "set" : "not set"}`,
  "Main",
);
```

Junjiequan (approved):

sbliven:
> If the url isn't set, we would see that in the old code. What would you think of this?

I just tested this, and with an empty `MONGODB_URI` you get a connection error before the main app even loads. So the log line wouldn't run, but you get a clear error `Unable to connect to the database` followed by ```MongooseError: The `uri` parameter to `openUri()` must be a string, got "undefined".```

PR #1853, opened 2025-04-23
test(attachmentV4): add api test for attachment v4

Junjiequan:
## Description
1. added API tests for attachments v4 service.
2. added isValid endpoint for attachment input validation
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/attachments/attachments.v4.controller.ts
+++ b/src/attachments/attachments.v4.controller.ts
@@ -13,6 +13,9 @@ import {
   ForbiddenException,
   NotFoundException,
   Patch,
+  Put,
+  ValidationError,
+  HttpCode,
 } from "@nestjs/common";
 import {
   ApiBearerAuth,
@@ -40,9 +43,14 @@ import { getSwaggerAttachmentFilterContent } from "./types/attachment-filter-con
 import { AttachmentFilterValidationPipe } from "./pipes/attachment-filter-validation.pipe";
 import { CreateAttachmentV4Dto } from "./dto/create-attachment.v4.dto";
 import { OutputAttachmentV4Dto } from "./dto/output-attachment.v4.dto";
-import { PartialUpdateAttachmentV4Dto } from "./dto/update-attachment.v4.dto";
+import {
+  PartialUpdateAttachmentV4Dto,
+  UpdateAttachmentV4Dto,
+} from "./dto/update-attachment.v4.dto";
 import { AttachmentsV4Service as AttachmentService } from "./attachments.v4.service";
 import { AllowAny } from "src/auth/decorators/allow-any.decorator";
+import { validate, ValidatorOptions } from "class-validator";
+import { plainToInstance } from "class-transformer";
 
 @ApiBearerAuth()
 @ApiTags("attachments v4")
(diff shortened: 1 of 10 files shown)


src/attachments/attachments.v4.controller.ts
```diff
@@ -383,6 +431,54 @@ export class AttachmentsV4Controller {
...
+    summary: "It validates the attachment provided as input.",
+    description:
+      "It validates the attachment provided as input, and returns true if the information is a valid attachment",
+  })
+  @ApiBody({
+    type: CreateAttachmentV4Dto,
+  })
+  @ApiResponse({
+    status: HttpStatus.OK,
+    type: Boolean,
```
martin-trajanovski: The response type here says `Boolean` but in the function you actually return something like: `{ valid: valid, reason: errorsAttachment }`

src/attachments/dto/update-attachment.v4.dto.ts
```diff
@@ -12,7 +18,9 @@ export class UpdateAttachmentV4Dto extends OwnableDto {
   readonly caption: string;
 
   @IsOptional()
-  readonly relationships?: AttachmentRelationshipClass[];
+  @ValidateNested({ each: true })
+  @Type(() => AttachmentRelationshipsV4Dto)
+  readonly relationships?: AttachmentRelationshipsV4Dto[];
```
martin-trajanovski, line 23: Why this needed a change from `AttachmentRelationshipClass` to `AttachmentRelationshipsV4Dto`? Do we still use the `AttachmentRelationshipClass` somewhere?

src/attachments/attachments.v4.controller.ts
Junjiequan: Goot catch

src/attachments/dto/update-attachment.v4.dto.ts
Junjiequan: The `@ValidateNested` decorator needs an actual object class to validate instead of a schema.
The schema class is only for Mongo and `ValidateNested` with `Type` combination does not work using the schema class

src/attachments/attachments.v4.controller.ts
```diff
@@ -350,6 +359,47 @@ export class AttachmentsV4Controller {
...
 
+  // PUT /attachments/:aid
+  @UseGuards(PoliciesGuard)
+  @CheckPolicies("attachments", (ability: AppAbility) =>
+    ability.can(Action.AttachmentUpdateEndpoint, Attachment),
+  )
+  @ApiOperation({
+    summary: "It updates the attachment.",
+    description:
+      "It updates the attachment specified through the id specified. it updates only the specified fields.",
```
nitrosx: I think this description is incorrect.
PUT replaces the whole object

nitrosx: If you you want to implement partial update, we should use PATCH

nitrosx (changes requested):

src/attachments/attachments.v4.controller.ts
Junjiequan: Yes you are right, the description is fixed now.

nitrosx (approved):

PR #1854, opened 2025-04-24
fix: migrate affected datasets where size numberOfFiles packedSize and numberOfFilesArchived were overwritten

martin-trajanovski:
## Description
This PR aims to provide a migration script for datasets properties that were overwritten by a bug that was found and fixed in the code.

## Motivation
`size`, `numberOfFiles`, `packedSize` and `numberOfFilesArchived` were all overwritten by a bug in the code and those needed to be fixed.

## Fixes

* https://jira.ess.eu/browse/SWAP-4629
* https://github.com/SciCatProject/scicat-backend-next/issues/1688

## Changes:

* Added new mongo migration script

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/migrations/20250423114619-populate-overwritten-datasets-size.js
+++ b/migrations/20250423114619-populate-overwritten-datasets-size.js
@@ -122,7 +122,7 @@ module.exports = {
 
     await db
       .collection("Dataset")
-      .find({ packedSize: 0 })
+      .find({ numberOfFilesArchived: 0 })
       .forEach(async (dataset) => {
         const datablocks = await db
           .collection("Datablock")
--- a/migrations/20250423114619-populate-overwritten-datasets-size.js
+++ b/migrations/20250423114619-populate-overwritten-datasets-size.js
@@ -1,0 +1,156 @@
+module.exports = {
+  async up(db, client) {
+    await db
+      .collection("Dataset")
+      .find({ size: 0 })
+      .forEach(async (dataset) => {
+        const origDatablocks = await db
+          .collection("OrigDatablock")
+          .find({
+            datasetId: dataset._id,
+          })
+          .toArray();
+        const origDatablockSizeSum = origDatablocks.reduce(
+          (acc, origDatablock) => acc + origDatablock.size,
+          0,
+        );
+
+        const datablocks = await db
+          .collection("Datablock")
+          .find({
+            datasetId: dataset._id,
+          })
+          .toArray();
+
+        const datablockSizeSum = datablocks.reduce(
+          (acc, datablock) => acc + datablock.size,
+          0,
+        );
+
+        const datasetSizeSum = origDatablockSizeSum + datablockSizeSum;
+
+        console.log(
+          `Updating Dataset (Id: ${dataset._id}) with size =>${datasetSizeSum}<=`,
+        );
+        await db.collection("Dataset").updateOne(
+          {
+            _id: dataset._id,
+          },
+          {
+            $set: {
+              size: datasetSizeSum,
+            },
+          },
+        );
+      });
+
+    await db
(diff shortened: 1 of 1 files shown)


migrations/20250423114619-populate-overwritten-datasets-size.js
```diff
@@ -0,0 +1,156 @@
...
+            datasetId: dataset._id,
+          })
+          .toArray();
+
+        const datablockSizeSum = datablocks.reduce(
+          (acc, datablock) => acc + datablock.size,
+          0,
+        );
+
+        const datasetSizeSum = origDatablockSizeSum + datablockSizeSum;
```
nitrosx, line 30: Are you sure that this is the way that it is calculated?
I think that the dataset total size is the sum of the size of the files listed in the `origDatablock`

migrations/20250423114619-populate-overwritten-datasets-size.js
```diff
@@ -0,0 +1,156 @@
...
+          })
+          .toArray();
+
+        const datablockNumberOfFilesSum = datablocks.reduce(
+          (acc, datablock) => acc + datablock.dataFileList.length,
+          0,
+        );
+
+        const datasetNumberOfFilesSum =
+          origDatablockNumberOfFilesSum + datablockNumberOfFilesSum;
```
nitrosx, line 75: This should include only the origdatablock number of files

migrations/20250423114619-populate-overwritten-datasets-size.js
```diff
@@ -0,0 +1,156 @@
...
+            $set: {
+              packedSize: datablockPackedSizeSum,
+            },
+          },
+        );
+      });
+
+    await db
+      .collection("Dataset")
+      .find({ packedSize: 0 })
```
nitrosx: shouldn't be `numberOfFilesArchived`?

nitrosx (changes requested):

migrations/20250423114619-populate-overwritten-datasets-size.js
martin-trajanovski: If you see this line for example on the creation of origDatablock: https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/datasets.controller.ts#L1945
Or this line in the update: https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/datasets.controller.ts#L2101
We can see that the size it is used and not the sum of the files listed.

migrations/20250423114619-populate-overwritten-datasets-size.js
martin-trajanovski: If we check the datasets controller you can see that even in the createDatablock endpoint we have the numberOfFiles updated. https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/datasets.controller.ts#L2230

Junjiequan (approved):
looks fine to me

martin-trajanovski:
@nitrosx let me know what are your final thoughts about this one. As I know you left some comments before.

migrations/20250423114619-populate-overwritten-datasets-size.js
nitrosx: OK. You are correct.
Let's resolve this conversation, and bring the discussion to the collaborators

migrations/20250423114619-populate-overwritten-datasets-size.js
nitrosx: Same as above, let's resolve this conversation and bring the discussion to the collaborators

nitrosx (approved):

PR #1855, opened 2025-04-24
feat: uniform handlebar variables for job actions

sbliven:
## Description

Many job actions accept handlebars templates. This exposes additional variables to the templates and makes accessing variables consistent across all the different scopes that actions are executed in.

All templates now receive the following context:
```ts
{
  request: JobDto,
  job?: JobClass,
  datasets?: DatasetClass[],
  env: Record<string,string?>
}
```

Before, `{{this}}` would correspond to different objects depending on how the action is
called; either a DTO or a Job or even a Dataset. Now these are accessed
consistently through top-level fields; thus `{{id}}` becomes `{{job.id}}`,
`{{jobParams}}` becomes `{{request.jobParams}}`, etc.

This also adds access to environmental variables via `{{env}}`, making
it easier to include secrets actions without hard-coding them.

Tests and example files should be up-to-date.

## Motivation

Adding env makes it easier to reference secrets securely without putting them in jobConfig.yaml.

Having a shared context among all jobs is also convenient for caching (eg of the datasets)

## Changes:

* Add a new top-level variable to all job templates

## Tests included

- [x] Passing? 

## Documentation

https://github.com/SciCatProject/documentation/pull/70


despadam (approved):
I like this a lot 👍

Issue #1798, opened 2025-03-27
Jobs: Make RabbitMQ disabled by default


despadam:
## Summary
The RabbitMQ action requires the following env variables to be configured correctly
```
RABBITMQ_ENABLED
RABBITMQ_HOSTNAME
RABBITMQ_PORT
RABBITMQ_USERNAME
RABBITMQ_PASSWORD
```

## Current Behavior
The variable `RABBITMQ_ENABLED` is optional.

If `RABBITMQ_ENABLED` is not provided then we currently assume that RabbitMQ should be enabled and proceed with configuring the action. This will result in an error, in the case that some of the other env variables are not provided. See code:
```
    if (!hostname || !port || !username || !password) {
      Logger.error(
        "RabbitMQ is enabled but missing one or more config variables.",
        "RabbitMQService",
      );
    } 
```

## Expected Behavior
Instead, if `RABBITMQ_ENABLED` is not provided then we should assume that RabbitMQ is disabled and we should not proceed with configuring the action.

PR #1856, opened 2025-04-28
fix: ensure RabbitMQ job action is disabled by default

despadam:
## Description

When the env variable `RABBITMQ_ENABLED` is not set to `"yes"`, the `RabbitMQModule` should be disabled. However, it was noticed that the module was still getting enabled and would result to throwing an error in case the other rabbitMQ env variables were not set.

It was confirmed that the module is set to be disabled by default, because it is defined as a `ConditionalModule` in `app.module` that is only going to be enabled when `RABBITMQ_ENABLED === "yes"`.  Additionally, the rule in `configuration.ts` sets `RABBITMQ_ENABLED` to `"no"` in case it is not defined.

The issue is caused by the way that `RabbitMQJobActionModule` is imported in `corejobactioncreators.module`. `RabbitMQJobActionModule` is not disabled by default, and depends on `RabbitMQModule`. This means that the job action module gets enabled regardless of the `RABBITMQ_ENABLED` value. As a result it also enables the `RabbitMQModule` which would then throw the error.

## Motivation

https://github.com/SciCatProject/scicat-backend-next/issues/1798

## Fixes

- Makes `RabbitMQJobActionModule` optional in `corejobactioncreators.module`

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/config/job-config/actions/corejobactioncreators.module.ts
+++ b/src/config/job-config/actions/corejobactioncreators.module.ts
@@ -1,4 +1,5 @@
 import { Module } from "@nestjs/common";
+import { ConditionalModule } from "@nestjs/config";
 import { LogJobActionCreator } from "./logaction/logaction.service";
 import { LogJobActionModule } from "./logaction/logaction.module";
 import { EmailJobActionCreator } from "./emailaction/emailaction.service";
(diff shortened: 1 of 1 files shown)


Junjiequan (approved):

Junjiequan (approved):
Looks good to me

PR #1857, opened 2025-04-28
fix: revert nested query object to individual parameters for Angular SDK compatibility

Junjiequan:
## Description
The generated Angular SDK cannot correctly handle nested query objects via @ApiQuery, resulting in malformed requests. This change reverts the composite query parameter introduced in PR #1490 back to separate top-level parameters to ensure proper SDK generation and request serialization.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -82,8 +82,12 @@ import {
 } from "./dto/update-derived-dataset-obsolete.dto";
 import { CreateDatasetDatablockDto } from "src/datablocks/dto/create-dataset-datablock";
 import {
+  datasetsFullQueryDescriptionFields,
+  datasetsFullQueryExampleFields,
   filterDescription,
   filterExample,
+  fullQueryDescriptionLimits,
+  fullQueryExampleLimits,
   replaceLikeOperator,
 } from "src/common/utils";
 import { HistoryClass } from "./schemas/history.schema";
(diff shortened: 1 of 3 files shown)


src/datasets/datasets.controller.ts
```diff
@@ -964,12 +968,22 @@ export class DatasetsController {
...
-    description: "Defines query limits and fields",
+    name: "fields",
+    description:
+      "Database filters to apply when retrieving datasets\n" +
+      datasetsFullQueryDescriptionFields,
     required: false,
-    type: FullQueryFilters,
-    example:
-      '{"limits": {"limit": 1, "skip": 1, "order": "creationTime:desc"}, fields: {"ownerGroup":["group1"],"scientific":[{"lhs":"sample","relation":"EQUAL_TO_STRING","rhs":"my sample"}]}}',
+    type: String,
```
martin-trajanovski, line 976: I can guess that the types were a bit problematic. But I don't understand exactly why when it worked until today. Is it related with the angular upgrade on the frontend or there is some other issue?

martin-trajanovski (approved):
I had just a small comment but otherwise it looks good.

src/datasets/datasets.controller.ts
Junjiequan: I'm not sure about it either. I encountered some errors when I generated the SDK locally, even before the Angular upgrade, such as errors from suffix and conflict from endpoint params etc.

PR #1866, opened 2025-04-29
fix: filters and limits in GET/jobs to get consistent results

despadam:
## Description

GET/jobs and fullQuery/jobs need to apply CASL rules after retrieving the jobs because these rules depend on each job configuration. This was causing inconsistencies in the results of endpoints because the filters and limits of the search where applied first, and the CASL rules afterwards. These rules would exclude jobs that the user does not have access to, causing discrepancies if the results where being paginated.

For example, when applying limits like so: `{ "where": { "type": "archive" }, "limits": { "skip": 0, "limit": 10 } }` one would expect the first page to contain 10 jobs, and similarly the rest of the pages, until we reach the end of the list. However, since the execution order was `1. where + limits, 2. CASL rules`, there were cases where a page would contain less than 10 jobs even if the remaining number of jobs was more than that.

## Motivation

Return consistent number of jobs per page, when pagination is applied.

## Changes:

Instead of using `findAll` and `findquery` from `jobs.service`, introduce `findByFilters` and `applyFilterLimits`, so that they can be applied separarately and the execution order can change to `1. where, 2. CASL rules, 3. limits`. This way the pagination will be done after all filtering has been applied. Now GET/jobs, fullQuery/jobs as well as fullFacet/jobs have been updated to always return consistent results.

## Tests included

- [ ] Passing? 

## Documentation


sofyalaski (approved):
Looks good to me! The only thing maybe is that we should also provide an example for SwaggerUI similar as we do in all other methods?

PR #1872, opened 2025-05-05
fix: send rabbitmq message to exchange and refactor

minottic:
## Description

Send message to channel rather than to queue, which bypassed the bound exchange. Class refactored to use async await.

## Motivation

Using the `channel.publish` method allows sending to the specified exchange and persists the message to the bound queue.

## Fixes

* allows sending to exchange set in config
* `channel.publish` instead of `channel.sendToQueue`
* uses async await and onModuleInit to make sure rabbitmq is set before sending message

## Changes:

* tests added
* try catch decorator for reusability
* config parser helper based on const list

## Tests included

- [x] Passing? 

## Documentation


sbliven (approved):
This was a bit hard to review because of all the refactoring. If I read it right the new features are
- Single channel for all messages which stays open throughout the nest lifetime
- Use `publish` instead of `sendToQueue`, which allows rabbitMQ to do routing.

The OnError decorator is a bit weird. You never use the "log" feature, and passing an error prefix via `_error` doesn't seem typesafe. I would have done error handling differently, but I guess this works.

minottic:
essentially the only change is the first commit, the rest is refactoring. The decorator param (log) was necessary before #1886 and I was lazy and did not change it after that was merged (which also meant the BE started fine if no connection could be established).

The channel stayed open even before since it was created in the constructor, but lifecycle hooks are used here to make sure stuff is initialized before being used.

despadam (commented):
Indeed, why have the "log" option in `onError` if it's never used? Apart from that I like the refactoring, looks good.

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -1,112 +1,135 @@
...
-        `Could not connect to RabbitMQ queue ${queue} with exchange ${exchange} and key ${key}.`,
-        { cause: error },
-      );
-    }
+  @OnError()
+  private async connect(): Promise<void> {
+    this._error = "Cannot connect to RabbitMQ";
+    this.connection = await amqplibConnect(this.connectionOptions);
+    this._error = "Channel error in RabbitMQService: ";
+    this.channel = await this.connection.createChannel();
```
Junjiequan, line 98: In my view, the `OnError` decorator brings unnecessary complexities. I personally would use `try catch` or just `.catch()` here  to make it easy to read. But the idea here is to make it DRY I guess.

src/common/rabbitmq/rabbitmq.service.ts
```diff
@@ -1,112 +1,135 @@
...
-    const hostname = this.configService.get<string>("rabbitMq.hostname");
-    const port = this.configService.get<number>("rabbitMq.port");
-    const username = this.configService.get<string>("rabbitMq.username");
-    const password = this.configService.get<string>("rabbitMq.password");
+  constructor(private readonly configService: ConfigService) {}
 
-    if (!hostname || !port || !username || !password) {
-      Logger.error(
-        "RabbitMQ is enabled but missing one or more config variables.",
-        "RabbitMQService",
```
Junjiequan, line 30: I prefer the old code, as you can see the required fields immediately, but I'm not against having it programmatically either

PR #1884, opened 2025-05-05
feat: add mongo session store

minottic:
## Description

Built based on #1804, it refactors express session to use nestjs [middlewars](https://docs.nestjs.com/middleware) and DI

## Motivation

Allows stateful sessions in case of multiple replicas deployment. Mongo store is configurable, and if left empty, it falls back to Memorystore. 

## Changes:

* refactor app.use to middleware
* DI configService
* DI mongoConnection
* configurable mongo Store
* reuse `session` DB from default
* uses JWT expiration as session expiration
* it restricts middleware to /oidc endpoints

## Tests included

- [x] Passing? 

## Documentation


minottic:
superseeds #1803

belfhi:
ah this looks great! I haven't had much time to look into my MR, so I'm happy my suggestion was taken up :)

nitrosx (approved):

src/auth/middlewares/session.middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+      store = MongoStore.create({
+        client: this.mongoConnection.getClient(),
+        ttl: this.configService.get<number>("jwt.expiresIn"),
+      });
+    this.requestHandler = session({
+      secret: this.configService.get<string>("expressSession.secret") as string,
+      resave: false,
+      saveUninitialized: true,
+      ...store,
+    });
```
Junjiequan: I tested it locally, and the session method does not register the mongoStore.
It appears to me that session function cannot handle `...store`
The correct type should be `store?: Store | undefined;`, you can check the type from the package interface.

```suggestion
    let store: Store | undefined;
    if (this.configService.get<string>("expressSession.store") === "mongo")
      store = MongoStore.create({
        client: this.mongoConnection.getClient(),
        ttl: this.configService.get<number>("jwt.expiresIn"),
      });
    this.requestHandler = session({
      secret: this.configService.get<string>("expressSession.secret") as string,
      resave: false,
      saveUninitialized: true,
      store,
    });
```

Junjiequan (changes requested):

src/auth/middlewares/session.middleware.ts
minottic: thanks for noticing this, I made a mistake indeed in the latest commit. I'll fix in a min PR and tests!

Junjiequan: The test looks fine to me. Maybe one additional API test to mock the process and see if user session is created in the db?

minottic: the test was also wrong, see the new commit. This controller test is a good idea I think, I did not add it before because all API tests are done with the /test folder from where I cannot check the mongo session. But I can add one indeed in the auth.controller.spec.ts

Junjiequan: You can run `db.collection("sessions").find({})` under the /test for checking I think, it's been done in many API tests. I think having a spec.ts is also fine, but maybe take more efforts

minottic: ah nice, thanks for pointing out I did not see that! Will do

minottic: done with jest as it required mocking the OidcGuard

Junjiequan (approved):
Looks good to me

PR #1887, opened 2025-05-06
feat: implement user password update functionality with validation

Junjiequan:
## Description
Add endpoints for users to change their own password and for admins to change other users’ passwords, with input validation and auth-strategy checks.

## Motivation
Users need a secure way to rotate their password; admins need to manage local-only accounts. Enforces that only “local” users can have their passwords updated.

## Changes:

- Added UpdateUserPasswordDto and AdminUpdateUserPasswordDto DTOs
- Implemented updateOwnPassword (POST /users/password) and updateUserPassword (PATCH /users/:id/password) in UsersController
- Validate current password, enforce authStrategy === "local", and return standardized response
- Extended UsersService to perform password hash/update logic

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/users/dto/update-user-password.dto.ts
+++ b/src/users/dto/update-user-password.dto.ts
@@ -1,0 +1,27 @@
+import { IsNotEmpty, IsString } from "class-validator";
+
+export class UpdateUserPasswordDto {
+  /**
+   * The current password of the user.
+   */
+
+  @IsString()
+  @IsNotEmpty()
+  currentPassword: string;
+
+  /**
+   * The new password of the user.
+   */
+  @IsString()
+  @IsNotEmpty()
+  newPassword: string;
+}
+
+export class AdminUpdateUserPasswordDto {
+  /**
+   * The new password of the user
+   */
+  @IsString()
+  @IsNotEmpty()
+  newPassword: string;
+}
--- a/src/users/dto/update-user-password.dto.ts
+++ b/src/users/dto/update-user-password.dto.ts
@@ -15,6 +15,13 @@ export class UpdateUserPasswordDto {
   @IsString()
   @IsNotEmpty()
   newPassword: string;
+
+  /**
+   * Confirmation of the new password. This should match the new password.
+   */
+  @IsString()
+  @IsNotEmpty()
+  confirmPassword: string;
 }
 
 export class AdminUpdateUserPasswordDto {
@@ -24,4 +31,11 @@ export class AdminUpdateUserPasswordDto {
   @IsString()
   @IsNotEmpty()
   newPassword: string;
+
+  /**
+   * Confirmation of the new password. This should match the new password.
+   */
+  @IsString()
+  @IsNotEmpty()
+  confirmPassword: string;
 }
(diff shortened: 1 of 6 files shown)


src/users/dto/update-user-password.dto.ts
```diff
@@ -0,0 +1,27 @@
...
+  @IsNotEmpty()
+  currentPassword: string;
+
+  /**
+   * The new password of the user.
+   */
+  @IsString()
+  @IsNotEmpty()
+  newPassword: string;
+}
```
nitrosx, line 25: Do we request the new password only one time?
Should we ask it twice, to be on the safe side... or are you thinking to do that in the FE?

src/users/users.controller.ts
```diff
@@ -189,14 +245,70 @@ export class UsersController {
...
+    }
+
+    await this.usersService.updateUserPassword(
+      updateUserPasswordDto.newPassword,
+      id,
+    );
+    return {
+      message: `Password updated successfully for userId: ${id}`,
+    };
+  }
```
nitrosx: Now that I'm reviewing this, I am wondering if it would be better to just reset the password with a random sequence instead of allowing admin to provide the password?

nitrosx (commented):
Just a couple of question/comments

src/users/users.controller.ts
Junjiequan: I think it's better to leave the choice for users? so they can choose a very complicated or easy to remember  password as prefered

src/users/dto/update-user-password.dto.ts
Junjiequan: That's a good point. 
It's probably better to have the password consistency validation in both BE and FE

test/Users.js
```diff
@@ -95,3 +97,169 @@ describe("2360: Users settings", () => {
...
+      });
+  });
+
+  it("0020: should fail when new password and confirmation password do not match", async () => {
+    return request(appUrl)
+      .post("/api/v3/users/password")
+      .send({
+        currentPassword: "wrongOldPassword",
+        newPassword: TestData.Accounts["user1"]["password"],
+        confirmPassword: "wrongConfirmPassword",
```
nitrosx: We should provide the correct password in the `currentPassword`, otherwise you do not know if it fails because of the wrong old password or the fact that the new password does not match

test/Users.js
```diff
@@ -95,3 +97,169 @@ describe("2360: Users settings", () => {
...
+    db.collection("User").updateOne(
+      { username: TestData.Accounts["user2"]["username"] },
+      { $set: { authStrategy: "oidc" } },
+    );
+  });
+
+  after(async () => {
+    db.collection("User").updateOne(
+      { username: TestData.Accounts["user2"]["username"] },
+      { $set: { authStrategy: "local" } },
```
nitrosx, line 132: Shouldn't `authStrategy` set to `oidc`?

nitrosx (changes requested):
Two minor changes on tests

test/Users.js
Junjiequan: authStrategy is set to oidc for validation testing. After the validation testing, the user data should revert to the original one to avoid any potential issues.

nitrosx (approved):

PR #1889, opened 2025-05-08
feat: make test env part of the test config

minottic:
## Motivation

This is to allow better tests reproducibility when running locally

## Changes:

* test/config/.env file added with env vars relevant for tests
* package.json scripts load them before execution
* start:test script to npm start loading test env
* possibility to override .env in file by supplying .env.override

## Tests included

- [ ] Passing? 

## Documentation


package.json
```diff
@@ -24,9 +24,10 @@
     "test:watch": "jest --watch --maxWorkers=25%",
     "test:cov": "jest --coverage --maxWorkers=50%",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:api": "npm run test:api:jest --maxWorkers=50% && concurrently -k -s first \"wait-on http://localhost:3000/explorer/ && npm run test:api:mocha\" \"npm run start\"",
-    "test:api:jest": "jest --config ./test/config/jest-e2e.json --maxWorkers=50%",
-    "test:api:mocha": "mocha --config ./test/config/.mocharc.json -r chai/register-should.js",
+    "start:test": "dotenv -o -e test/config/.env -e test/config/.env.override -- npm run start",
+    "test:api": "npm run test:api:jest --maxWorkers=50% && concurrently -k -s first \"wait-on http://localhost:3000/explorer/ && npm run test:api:mocha\" \"npm run start:test\"",
+    "test:api:jest": "dotenv -o -e test/config/.env -e test/config/.env.override -- jest --config ./test/config/jest-e2e.json --maxWorkers=50%",
+    "test:api:mocha": "dotenv -o -e test/config/.env -e test/config/.env.override -- mocha --config ./test/config/.mocharc.json -r chai/register-should.js",
```
Junjiequan, line 30: I assume the `.env.override` is expected to be mounted during image creation ?

minottic: it's optional. I added that because different systems might require it. For example, scicatlive will need it because mongo will not use localhost but the container name in the docker network. So scicatlive will mount it indeed, but if you run the tests locally you can also create it in your env and use it to override what required. If missing, the tests will still run as the CI proved

Junjiequan (approved):
It looks good to me. 
Could you document the changes, so that users know which env file they need to modify before running the tests.

minottic:
> It looks good to me. Could you document the changes, so that users know which env file they need to modify before running the tests.

done. Just a minor note to avoid confusion, users will not need to modify anything in the test/config/env file in most cases, the point of this PR is streamlining running the tests by removing the need to manually set ENV vars

Issue #954, opened 2023-12-11
add support for atomic patching of scientificMetadata sub-fields


jkotan:
### Summary

I would be nice to support atomic partial update of scientificMetadata. It would allowes to update sientificMetadata simultaneously by a few ingestor  instances i.e. a scan ingestor could update `scientificMetadata.instrument` while almost at the same time an user UI  ingestor could update `scientificMetadata.userData`   

### Current Behaviour

Currenly, you can patch the scinetificMetadata by fetching the current scienficMetadata an pushing back the update field which is not atomic operation (one transaction on DB) e.g.
```python                                                                                                                                                     
import requests
import json

baseurl = "http://localhost:3000/api/v3"
headers = {'Content-Type': 'application/json', 'Accept': 'application/json'}
pwd="aman"
response = requests.post(baseurl+ "/Users/login", headers=headers, json={"username":"ingestor", "password":pwd})
token = response.json()['id']

headers2 = dict(headers)
headers2["Authorization"] = "Bearer {}".format(token)
pid="testjk01/mysute_00019"
w = requests.get("{burl}/{url}/{pid}".format(burl=baseurl, url="Datasets",pid=pid.replace("/" ,"%2F")),params={"access_token":token},headers=headers2)
sc = w.json()["scientificMetadata"]

sc["instrument"]["current"] = {"value":99.2, "unit":"mA"}
jsc = json.dumps(sc)

requests.patch("{burl}/{url}/{pid}".format(burl=baseurl, url="Datasets",pid=pid.replace("/" ,"%2F")),params={"access_token":token},headers=headers2,data=jsc)
``` 
On the other hand if one try to update only  the `instrument current` in one call it of course removes all other scientificMetadata content, e.g.
```python

import requests
import json

baseurl = "http://localhost:3000/api/v3"
headers = {'Content-Type': 'application/json', 'Accept': 'application/json'}
pwd="aman"
response = requests.post(baseurl+ "/Users/login", headers=headers, json={"username":"ingestor", "password":pwd})
token = response.json()['id']

headers2 = dict(headers)
headers2["Authorization"] = "Bearer {}".format(token)
pid="testjk01/mysute_00019"

jsc = json.dumps({"scientificMetadata":{"instrument":{"current":{"value":99.1, "unit":"mA"}}}})

[...]

PR #1891, opened 2025-05-09
feat: mergepatch to support partial update of nested objects

sofyalaski:
## Description
As discussed in issue #954, this PR implements support for `application/merge-patch+json` type of content on requests to update endpoints. Support is added to the newer V4 endpoints for now. 

## Motivation
easier updates, especially of scientific metadata

## Changes:

* `application/merge-patch+json` content-type support
* `application/json` content-type remains
* in the controller function body it's checked what content-type user sends in the request, and if it's merge-patch+json, additional merge with the old document is applied
* changes are applied only to the new V4 endpoints for now
* for datasets, specific to ScientificMetadata and its interception, additional check is added that doesn't allow updating either value/unit without providing the second, when it was originally present in the dataset, to avoid user errors and keep functionality of the interceptor
* eslint applied to jobs test files

## Tests included

- [ ] Passing? 

## Documentation


src/attachments/attachments.v4.controller.ts
```diff
@@ -326,13 +328,14 @@ export class AttachmentsV4Controller {
   @ApiOperation({
     summary: "It updates the attachment.",
     description:
-      "It updates the attachment specified through the id specified. it updates only the specified fields.",
+      "It updates the attachment specified through the id specified. it updates only the specified fields. set to `application/merge-patch+json` if you would like to update nested objects.",
```
despadam:
```suggestion
      "It updates the attachment specified through the id specified. It updates only the specified fields. Set `content-type` to `application/merge-patch+json` if you would like to update nested objects.",
```

src/datasets/datasets.v4.controller.ts
```diff
@@ -674,6 +722,7 @@ export class DatasetsV4Controller {
     description: "Id of the dataset to modify",
     type: String,
   })
+  @ApiConsumes("application/merge-patch+json", "application/json")
```
despadam, line 725: Missing description: "Set `content-type` to `application/merge-patch+json` if you would like to update nested objects."

src/jobs/jobs.controller.ts
```diff
@@ -814,7 +825,8 @@ export class JobsController {
   @Version("3")
   @ApiOperation({
     summary: "It updates an existing job.",
-    description: "It updates an existing job.",
+    description:
+      "It updates an existing job. set to `application/merge-patch+json` if you would like to update nested objects.",
```
despadam:
```suggestion
      "It updates an existing job. Set `content-type` to `application/merge-patch+json` if you would like to update nested objects.",
```

src/jobs/jobs.controller.ts
```diff
@@ -814,7 +825,8 @@ export class JobsController {
   @Version("3")
```
despadam: Is this only for version 3?

test/DatasetV4.js
```diff
@@ -947,6 +965,225 @@ describe("2500: Datasets v4 tests", () => {
...
+            value: -2,
+            unit: "cm",
+            valueSI: -0.02,
+            unitSI: "m",
+          });
+          res.body.scientificMetadata.should.not.have.property("with_number");
+        });
+    });
+
+    it("0605: should not be able to partially update dataset's scientific metadata field when only value is passed without units, unless there were no units", () => {
```
despadam: You state "unless there were no units". How can we identify that?

test/DatasetV4.js
```diff
@@ -947,6 +965,225 @@ describe("2500: Datasets v4 tests", () => {
...
+          res.body.should.be.a("object");
+          res.body.should.have.property("pid");
+          res.body.should.have
+            .property("datasetName")
+            .and.be.eq("Updated dataset with scientific metadata");
+          res.body.scientificMetadata.with_unit_and_value_si.should.deep.eq({
+            value: -2,
+            unit: "cm",
+            valueSI: -0.02,
+            unitSI: "m",
```
despadam, line 1067: What if someone provides something like this, which is clearly wrong? 
```
            value: -2,
            unit: "cm",
            valueSI: 100,
            unitSI: "m",
```

despadam: Or something that contains `unitSI` but the wrong one? For example
```
            value: -2,
            unit: "cm",
            valueSI: -2,
            unitSI: "V",
```

despadam (commented):
I left some comments/questions. Once clarified I will review again.

The warnings about using `any` perhaps can be resolved if we use `unknown` instead.

test/DatasetV4.js
sofyalaski: I will write a test, but the interceptor itself takes care of that

src/jobs/jobs.controller.ts
sofyalaski: okay, that's my bad! Line 828 with `@ApiOperation` should've remained with no changes, as the option to choose `content-type` should be only available for v4 endpoints.

test/DatasetV4.js
```diff
@@ -1184,6 +1184,84 @@ describe("2500: Datasets v4 tests", () => {
...
+        .auth(accessTokenAdminIngestor, { type: "bearer" })
+        .expect(TestData.SuccessfulPatchStatusCode)
+        .expect("Content-Type", /json/)
+        .then((res) => {
+          res.body.should.be.a("object");
+          res.body.scientificMetadata.with_undefined_unit.should.have.property("value").and.be.eq(555);
+          res.body.scientificMetadata.with_undefined_unit.should.not.have.property("unit");
+        });
+    });
+    it("0610: should be able to partially update dataset's scientific metadata field when value, unit, valueSI, unitSI and overwrite the passed body with values convirted by interceptor", () => {
```
despadam: Please make description clearer.

test/DatasetV4.js
```diff
@@ -1184,6 +1184,84 @@ describe("2500: Datasets v4 tests", () => {
             );
         });
     });
+    it("0609: should be able to partially update dataset's scientific metadata field when only value if there never were units defined for it", () => {
```
despadam: Please make description clearer.

test/DatasetV4.js
```diff
@@ -1184,6 +1184,84 @@ describe("2500: Datasets v4 tests", () => {
...
+        .send(updatedDataset)
+        .auth(accessTokenAdminIngestor, { type: "bearer" })
+        .expect(TestData.SuccessfulPatchStatusCode)
+        .expect("Content-Type", /json/)
+        .then((res) => {
+          res.body.should.be.a("object");
+          res.body.scientificMetadata.should.have.property("with_no_unit").and.be.deep.eq(updatedDataset.scientificMetadata.with_no_unit);
+        });
+    });
+    it("0610: should be able to partially update dataset's scientific metadata field when only value if the units were undefined", () => {
```
despadam: Please make description clearer.

test/DatasetV4.js
```diff
@@ -1129,7 +1129,7 @@ describe("2500: Datasets v4 tests", () => {
             );
         });
     });
-    it("0607: should not be able to partially update dataset's scientific metadata field when only unit is passed without values, unless there were no units", () => {
+    it("0607: should not be able to partially update dataset's scientific metadata field when only value and valueSI passed, unless there were no units", () => {
```
despadam: Please make description clearer.

despadam (changes requested):

src/datasets/datasets.v4.controller.ts
```diff
@@ -233,6 +236,50 @@ export class DatasetsV4Controller {
     return filter;
   }
 
+  isRecord(obj: unknown): obj is Record<string, unknown> {
+    return typeof obj === "object" && obj !== null && !Array.isArray(obj);
+  }
+  isValueUnitObject(obj: unknown): obj is { value?: unknown; unit?: unknown } {
+    return this.isRecord(obj) && ("value" in obj || "unit" in obj);
+  }
+  findInvalidValueUnitUpdates(
```
despadam, line 243: Please also add some comments to this function to explain what is the logic

despadam (approved):

test/AttachmentV4.js
```diff
@@ -172,6 +172,46 @@ describe("Attachments v4 tests", () => {
...
+        .auth(accessTokenAdminIngestor, { type: "bearer" })
+        .expect(TestData.SuccessfulPatchStatusCode)
+        .expect("Content-Type", /json/)
+        .then((res) => {
+          res.body.should.be.a("object");
+          res.body.caption.should.equal("Updated caption text");
+          res.body.thumbnail.should.equal("Updated thumbnail URL");
+          res.body.relationships.should.deep.equal([
+            {
+              targetId: "testId1-modified-twice",
```
Junjiequan, line 203: Does the `jmp` also supports update for the specific item in an array? If so, can we add test for it as well?

Junjiequan (approved):
I've left one comment, otherwise looks good to me.
I assigned @martin-trajanovski to review also, as he developed datasets v4 controller he knows it better than me.

src/datasets/datasets.v4.controller.ts
```diff
@@ -233,6 +236,56 @@ export class DatasetsV4Controller {
     return filter;
   }
 
+  isRecord(obj: unknown): obj is Record<string, unknown> {
```
martin-trajanovski: Maybe it is better to move this function in the `utils.ts` file. And btw we already have a function called `isObject` there but I am not sure if we can reuse that one instead of introducing the `isRecord`.

sofyalaski:
Thank you Jay!

src/datasets/datasets.v4.controller.ts
```diff
@@ -233,6 +236,56 @@ export class DatasetsV4Controller {
...
+    // checks if the argument is object and containe either property 'value' or 'unit'
+    return this.isRecord(obj) && ("value" in obj || "unit" in obj);
+  }
+
+  findInvalidValueUnitUpdates(
+    updateDto: Record<string, unknown>,
+    dataset: Record<string, unknown>,
+    path: string[] = [],
+  ): string[] {
+    // collect properties that have both 'value' and 'unit' in original dataset but one of eother is missing in the updateDto body
```
martin-trajanovski: The comment here seems to have a small typo `eother`

src/datasets/datasets.v4.controller.ts
```diff
@@ -233,6 +236,56 @@ export class DatasetsV4Controller {
     return filter;
   }
 
+  isRecord(obj: unknown): obj is Record<string, unknown> {
+    // checks if value is (nested) object
+    return typeof obj === "object" && obj !== null && !Array.isArray(obj);
+  }
+  isValueUnitObject(obj: unknown): obj is { value?: unknown; unit?: unknown } {
+    // checks if the argument is object and containe either property 'value' or 'unit'
```
martin-trajanovski: Comment with typo `containe`

test/AttachmentV4.js
sofyalaski: As far as the [RFC 7396](https://www.rfc-editor.org/rfc/rfc7396) explains, no, instances other than objects aren't supported (chapter 2 right after pseudocode.). In fact, in this case, the target will be reset to an empty object first and then updated with whatever is in the patch request. I will add a line to the API docs, that users should be careful when updating arrays

Junjiequan: Thanks for the clarification. And yes, it would be nice to have an additional warning description for using this feature

sofyalaski:
Thank you @Martin (also for pointing out the utils function). The present `isObject` checks that the object does not contain the units, so it's the opposite of what I need. I put my function in the utils. Is it okay if the rest stay in the controller or should I move them too?

src/attachments/attachments.v4.controller.ts
```diff
@@ -328,7 +328,7 @@ export class AttachmentsV4Controller {
   @ApiOperation({
     summary: "It updates the attachment.",
     description:
-      "It updates the attachment specified through the id specified. It updates only the specified fields. Set `content-type` to `application/merge-patch+json` if you would like to update nested objects.",
+      "It updates the attachment specified through the id specified. It updates only the specified fields. Set `content-type` to `application/merge-patch+json` if you would like to update nested objects. Warning! `application/merge-patch+json` doesn't support updating elements of array!",
```
Junjiequan: Could we change the warning text to Warning! application/merge-patch+json doesn’t support updating a specific item in an array—the result will always replace the entire target if it’s not an object.

Junjiequan: So the description is a bit more specific

martin-trajanovski:
> Thank you @martin (also for pointing out the utils function). The present `isObject` checks that the object does not contain the units, so it's the opposite of what I need. I put my function in the utils. Is it okay if the rest stay in the controller or should I move them too?

I think you can move just those two that are more general ones like `isRecord` and `isValueUnitObject`. Other ones can stay as they are more specific to the dataset controller.

test/config/.env.override
```diff
@@ -0,0 +1,2 @@
+MONGODB_URI=mongodb://mongodb:27017/scicat-test-db
+ES_HOST=http://elastic:9200
```
Junjiequan: I think the `.env.override` is expected to be mounted instead of having it in the core

https://github.com/SciCatProject/scicat-backend-next/pull/1889

PR #1910, opened 2025-05-14
feat: add all config files to env of tests

minottic:
## Description

Builds on top of #1889 setting config files in env file

## Motivation

Allows for reproducing tests, removing the need to set config manually 

## Changes:

* config files are set in test/config/.env rather than copied in the test setup
* functionalAccounts file can be supplied as an ENV var
* functionalAccounts.json.test moved to test/config/functionalAccounts.json for consistency
* gitignore has a test section

## Tests included

- [ ] Passing? 

## Documentation


sbliven (approved):
lgtm

test/config/.env
```diff
@@ -18,6 +18,8 @@ SAMPLE_PRIVILEGED_GROUPS=sampleingestor
 SITE=PSI
 UPDATE_JOB_PRIVILEGED_GROUPS=group3
 RABBITMQ_ENABLED=no
+PROPOSAL_TYPES_FILE=proposalTypes.example.json
+FUNCTIONAL_ACCOUNTS_FILE=test/config/functionalAccounts.json
```
Junjiequan, line 22: it seems `DATASET_TYPES_FILE` is missing here

minottic: yeah but from what I could say it's not used

Junjiequan: I'm not sure if it's intentional or the test is missing. We can add it back later maybe then.

Junjiequan (approved):

Issue #1776, opened 2025-03-20
Issues related to release-jobs backwards compatibility


despadam:
## Summary

To ensure that the new jobs subsystem will be backwards compatible, this PR was created: https://github.com/SciCatProject/scicat-backend-next/pull/1742
During the review process, a few topics came up that still need to be discussed, to ensure there is no unexpected behavior.

### Topic 1:
For jobs created by admins for other users, we use `findByUsername2JWTUser()` in `jobs.controller` to retrieve the `ownerUser`. 

If the user is not found then the jobs gets automatically assigned to the admin and we show a relevant warning.
Should we throw an error instead of logging?

### Topic 2:
For jobs created by normal users, v4 does not allow creating a job with `ownerUser` but without an `ownerGroup`. In `jobs.controller` we check that a group is provided and that the user belongs to that group.

To make this compatible with v3, the solution is to use `findByUsername2JWTUser()` to retrieve the user's groups. Then we pick the last one from the returned list of groups. This means that when a user wants to create a job for himself in v3, the job will be automatically shared with his group (or one of his groups in case he belongs to multiple).

The other solution, instead of automatically adding an arbitrary group, would be assigning an empty string also to `ownerUser`, but then that user will not have access to the job he just created.
 
Would we prefer to:
1. keep sharing the job with a group, even though the user did not specify this?
2. follow a more restrictive approach and assign an empty string to `ownerUser`, meaning that access to the job would be restricted also for the user himself?
3. redesign the checks in `jobs.controller` to make it compatible? If yes, how?

sbliven:
The `POST /api/v3/jobs` endpoint is currently broken due to topic 2. All v3 jobs return 400 "Invalid new job. Owner group should be specified" due to the empty `ownerGroup` field.

Maybe we should just drop the `ownerGroup` field in line with #1775?

despadam:
Suggestions for **topic 2**:
- If `username` in `jobParams` of v3 Dto is omitted, use the currently logged in user
- Rather than "randomly" picking a group, what if we take the group of the first dataset? If the user isn't in it we would throw an error anyways, so it should be safer

sbliven:
Topic 2 is closed.

For topic 1, let's assign to the current user, not admin

despadam:
Turns out that in the meantime **Topic 1** has already been covered by previous PRs, and as a fallback rule the job gets assigned to the currently logged in user. See the relevant changes here:
[...]

PR #1913, opened 2025-05-19
fix: update ownerUser, ownerGroup logic for v3 job creation

despadam:
## Description

Updating `src/jobs/interceptors/create-job-v3-mapping.interceptor.ts` for improved backwards compatibility.

## Motivation

https://github.com/SciCatProject/scicat-backend-next/issues/1776 (topic 2)

## Changes:

- If `username` is not provided in `jobParams` (v3 dto) then we assign the jobs to the currently signed in user.
- If `ownerGroup` is not provided in `jobParams` (v3 dto) then, depending on the job auth configuration:
  - For `#datasetOwner` we retrieve the ownerGroup of the first dataset in `datasetList`, since for this job config all chosen datasets should have the same ownerGroup as the job.
  - For `#datasetAccess` we pick the first common group of both accessGroups and ownerGroups of all chosen datasets

## Tests included

- [x] Passing? 

## Documentation

https://github.com/SciCatProject/documentation/pull/74


src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
```diff
@@ -24,55 +28,112 @@ interface JobParams {
...
+        };
+      }
+      // ensure compatibility with v4 which requires ownerGroup in the dto of jobs created by normal user
+      if (Object.keys(jobParams).includes("ownerGroup")) {
+        newBody = {
+          ...newBody,
+          ownerGroup: jobParams.ownerGroup as string,
+        };
+      } else if (Array.isArray(jobParams.datasetList)) {
+        if (jobConfig.create.auth === "#datasetOwner") {
```
sbliven: There's no final `else` clause on here, so the group doesn't get set at all for other auth values. I think if it's not `#dataset*` it should pull the ownerGroup from the first dataset, if one is set.

src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
```diff
@@ -24,55 +28,112 @@ interface JobParams {
...
+              const dataset = await this.datasetsService.findOne({
+                where: { pid: datasetDto.pid },
+              });
+              datasetGroups.push([
+                ...(dataset?.accessGroups ?? []),
+                dataset?.ownerGroup,
+              ]);
+            }
+          }
+          const commonGroups = intersection(datasetGroups);
```
sbliven: Also need to intersect this with request.user.currentGroups

sbliven (changes requested):
I found some more edge cases that need to be fixed.

src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
despadam: done in 370d0450a2fdf2c693e38f73aaf11c336dd10c8f: for `#datasetAccess` we intersect `ownerGroup` and `accessGroups`, and for all other cases apart from `#datasetPublic` we use the `ownerGroup` of the first dataset

src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
despadam: done in 370d0450a2fdf2c693e38f73aaf11c336dd10c8f: also intersecting with `jobUser.currentGroups`, where `jobUser` is either the one provided by `jobParams.username` or the `request.user`

src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
```diff
@@ -24,55 +28,116 @@ interface JobParams {
...
+      request.body = newBody;
     }
-    request.body = newBody;
     return next.handle();
   }
 }
+
+function intersection<T>(arrays: T[][]): T[] {
+  if (arrays.length === 0) return [];
+  return arrays.reduce((a, b) => a.filter((x) => b.includes(x)));
```
sbliven: The performance on large numbers of groups would be better if you reduced over a Set (`Set.has` is O(1) compared with O(n) for `Array.includes`). However that's premature optimization; the current algorithm should be fine.

src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
```diff
@@ -24,55 +28,116 @@ interface JobParams {
...
+                ...(dataset?.accessGroups ?? []),
+                dataset?.ownerGroup,
+              ]);
+            }
+          }
+          const commonDatasetGroups = intersection(datasetGroups);
+          const commonGroups = intersection([
+            commonDatasetGroups,
+            jobUser?.currentGroups ?? [],
+          ]);
```
sbliven: This can be simplified into a single call
```suggestion
          const commonGroups = intersection([
            ...datasetGroups,
            jobUser?.currentGroups ?? [],
          ]);
```

sbliven (approved):
This looks good now. I did some manual testing and didn't see any issues.

PR #1914, opened 2025-05-19
feat: customizable default job.statusCode

sbliven:
## Description

The status code and message can be customized for new jobs using the `JOB_DEFAULT_STATUS_CODE` (default: "jobSubmitted" and `JOB_DEFAULT_STATUS_MESSAGE` (default: "Job submitted") variables.

## Motivation

The default was previously hard-coded to "jobCreated", which was used in v3 to indicate the job was being processed by the archiver. This changes the default to the backwards-compatible "jobSubmitted" value but allows it to be overwritten if needed.

## Tests included

- [x] Passing? 

## Documentation

https://github.com/SciCatProject/documentation/pull/73


despadam (approved):
Looks good! 
My only question is that in `jobs.controller` in `instanceAuthorizationJobCreate()` we set
```
    jobInstance.statusCode = "Initializing";
    jobInstance.statusMessage =
      "Building and validating job, verifying authorization";
```
at a first step, **before** setting the status and message to the default creation values. 
Is this something we want to change as well, or it's ok to keep it hardcoded there?

sbliven:
@despadam I wasn't aware of `instanceAuthorizationJobCreate`. I guess the value doesn't matter too much, since it's discarded after authorization. However, I will update the values to also use the env variable.

sofyalaski (approved):

Issue #1897, opened 2025-05-12
Add DatasetLifecycle endpoints


sbliven:
## Summary
Dataset.datasetlifecycle contains dataset properties relating to archive state that are intended to be managed by the archiveManager rather than the user. It would be nice to add stricter permissions for who can update these. One suggestion was to limit updates to a datasetLifecycle endpoint which could be limited to the archiveManager.

## Details

- Add endpoints for GET and PATCH `/api/v3/Datasets/{id}/datasetLifecycle`. This should be backwards compatible with the [backend-legacy](https://dacat.psi.ch/explorer/#!/Dataset/Dataset_prototype_get_datasetLifecycle) endpoints of the same name
- Remove datasetlifecycle from the create and update Dataset DTOs, forcing updates to go through the above endpoints
- Add a new env variable `UPDATE_DATASETLIFECYCLE_GROUPS` (spelling TBD). Only members of these groups (typically just the archiveManager user) can PATCH the datasetLifecycle endpoint.

One issue I see with this proposal is that users need to be able to set `datasetlifecycle.storageLocation` during create and update `datasetlifecycle.archivable`. We would need to come up with work-arounds for these operation.

PR #1928, opened 2025-05-21
feat: add dataset lifecycle endpoints

sofyalaski:
## Description
As described in the issue #1897, there were dedicated endpoints for datasetlifecycle [in v3](https://dacat.psi.ch/explorer/#!/Dataset/Dataset_prototype_get_datasetLifecycle). This PR introduces the GET and PATCH methods for dataset lifecycle. 

## Changes:

1. ~~`storageLocation` property of the `LifecycleClass` schema became part of `DatasetClass`.~~
2. ~~datasetlifecycle is created automatically with the dataset based on the defaults described in the LifecycleSchema. Trying to provide a lifecycle while creating a dataset will throw an error. Previous use cases, where datasetlifecycle was defined with dataset creation, must use PATCH method right after dataset creation.~~
3. ~~trying to update dataset lifecycle through PATCH `/dataset/:id` will throw an error and message to use dedicated endpoint.~~
4. `/api/v3` and `/api/v4` are supported.
5. GET method to access datasetlifecycle is implemented.
6. PATCH method to update datasetlifecycle is implemented. It's implemented with `content-type` `application/merge-patch+json` in `/api/v4/`. Whoever updates the lifecycle needs only to provide the values that need to be updated. Setting a value to `null` in `jmp` means resetting it to undefined, which is then reset to the defaults from the schema in the service. ( I was unsure if `conten-type` `application/json` should be supported as well, as the behavior will not change and `jmp` will be applied in any case - I can add it as needed). In `/api/v3/` just the `application/json' is supported.
7. A new role `UPDATE_DATASET_LIFECYCLE_GROUPS` is added to update the lifecycle. As the discussion seems to be ongoing, who should be able to update the datasetlifecycle (i.e at PSI it's part of an existing workflow that users themselves set the `archivable` status), the rules for the new role are defined on top of the existing.
8. I extended explanation of `@ApiOperation` decorators that use `jmp` in `/api/v4/` to include the difference in behaviour when setting a value to `null` based on the `content-type` that is missing in earlier [PR merge](https://github.com/SciCatProject/scicat-backend-next/pull/1891). 

## Tests included

- [x] Passing? 
## Documentation


sofyalaski:
Would facilities prefer to keep an option to provide a custom `datasetlifecycle` when creating a dataset (point 2) - as it was in the V3?

sbliven:
I think datasetlifecycle properties should still be allowed when creating datasets. The ingestor needs to set most of the properties. Also, the more I think about it the more it seems like owners should be able to update these fields too. For instance, shouldn't users be able to update the publication date? @minottic What's your opinion on this?

Can `UPDATE_DATASET_LIFECYCLE_GROUPS` still do updates via `PATCH /dataset/:id` or does it have to go via  `PATCH /dataset/:id/datasetLifecycle`?

sofyalaski:
> Can `UPDATE_DATASET_LIFECYCLE_GROUPS` still do updates via `PATCH /dataset/:id` or does it have to go via `PATCH /dataset/:id/datasetLifecycle`?

no, no one can update dataset lifecycle through the  `PATCH /dataset/:id`

nitrosx:
@sbliven letting owners to update/modify dataset life cycle is dependent from the data policy of the institution.
I can see two use cases: 
1) a facility that let's dataset owner to decide everything about the dataset life cycle, as long as the requested values are there
2) a facility that wants to complete control the dataset life cycle and the user does not have any direct control and any update is centrally controlled

src/datasets/dto/update-dataset.dto.ts
```diff
@@ -458,3 +458,149 @@ export class UpdateDatasetWithHistoryDto extends UpdateDatasetDto {
...
+
+  @ApiProperty({
+    type: Boolean,
+    required: false,
+    description:
+      "Flag indicating if full dataset is available on central fileserver. If false, data needs to be copied from decentral storage places to a cache server before the ingest. This information needs to be transferred to the archive system at archive time.",
+  })
+  @IsOptional()
+  @IsBoolean()
+  readonly isOnCentralDisk?: boolean;
```
nitrosx, line 549: Where do we store the location of the decentralized copy/copies?

src/datasets/dto/update-dataset.dto.ts
```diff
@@ -458,3 +458,149 @@ export class UpdateDatasetWithHistoryDto extends UpdateDatasetDto {
...
+  readonly dateOfPublishing?: Date;
+
+  @ApiProperty({
+    type: Date,
+    required: false,
+    description: "Day when dataset was published.",
+  })
+  @IsOptional()
+  @IsDateString()
+  readonly publishedOn?: Date;
```
nitrosx, line 539: Should change this to `dateWhenPublished`?

src/datasets/dto/update-dataset.dto.ts
nitrosx, line 1: Should we review naming convention and come up with a consistent one?
Also, how do we deal if a facility has a customized set of fields? Do we allow and how do we manage it?

src/datasets/schemas/dataset.schema.ts
```diff
@@ -285,6 +285,15 @@ export class DatasetClass extends OwnableClass {
...
 
+  @ApiProperty({
+    type: String,
+    required: false,
+    description:
+      "Location of the data on the archive system. Allowed values are facility-specific and may be validated when submitting scicat jobs relating to this dataset. Facilities with a single storage location can leave this field empty.",
+  })
+  @Prop({ type: String, required: false })
+  storageLocation?: string;
+
```
nitrosx: I'm not sure that I like this field here, but I do not have a better solution to offer.

martin-trajanovski: I am also thinking if this field has something in common with the `creationLocation` field on the dataset schema or it is completely different purpose.

sofyalaski: PSI and OpenEM introduced this field. The idea is to designate different archival storage destinations, so it has different purpose than `creationLocation`.

src/datasets/dto/update-dataset.dto.ts
sofyalaski: In the `LifecycleClass`schema, there are two objects of `unknown` type: `retrieveReturnMessage` and `archiveReturnMessage`. My assumption, these fields are designed to be used by facility services to pass around customized fields.

sofyalaski: I feel like there's a need to revisit the whole schema, following your messages

sofyalaski:
I ended up implementing folllowing rules, different from the initial PR: 

- users that are owners of the dataset can both create and update the dataset lifecycle from the `POST /dataset` and `PATCH /dataset/:id` endpoints.  They can also patch with the new `PATCH /dataset/:id/datasetlifecycle` endpoints. 
- users from UPDATE_DATASET_LIFECYCLE_GROUPS can `PATCH /dataset/:id/datasetlifecycle` any user's datasetlifecycle, but have no access to `PATCH /dataset/:id` on the endpoint level. For that, I introduced a new action in casl `DatasetLifecycleUpdate`. 
- `storageLocation` is still part of `Lifecycle Class` schema

src/datasets/dto/update-dataset.dto.ts
nitrosx: I agree with you. I think we should brainstorm about how to improve the dataset life cycle management and structure

nitrosx: Let's table it for now and see start the discussion after er merge this PR

src/datasets/dto/update-dataset.dto.ts
nitrosx: Let's postpone this decision during future brainstorming sessions

src/datasets/dto/update-dataset.dto.ts
nitrosx: Same as above. LEt's table it for now and discuss it at a later stage

nitrosx (approved):

src/datasets/schemas/dataset.schema.ts
sbliven: @martin-trajanovski creationLocation is the name of the beamline/instrument where the data was collected

src/attachments/attachments.v4.controller.ts
```diff
@@ -326,8 +326,13 @@ export class AttachmentsV4Controller {
   )
   @ApiOperation({
     summary: "It updates the attachment.",
-    description:
-      "It updates the attachment specified through the id specified. It updates only the specified fields. Set `content-type` to `application/merge-patch+json` if you would like to update nested objects. Warning! `application/merge-patch+json` doesn’t support updating a specific item in an array — the result will always replace the entire target if it’s not an object.",
+    description: `It updates the dataset through the pid specified. It updates only the specified fields.
```
despadam: Updates dataset or attachment?

src/jobs/jobs.controller.ts
```diff
@@ -877,8 +877,13 @@ export class JobsController {
   @Version("4")
   @ApiOperation({
     summary: "It updates an existing job.",
-    description:
-      "It updates an existing job. Set `content-type` to `application/merge-patch+json` if you would like to update nested objects. Warning! `application/merge-patch+json` doesn’t support updating a specific item in an array — the result will always replace the entire target if it’s not an object.",
+    description: `It updates the dataset through the pid specified. It updates only the specified fields.
```
despadam: updates job

despadam (dismissed):

despadam (approved):
Looks good, thank you! 👍

PR #1930, opened 2025-05-21
feat: review and refactor published data schema and add data migration script

martin-trajanovski:
## Description
This PR aims to refactor the published data schema according to the new standard and latest review.

## Motivation
The plan is to get all the changes and refactors of the published data into one branch `published-data-refactor` and after that open a PR towards the `master` branch including all the needed changes for the first working version after the refactor.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/published-data/interfaces/published-data.interface.ts
+++ b/src/published-data/interfaces/published-data.interface.ts
@@ -45,4 +45,5 @@ export enum PublishedDataStatus {
   PRIVATE = "private",
   PUBLIC = "public",
   REGISTERED = "registered",
+  AMENDED = "amended",
 }
--- a/src/published-data/interfaces/published-data.interface.ts
+++ b/src/published-data/interfaces/published-data.interface.ts
@@ -40,3 +40,9 @@ export class FormPopulateData {
 export interface IRegister {
   doi: string;
 }
+
+export enum PublishedDataStatus {
+  PRIVATE = "private",
+  PUBLIC = "public",
+  REGISTERED = "registered",
+}
--- a/src/published-data/published-data.controller.ts
+++ b/src/published-data/published-data.controller.ts
@@ -509,13 +509,11 @@ function formRegistrationXML(publishedData: PublishedData): string {
     title,
     abstract,
     resourceType,
-    creator,
+    creators,
     metadata,
   } = publishedData;
   const doi = publishedData.doi;
-  const uniqueCreator = creator.filter(
-    (author, i) => creator.indexOf(author) === i,
-  );
+  const uniqueCreator = creators.filter((c, i) => creators.indexOf(c) === i);
 
   const creatorElements = uniqueCreator.map((author) => {
     const names = author.split(" ");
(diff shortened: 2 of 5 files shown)


src/published-data/interfaces/published-data.interface.ts
```diff
@@ -40,3 +40,9 @@ export class FormPopulateData {
 export interface IRegister {
   doi: string;
 }
+
+export enum PublishedDataStatus {
+  PRIVATE = "private",
+  PUBLIC = "public",
+  REGISTERED = "registered",
```
nitrosx, line 47: We should add AMENDED.

src/published-data/published-data.controller.ts
```diff
@@ -503,23 +504,22 @@ export class PublishedDataController {
...
 function formRegistrationXML(publishedData: PublishedData): string {
   const {
-    affiliation,
     publisher,
     publicationYear,
     title,
     abstract,
     resourceType,
-    creator,
+    creators,
```
nitrosx: Didn't we decided to move creators to metadata as we are not sure which formats the facility would like to use?

src/published-data/schemas/published-data.schema.ts
```diff
@@ -12,7 +14,12 @@ export type PublishedDataDocument = PublishedData & Document;
   },
   timestamps: true,
 })
-export class PublishedData {
+export class PublishedData extends QueryableClass {
+  @Prop({
+    type: String,
+  })
+  _id: string;
+
```
nitrosx, line 22: Will `_id` be the same as `pid`?

src/published-data/schemas/published-data.schema.ts
```diff
@@ -49,56 +56,46 @@ export class PublishedData {
...
-      "Creator of dataset/dataset collection.  This field has the semantics" +
+      "The main researchers involved in producing the data, or the authors of the publication, in priority order. This field has the semantics" +
       " of Dublin Core [dcmi:creator](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/terms/creator/)" +
-      " and [DataCite Creator/creatorName](https://datacite-metadata-schema.readthedocs.io/en/4.5/properties/creator/#creatorname).",
+      " and [DataCite Creator/creatorName](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/creator/#creatorname).",
   })
-  @Prop({ type: [String], required: true })
-  creator: string[];
+  @Prop({ type: [String], required: false })
+  creators: string[];
```
nitrosx: I though that we were moving this field to the metadata?

src/published-data/schemas/published-data.schema.ts
```diff
@@ -166,80 +163,66 @@ export class PublishedData {
...
     type: [String],
     required: false,
-    description: "List of Names of authors of the to be published data",
+    description:
+      "The institution or person responsible for collecting, managing, distributing, or otherwise contributing to the development of the resource." +
+      "This field has the semantics of [DataCite contributor](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/contributor/)",
   })
   @Prop({ type: [String], required: false })
-  authors: string[];
+  contributors?: string[];
```
nitrosx: I though that we were moving this field to the metadata?

src/published-data/schemas/published-data.schema.ts
martin-trajanovski: Yes that is the idea to have it similar as we have in the dataset. The addition here is that `pid` is equal to `_id` as we had `_id` from before and now we are adding the `pid`.

nitrosx (approved):

PR #1931, opened 2025-05-22
fix: refactor switch action to fix #1927

### LINKED ISSUE #1927: Support switching on dataset properties in update actions
## Summary

Applying a `switch` job action to a dataset property when updating a job isn't possible, because the datasetList isn't available in the DTO. `SwitchJobAction` should be refactored to allow this, at least for the performJob phase where the full job should be available.

sbliven:
## Description

This refactors the switch action considerably to support dataset and job properties in more circumstances, fixing #1927.

Configuration Changes:
- Remove `scope` from the switch action. `property` now references the full context. Scopes should be included in the property, eg  `request.jobParams` or `datasets[*].datasetlifecycle` (**note the wildcard json path, required for arrays**).
- Add `phase` configuration to control whether the switch applies during validate, perform, or all phases.
- Rename `performJob` to `perform` in the `log` action to match the switch phase

Behavioral Changes:
- Fall through to the default case if the property doesn't match anything (this used to be an error)
- Datasets can be accessed in create jobs during the perform phase
- Job properties can be accessed in update jobs during the validate phase

## Motivation

Previously the switch property always applied to both phases, and threw an error if the property was not found. The validate phase only was able to switch over request properties. This was particularly limiting for update operations, since the request DTO doesn't overlap much with the job, so it wasn't possible to access most properties (eg any datasets). This provides significantly more flexibility.

## Fixes

* Allow `switch` action to work with more properties (#1927)

## Changes:

* jobConfig.yaml should be updated. The `switch` and `log` actions have different options.

## Tests included

- [x] Passing? 

## Documentation

https://github.com/SciCatProject/documentation/pull/75

### DIFF
--- a/jobConfig.example.yaml
+++ b/jobConfig.example.yaml
@@ -126,7 +126,7 @@ jobs:
       actions:
         - # Use 'switch' for conditional actions
           actionType: switch
-          phase: performJob
+          phase: perform
           property: request.statusCode
           cases:
             - # Literal match
--- a/jobConfig.example.yaml
+++ b/jobConfig.example.yaml
@@ -126,8 +126,8 @@ jobs:
       actions:
         - # Use 'switch' for conditional actions
           actionType: switch
-          scope: request
-          property: statusCode
+          phase: performJob
+          property: request.statusCode
           cases:
             - # Literal match
               match: finishedSuccessful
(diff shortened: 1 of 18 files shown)


sbliven:
I am working on updating docs.
I may rename the phases based on discussion in #1927.
Otherwise it's ready for review

despadam (approved):

despadam:
For the naming of the phases, I would still vote for validate / perform (or performJob if you would prefer to keep it as it currently is).
Regarding the documentation, are you referring to [this PR](https://github.com/SciCatProject/documentation/pull/73)?

sbliven:
I like @despadam's suggestion of validate/perform phases. I've updated to use those.

https://github.com/SciCatProject/documentation/pull/73 corresponds to #1914, which has already merged. I need to incorporate your changes and merge that. Afterwards I will add documentation of the #1931 changes.

PR #1934, opened 2025-05-23
fix: update value structure in elasticsearch initialSyncTransform

Junjiequan:
## Description
This PR updates the initialSyncTransform function to ensure that all scientific metadata values follow a consistent structure of: `{ value: unknown, unit: string }` 

## Motivation
Previously, some metadata fields were stored as plain values (e.g., "someKey": 123). This caused inconsistencies when indexing into Elasticsearch.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/elastic-search/helpers/utils.ts
+++ b/src/elastic-search/helpers/utils.ts
@@ -65,7 +65,7 @@ export const initialSyncTransform = (obj: DatasetClass) => {
           const transformedKey = transformKey(key);
 
           if (!isObject(value)) {
-            return [transformedKey, value];
+            return [transformedKey, { value: value, unit: "" }];
           }
 
           if ("value" in value) {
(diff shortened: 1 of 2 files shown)


martin-trajanovski (approved):
This looks good to me. Just a small clarification needed. So in ES you wont be able to store plain key-value pairs anymore `ex. someKey: "value"`. All of them need to follow the structure of `{value: unknown, unit: string}`.

Junjiequan:
> This looks good to me. Just a small clarification needed. So in ES you wont be able to store plain key-value pairs anymore `ex. someKey: "value"`. All of them need to follow the structure of `{value: unknown, unit: string}`.

Yes. 
ElasticSearch creates nested fields mapping type (schema) based on the very first doc that is inserted, and the type is shared across all the records. For example, if the mapping type is created based on `scientificMetadata.run_number:"string"`, then the following sync will throw error if they do not follow the mapping type.  `scientificMetadata.run_number:{value:"string"}` will throw error because it do not match `key: value` pair

PR #1939, opened 2025-05-26
feat: implements history for many entities

totopoloco:
## Description
This PR implements a robust history tracking system for data models in SciCat. The system provides a comprehensive audit trail by recording document changes in a centralized History collection, supporting both full document snapshots and delta-only tracking strategies.

## Motivation
Scientific data management requires detailed audit trails to track who changed what and when. This system:

* Provides accountability for data modifications
* Supports regulatory compliance requirements
* Enables reverting changes when needed
* Centralizes history records for easier querying and management

## Fixes

* Bug fixed (#X)

## Changes:
* Created a generic history schema that stores changes across multiple document types
* Implemented a reusable history plugin that can be applied to any Mongoose schema
* Added support for two tracking strategies: "document" (full snapshots) and "delta" (only changed fields)
* Integrated history tracking with Proposal and Dataset models
* Added environment variables to configure which models are tracked and how
* Implemented a history endpoint for querying historical changes
* Connected history tracking with the authentication system to record originators

### Implementation
Technical Implementation
History Schema
Created a centralized schema for storing history records with:

* Reference to original collection and document
* Before/after states
* Operation type (update/delete)
* Timestamp
* Originator (username)

### History Plugin
The plugin uses Mongoose middleware to:

* Intercept database operations (findOneAndUpdate, findOneAndDelete, etc.)
* Capture document state before changes
* Record changes after operations complete
* Support configuration through environment variables

### Configuration System
Added environment variables:

`TRACKABLES=Dataset,Proposal: Comma-separated list of models to track`
`TRACKABLE_STRATEGY=delta: Strategy for storing history ("delta" or "document")`

### Permission System
Integrated with CASL ability system to ensure:

Only authorized users can query history records

## Tests included

- [ ] Passing? 

## Documentation
- [X] official documentation updated


src/casl/casl-ability.factory.ts
```diff
@@ -407,6 +408,49 @@ export class CaslAbilityFactory {
...
+    return buildDetect();
+
+    function buildDetect() {
+      return build({
+        detectSubjectType: (item) =>
+          item.constructor as ExtractSubjectType<Subjects>,
+      });
+    }
+  }
+
```
nitrosx: Can you elaborate on this function, please?

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,377 @@
...
+    trackables = (process.env.TRACKABLES?.split(",") || []).map((t) =>
+      t.trim(),
+    ),
+    getOriginator,
+  } = options;
+
+  const trackableStrategy =
+    process.env.TRACKABLE_STRATEGY?.toLowerCase() === "delta"
+      ? "delta"
+      : "document";
```
nitrosx: We should retrieve this info from the configuration object

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,377 @@
...
+}
+
+export function historyPlugin(
+  schema: Schema,
+  options: HistoryPluginOptions = {},
+) {
+  const {
+    historyModelName = GenericHistory.name,
+    modelName: optionsModelName, // Extract modelName if provided in options
+    trackables = (process.env.TRACKABLES?.split(",") || []).map((t) =>
```
nitrosx: let's use the configuration object to retrieve this value

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,377 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const constructor = (schema as any).constructor;
+        if (constructor && constructor.modelName) {
+          modelName = constructor.modelName;
+        }
+      } catch (e) {
+        console.warn("Error accessing schema constructor:", e);
+      }
+    }
+  }
```
nitrosx: Do we need all this logic? ...or can we simplify and maybe require the model name

src/common/schemas/generic-history.schema.ts
```diff
@@ -0,0 +1,49 @@
...
+  @Prop({ type: String, required: true, index: true })
+  operation: "update" | "delete";
+
+  @ApiProperty({ description: "Timestamp of when the change occurred" })
+  @Prop({ type: Date, index: true })
+  timestamp: Date; // Automatically managed by timestamps: { createdAt: "timestamp" }
+
+  @ApiProperty({ description: "Username who performed the change" })
+  @Prop({ type: String, index: true })
+  originator?: string; // The user who made the change
```
nitrosx: Can we call this field `user`?

src/history/history.controller.ts
nitrosx, line 1: The history controller just provide an endpoint to retrieve history entries by collections/sub-systems.
Would it be possible to add an endpoint to retrieve all of them without restrictions?
You can take a look at GET Dataset v4

src/history/history.service.ts
nitrosx, line 1: We should update the service with functions to retrieve history entries with generic query/condition

nitrosx (changes requested):

src/casl/casl-ability.factory.ts
totopoloco: This enhancement has been done, it is as well in the documentation of `historyEndpointAccess` in `casl-ability.factgory.ts`:

```
 /**
   * Controls user access to the history endpoints based on role-based permissions.
   *
   * This method implements the authorization logic for accessing history records across
   * different collections (e.g., Dataset, Proposal, Sample). It follows a hierarchical
   * permission structure where:
   *
   * 1. Unauthenticated users have no access to any history
   * 2. Administrators have unrestricted access to all history records
   * 3. Regular users have access only to history for collections relevant to their role
   *
   * The third parameter in the permission definitions is particularly important:
   * - For admin users: "ALL" indicates access to all collections
   * - For specialized users: Collection name (e.g., "Dataset", "Proposal", "Sample")
   *   restricts access to only that specific collection
   *
   * When a history request is made, the controller should verify the user has
   * permission to access the requested collection by checking:
   * `ability.can(Action.HistoryRead, "GenericHistory", collectionName)`
   *
   * @param user - The authenticated user object from the JWT token
   * @returns An AppAbility object that can be used to check history access permissions
   *
   * @example
   * // In a controller:
   * const ability = this.caslFactory.historyEndpointAccess(request.user);
   * if (!ability.can(Action.HistoryRead, "GenericHistory", "Dataset")) {
   *   throw new ForbiddenException("No access to Dataset history");
   * }
   *
   * @security This method is critical for enforcing access control to potentially
   * sensitive history data. Any changes should be carefully tested to ensure proper
   * access restrictions are maintained.
   */
```

src/common/mongoose/plugins/history.plugin.ts
totopoloco: In the last commit this now uses the values from configuration.ts

src/common/mongoose/plugins/history.plugin.ts
totopoloco: Done in the same commit

src/common/mongoose/plugins/history.plugin.ts
totopoloco: Well, some developers would prefer to get the model name from different ways, but for simplicity let me rework this code

totopoloco: Simplified all the redundant checks, simplified see commit 02a1d0ff86814c207b9a840d622cfb9ad8011c65

src/common/schemas/generic-history.schema.ts
totopoloco: Sure, no problem, let me rename this field, however, a more descriptive variable name and database entry would be "triggeredBy" or "userWhoTriggered"

totopoloco: Done:

image

src/history/history.service.ts
totopoloco: **New endpoints**:

GET /history?filter={...} - Flexible querying with JSON filter
GET /history/count?filter={...} - Count history records with filters
Kept the existing GET /history/collection/:collectionName for backward compatibility

----
**Example usage**:

// Find history entries for a specific document in a collection
`GET /history?filter={"collectionName":"Dataset","documentId":"12345"}`
// Find all history entries by a specific user
`GET /history?filter={"collectionName":"Dataset","user":"john.doe"}`
// Count all delete operations for a specific collection
`GET /history/count?filter={"collectionName":"Dataset","operation":"delete"}`

totopoloco: Note that collectionName is mandatory in all queries, we need the collectionName to check for permissions.

src/history/history.controller.ts
totopoloco: See my comment below, I think it is fixed in the comment below.

src/history/history.service.ts
totopoloco: [sample_scripts_linux.zip](https://github.com/user-attachments/files/20691603/sample_scripts_linux.zip)
Added two working examples as a demo of this specific functionality.

E.g.

image

src/common/schemas/generic-history.schema.ts
nitrosx: I agree with you, although I think that one of the longer names might lead to a lot of questions

nitrosx: Let's evaluate the name after we have merged this PR in master

src/history/history.service.ts
nitrosx: Thank @totopoloco 
Questions/Feedback: 
- "collectionName" might not be the best name to use, as that is specific of mongodb. Would it be possible to change it to `subsystem` or `entity`?
- can we expand the endpoint to retrieve all the history entries between certain dates, no matter which entities they relate to?

totopoloco: I think you already get it in this endpoint:

Q. can we expand the endpoint to retrieve all the history entries between certain dates, no matter which entities they relate to?

A. Using a filter any combination should be possible.
`GET /history?filter={"collectionName":"Dataset","user":"john.doe"}`

src/common/schemas/generic-history.schema.ts
totopoloco: Already applied the change to "user":

image

src/history/history.service.ts
totopoloco: But let me create an example for the entries between dates, one moment.

totopoloco: And let me rename the other variable from `collectionName` to `subsystem`

src/casl/casl-ability.factory.ts
```diff
@@ -407,6 +409,98 @@ export class CaslAbilityFactory {
...
+      createMongoAbility<PossibleAbilities, Conditions>,
+    );
+    // Unauthenticated users cannot access history
+    if (!user || !user.currentGroups || !Array.isArray(user.currentGroups)) {
+      return buildDetect();
+    }
+
+    // Admin users have unrestricted access to all history records
+    if (
+      user.currentGroups.includes("admin") ||
```
nitrosx: Please remove this line. Admins group are listed in accessGroups.admin.

src/casl/casl-ability.factory.ts
```diff
@@ -407,6 +409,98 @@ export class CaslAbilityFactory {
...
+     * @returns A built ability object with the detectSubjectType function configured
+     */
+    function buildDetect() {
+      return build({
+        detectSubjectType: (item) =>
+          item.constructor as ExtractSubjectType<Subjects>,
+      });
+    }
+  }
+
```
nitrosx: I'm not sure I understand the purpose of this function. Can you please elaborate?

src/casl/casl-ability.factory.ts
```diff
@@ -407,6 +409,98 @@ export class CaslAbilityFactory {
...
+    }
+
+    // Users in HISTORY_DATASET_GROUPS can access only Dataset history
+    if (
+      user.currentGroups.some((g) =>
+        this.accessGroups?.historyDataset.includes(g),
+      )
+    ) {
+      can(Action.HistoryRead, "GenericHistory", "Dataset");
+      return buildDetect();
```
nitrosx: Using this function here does not allow to have a user matching multiple cases.
why not calling it only at the end of the function?

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,338 @@
...
+      try {
+        const docToUpdate = await this.model
+          .findOne(this.getQuery())
+          .lean<Document>() // Keep lean<Document> or use a specific interface for the plain object
+          .exec();
+        if (docToUpdate) {
+          // Assign the potentially plain object from lean() to _originalDoc
+          // Cast lean result if needed, or adjust _originalDoc type
+          this._originalDoc = docToUpdate as Document;
+        }
```
nitrosx, line 176: I wonder if we can consolidate all the calls to the database only in `schema.pre` or in `schema.post`?

src/common/schemas/generic-history.schema.ts
```diff
@@ -0,0 +1,49 @@
...
+
+@Schema({
+  collection: "History", // Dedicated collection for history
+  timestamps: { createdAt: "timestamp", updatedAt: false }, // Only record creation time
+  minimize: false,
+})
+export class GenericHistory {
+  @ApiProperty({ description: "The name of the collection being tracked" })
+  @Prop({ type: String, required: true, index: true })
+  collectionName: string;
```
nitrosx: rename it `entity` or `subsystem`?

src/history/history.service.ts
totopoloco: renamed to subsystem
image

nitrosx (changes requested):

src/history/history.service.ts
totopoloco: This filter can be applied:

`{{url}}/api/v3/history?filter={"timestamp":{"$gte":"2025-06-01T00:00:00.000Z","$lte":"2025-06-30T23:59:59.999Z"}}`

Encoded looks like this:

`{{url}}/api/v3/history?filter=%7B%22timestamp%22%3A%7B%22%24gte%22%3A%222025-06-01T00%3A00%3A00.000Z%22%2C%22%24lte%22%3A%222025-06-30T23%3A59%3A59.999Z%22%7D%7D`

src/common/schemas/generic-history.schema.ts
totopoloco: Done

src/casl/casl-ability.factory.ts
totopoloco: The same call is encapsulated to avoid repetition of code in every return. This is used, for example, in `jobsEndpointAccess`. I believe that the implementation of `historyEndpointAccess` is more appropriate, as it avoids chaining long `if` statements and makes the code more readable. Would you agree, or would you suggest an alternative approach?

totopoloco: ah in a new commit I adapted to the style of the other functions, see the commit: 0f44e0d6111c8267f4f07bcca692969a4ed62c84

src/casl/casl-ability.factory.ts
totopoloco: Done in 0f44e0d6111c8267f4f07bcca692969a4ed62c84

src/casl/casl-ability.factory.ts
totopoloco: Done in 0f44e0d6111c8267f4f07bcca692969a4ed62c84

src/common/mongoose/plugins/history.plugin.ts
totopoloco: ## Explanation

1. In `pre` hooks: We fetch the original document
2. In `post` hooks: We create the history record

We can't completely move all database operations to just one type of hook because:

- **Pre hooks**: We need to capture the document's state before modification, but we don't know the final state yet
- **Post hooks**: We need to record after the operation is complete, but by then the original state would be lost (especially for deletions)

## Optimized Approach

However, in commit ad60dbf4d3f808ae02930b5192215e47b1f23304 was minimized database calls and making the code more maintainable

src/casl/casl-ability.factory.ts
```diff
@@ -425,6 +430,126 @@ export class CaslAbilityFactory {
...
+   */
+  historyEndpointAccess(user: JWTUser) {
+    const { can, build } = new AbilityBuilder(
+      createMongoAbility<PossibleAbilities, Conditions>,
+    );
+
+    // -------------------------------------
+    // Unauthenticated users
+    // -------------------------------------
+    if (!user || !user.currentGroups || !Array.isArray(user.currentGroups)) {
```
Junjiequan: Users without currentGroup will always log `unauthenticated user attempted` falsy message.
I'd suggest to follow `if(user)...`  instead of `if(!user)...` patterns like the rest of the functions for easier maintainance

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,305 @@
...
+    getActiveUser: getActiveUser,
+    configService,
+    // Extract from options with fallbacks to ConfigService values
+    trackableStrategy: explicitStrategy,
+    trackables: explicitTrackables,
+  } = options;
+
+  // Use options if provided, otherwise try ConfigService, then fall back to defaults
+  const trackableStrategy =
+    explicitStrategy ??
```
Junjiequan: I think `||` operator should be used here.  If explicitStrategy is a empty string then trackableStrategy will be assigned with "delta", as empty string is not nullish

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,305 @@
...
+  // Get the model name from options
+  if (!optionsModelName) {
+    // If not provided, warn and skip setup
+    console.warn(
+      "HistoryPlugin: Could not determine model name for schema. Please provide a modelName in plugin options.",
+    );
+    return; // Skip setup if we can't determine the model name
+  }
+
+  const modelName = optionsModelName;
```
Junjiequan: Do we need optionsModelName alias in the first place?

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,305 @@
...
+      await HistoryModel.create({
+        subsystem: modelName,
+        documentId: originalDoc._id,
+        before: beforeData,
+        after: afterData,
+        operation: operation,
+        user: user,
+      });
+    } catch (error) {
+      console.error(
```
Junjiequan: Please replace all the console log with Logger, so that logs can be collected in the logging system if enabled

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,305 @@
...
+}
+
+export function historyPlugin(
+  schema: Schema,
+  options: HistoryPluginOptions = {},
+) {
+  const {
+    historyModelName = GenericHistory.name,
+    modelName: optionsModelName, // Extract modelName if provided in options
+    getActiveUser: getActiveUser,
```
Junjiequan: Is the same name alias needed here?

README.md
```diff
@@ -219,7 +219,17 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
...
+| `TRACKABLES` | string | Yes | The TRACKABLES environment variable configures which data models are tracked by the history system. When specified, only models listed in this variable will have their changes recorded in the History collection. | Dataset\[,Proposal\]\[,Sample\]\[,Instrument\]\[,PublishedData\] |
+| `HISTORY_ACCESS_DATASET_GROUPS` | string | Yes | Roles in this list will be able to access history dataset records | \[role1\]\[,role2\]\[,roleN\]... |
+| `HISTORY_ACCESS_PROPOSAL_GROUPS`| string | Yes | Roles in this list will be able to access history proposal records | \[role1\]\[,role2\]\[,roleN\]... |
+| `HISTORY_ACCESS_SAMPLE_GROUPS` | string | Yes | Roles in this list will be able to access history sample records | \[role1\]\[,role2\]\[,roleN\]... |
+| `HISTORY_ACCESS_INSTRUMENT_GROUPS` | string | Yes | Roles in this list will be able to acces history instrument records | \[role1\]\[,role2\]\[,roleN\]... |
```
abdimo101: Minor nitpick: "acces" --> "access" in line 228.

README.md
abdimo101, line 1: Noticed the order of the new env variables in env.example and the README is a bit different. Could be nice to keep them consistent.

abdimo101 (changes requested):
Tried out the history feat and its working well on my end. I have left some small comments.

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,305 @@
...
+      next();
+    },
+  );
+
+  schema.post(
+    "findOneAndDelete",
+    // Specify the expected result type (Document | null) and the Document type
+    async function (
+      this: QueryWithHistory<Document | null, Document>,
+      doc: Document | null, // The document *before* it was deleted (Mongoose behavior)
```
Junjiequan: es-lint complains `'doc' is defined but never used`

src/casl/action.enum.ts
```diff
@@ -248,4 +248,9 @@ export enum Action {
   AttachmentReadInstance = "attachment_read_instance",
   AttachmentUpdateInstance = "attachment_update_instance",
   AttachmentDeleteInstance = "attachment_delete_instance",
+
+  // -------------------------------------
+  // History
+  // -------------------------------------
+  HistoryRead = "history_read",
```
Junjiequan: We have two layers of access check: endpoint access, instance access.
Endpoint access checks if user has permission to access the service, and instance access checks if user has access to the specific data or not.
I'd suggest applying the same logic for consistency.

src/app.module.ts
```diff
@@ -165,4 +180,8 @@ import { MetricsModule } from "./metrics/metrics.module";
     },
   ],
 })
-export class AppModule {}
+export class AppModule implements NestModule {
+  configure(consumer: MiddlewareConsumer) {
+    consumer.apply(AccessTrackingMiddleware).forRoutes("*");
+  }
+}
```
Junjiequan: `consumer.apply(AccessTrackingMiddleware).forRoutes("*");` will override middleware usage from **metrics module** `consumer.apply(AccessTrackingMiddleware).exclude(...exclude).forRoutes(...include);` which is customizable - where siteAdmin can decide which access data to collect

README.md
totopoloco: Fixed in 6eb78dc3b5384487fd9068719995b822db82ad9d

README.md
totopoloco: Fixed in 69d5fa4c1efed61976c82e3088162119f3b20553

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,305 @@
...
+          docObj,
+        );
+
+        beforeData = originals;
+        afterData = delta;
+      } else {
+        beforeData = originalDoc;
+        afterData = null;
+      }
+
```
Junjiequan: I find this code hard to read, shouldn't it be `if (x == document){} else if (x === delta){} else{};` ?
Is there any reason to use updatedDoc as condition for `else if` rather than put it inside `else if`?

src/datablocks/datablocks.service.ts
```diff
@@ -62,16 +62,32 @@ export class DatablocksService {
...
   async remove(filter: FilterQuery<DatablockDocument>): Promise<unknown> {
-    return this.datablockModel.findOneAndDelete(filter).exec();
+    console.log("Removing datablock with filter:", filter);
+    const existingDoc = await this.datablockModel.findOne(filter).exec();
+
+    if (!existingDoc) {
+      console.log("No document found to remove with filter:", filter);
+      return null;
+    }
+    const result = await this.datablockModel.findOneAndDelete(filter).exec();
```
Junjiequan: `findOneAndDelete` returns null if no document mataches https://www.geeksforgeeks.org/mongodb/mongodb-findoneanddelete-method/  
the extra `findOne` call can be removed

src/datasets/datasets.service.ts
```diff
@@ -72,6 +72,18 @@ export class DatasetsService {
...
+      // This assumes your request.user has a username property
+      // Cast request.user to JWTUser since you're using that type elsewhere
+      const user = this.request.user as JWTUser;
+      return user?.username || "system";
+    } catch (error) {
+      console.warn("Could not get username from request:", error);
+      return "system";
+    }
+  }
+
```
Junjiequan: I don't see this function imported from anywhere?

src/history/history.controller.ts
```diff
@@ -0,0 +1,274 @@
...
+    @Req() request: Request,
+    @Query("filter") filterStr: string,
+    @Query("skip") skip?: number,
+    @Query("limit") limit?: number,
+  ) {
+    if (!request.user) {
+      throw new UnauthorizedException(
+        "Authentication required to access history data",
+      );
+    }
```
Junjiequan: This should be handled by the `@CheckPolicies` at endontPointAccess layer, instead of explicitly throwing error here

src/history/history.controller.ts
```diff
@@ -0,0 +1,274 @@
...
+  })
+  async countHistory(
+    @Req() request: Request,
+    @Query("filter") filterStr: string,
+  ) {
+    if (!request.user) {
+      throw new UnauthorizedException(
+        "Authentication required to access history data",
+      );
+    }
```
Junjiequan: same as above

src/history/history.controller.ts
```diff
@@ -0,0 +1,274 @@
...
+    @Param("subsystem") subsystem: string,
+    @Req() request: Request,
+    @Query("skip") skip?: number,
+    @Query("limit") limit?: number,
+  ) {
+    if (!request.user) {
+      throw new UnauthorizedException(
+        "Authentication required to access history data",
+      );
+    }
```
Junjiequan: same here

src/common/utils/request-context.util.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (!request.user) {
+      return "system"; // Default value if no user is available
+    }
+    const requestUser = request.user;
+    const anyRequestUser = requestUser as RequestUser;
+    return anyRequestUser.username || "system"; // Return the username or a default value
+  } catch (e) {
+    console.error("Error getting current username:", e);
+    return "system";
+  }
```
Junjiequan: nitpick:  Isn't it misleading to use `system` for `unknown user`?

src/published-data/published-data.service.ts
```diff
@@ -106,16 +107,38 @@ export class PublishedDataService {
...
       )
       .exec();
   }
 
   async remove(filter: FilterQuery<PublishedDataDocument>): Promise<unknown> {
-    return this.publishedDataModel.findOneAndDelete(filter).exec();
+    console.log("Removing published data with filter:", filter);
+
+    const existingDoc = await this.publishedDataModel.findOne(filter).exec();
+
```
Junjiequan: `findOne` is redundant, as mentioned before

Junjiequan (changes requested):
The implementation of the dynamic history system plugin looks great to me. 

I've left a few comments for some minor details and would love to hear your thoughts about them.

src/casl/casl-ability.factory.ts
totopoloco: Fixed in 3e42b58a5dd6e92e69e53b2b96341ebc8fef3909

src/common/mongoose/plugins/history.plugin.ts
totopoloco: Fixed in 445d2385d04a4aa9848e1aeaaa49ac1b8898f643

src/common/mongoose/plugins/history.plugin.ts
totopoloco: Fixed in 3320ff7c6f2805db345dddb2f7785bc60f6874d5

src/common/mongoose/plugins/history.plugin.ts
totopoloco: Fixed in 24834e031ab9eb5f324592bfe609485925ac0308

src/common/mongoose/plugins/history.plugin.ts
totopoloco: Fixed in 46843902e62c646f2c1920b1171614c5f3bca2a6

src/history/history.controller.ts
totopoloco: Fixed in bf0a9536deda4bc8d17e05607c6d76e91a8979a2

src/history/history.controller.ts
totopoloco: Fixed in bf0a9536deda4bc8d17e05607c6d76e91a8979a2

src/history/history.controller.ts
totopoloco: Fixed in bf0a9536deda4bc8d17e05607c6d76e91a8979a2

src/common/mongoose/plugins/history.plugin.ts
totopoloco: Fixed in f28f33acf55d43f8e3fd08b87490ade2eac90254

src/common/mongoose/plugins/history.plugin.ts
totopoloco: Fixed in bc7e8a671be3162328499c0b181a015049670eaa

src/datablocks/datablocks.service.ts
totopoloco: Fixed in ef9b12c66a672c463841ecc542f0f4ee91441354

src/datasets/datasets.service.ts
totopoloco: Fixed in e75d1f3d122133abe03ac42406622da9cde38230

src/common/utils/request-context.util.ts
totopoloco: Fixed in 672c2b249ac1ef37e380d80bab109c19b5f65f91

src/published-data/published-data.service.ts
totopoloco: Fixed in 40b1fe1ae2b47abc2efcf05d34918af9f7ffbeec

src/casl/action.enum.ts
totopoloco: Dear @Junjiequan ,
It is my hope that this commit will fulfil your requirement. However, should it not do so, I ask that you take ownership of the issue:

Fixed in 2b7ec218c9379bdccfcc15478d74286d901dda9f

src/app.module.ts
totopoloco: Fixed in a1e6ae44ca60daf43e681e44395ca4e991a6be0a

src/common/mongoose/plugins/history.plugin.ts
```diff
@@ -0,0 +1,369 @@
...
+  GenericHistoryDocument,
+} from "../../schemas/generic-history.schema";
+import { computeDeltaWithOriginals } from "../../utils/delta.util";
+import { Logger } from "@nestjs/common";
+
+/**
+ * Logger instance for the history plugin
+ * All logs will be prefixed with "HistoryPlugin" for easy filtering
+ */
+const logger = new Logger("HistoryPlugin");
```
Junjiequan: It’s generally a good idea to have separate logger instances for specific logs.
However, in Scicat, we already have a centralized scicatLogger service that handles both local and external logging.
So, instead of creating a new logger instance here, we should just use the scicatLogger.
For easy filtering, pass the filter string as the second argument, e.g.:
`Logger.warn(
  "Could not determine model name for schema. Please provide a modelName in plugin options.",
  "HistoryPlugin"
);`

totopoloco: Dear @nitrosx and @Junjiequan, I would like to ask you to take ownership of this issue.

Kind regards,

~M

Junjiequan: Thank you, @totopoloco, for your great contribution to this feature and for addressing all my comments. I’ll take over from here to make the final adjustments. I just want to acknowledge and appreciate your effort—it’s been a big help!

Junjiequan (approved):
Looks good for me, but need an another review

abdimo101 (approved):

src/policies/policies.service.ts
```diff
@@ -138,14 +134,34 @@ export class PoliciesService implements OnModuleInit {
...
+          new: true,
+          runValidators: true,
+        },
+      )
       .exec();
   }
 
   async remove(filter: FilterQuery<PolicyDocument>): Promise<unknown> {
-    return this.policyModel.findOneAndDelete(filter).exec();
+    console.log("Removing policy with filter:", filter);
```
nitrosx: should we change it to use the main logging facility?

src/attachments/attachments.v4.service.ts
```diff
@@ -79,14 +78,33 @@ export class AttachmentsV4Service {
   ): Promise<Attachment | null> {
     const username = (this.request?.user as JWTUser).username;
 
+    //log username for auditing purposes
+    if (!username) {
+      console.warn("No username found for auditing");
```
nitrosx: should we change it to use the main logging facility?

src/attachments/attachments.v4.service.ts
```diff
@@ -79,14 +78,33 @@ export class AttachmentsV4Service {
...
+            updatedBy: username,
+            updatedAt: new Date(),
+          },
+        },
+        { new: true, runValidators: true },
       )
       .exec();
 
+    if (!result) {
+      console.warn(
```
nitrosx: should we change it to use the main logging facility?

nitrosx (changes requested):
minor changes to remove console log and use main scicat logger

nitrosx (approved):

GBirkel:
I don't suppose we can get a patch to the front end to bring the API in sync?
```
gwbirkel@MacBook-Air scicat-frontend-als % npx ng build
Application bundle generation failed. [6.270 seconds]

✘ [ERROR] TS2339: Property 'history' does not exist on type 'DatasetClass'. [plugin angular-compiler]

    src/app/datasets/dataset-lifecycle/dataset-lifecycle.component.ts:68:27:
      68 │     if (dataset && dataset.history) {
         ╵                            ~~~~~~~


✘ [ERROR] TS2339: Property 'history' does not exist on type 'DatasetClass'. [plugin angular-compiler]

    src/app/datasets/dataset-lifecycle/dataset-lifecycle.component.ts:69:30:
      69 │       const history = dataset.history.map(
         ╵                               ~~~~~~~
```

Issue #1757, opened 2025-03-13
Cleanup Legacy Fields and Migrate isPublished to Child Schema


Junjiequan:
## Summary
We need to clean up legacy fields across our schemas and refactor the usage of the isPublished field. Specifically:
- Review fields such as instrumentGroup that may no longer be relevant or used
- The isPublished field should be moved to the child schema class (i.e., those extending OwnableSchema) where it belongs.

PR #1953, opened 2025-05-30
feat(APIv4): add origdatablock v4 controllers

### LINKED ISSUE #2034: Anonymous user can read non-public `OrigDatablock`
## Summary
An anonymous user can see non-public (`isPublished == false`) `OrigDatablocks` by calling the `/api/v3/origdatablocks` endpoints.

## Steps to Reproduce
1. Start scicalive
2. Run `curl 'http://backend.localhost/api/v3/origdatablocks/psi_orig3' -H 'accept: application/json'` or `curl 'http://backend.localhost/api/v3/origdatablocks' -H 'accept: application/json'`

## Current Behaviour
The GET endpoints on the origdatablock controller don't properly filter the results based on group membership.
Similarly to the anonymous access, you can log in as `proposalIngestor` and read the `psi_orig3` object if you know its ID without the user being in the `ownerGroup` or `accessGroups`

## Expected Behaviour
Same as every `Ownable` Resource

HayenNico:
## Description
Add new controllers `OrigDatablocksV4Controller` and `OrigDatablocksPublicV4Controller`, mirroring the structure for `Datasets`. Add required functionality in the `OrigDatablocksService` and refactor support functions from DatasetsV4 to the common module where possible

## Motivation
The improvements from API v4 currently only cover the `Dataset` subsystem, other collections are not yet available under `/api/v4/` endpoints. Instead of copying large parts of the v3 API, v4 should use the new querying standards developed for Datasets.
This PR is a trial run to transfer the same functionality to the `OrigDatablocks` system.

## Changes:

* add `OrigDatablocksV4Controller` and `OrigDatablocksPublicV4Controller` with full v3 functionality
* merge GET /api/v3/origdatablocks and /api/v3/fullquery endpoints to /api/v4/origdatablocks
* move GET /api/v3/origdatablocks/fullquery/files endpoint to /api/v4/origdatablocks/files
* copy querying machinery from Datasets v4 implementation
* update `OrigDatablocksService` with new functions to accept queries in the v4 format
* add `isPublished` as optional field in `UpdateOrigDatablockDto`
* add unit and api tests for new systems
* add `pid` as required field in `OutputDatasetDto` (was optional as inherited from `CreateDatasetDto`)

* Refactor of support functions: 
* FilterValidationPipe and IncludeValidationPipe moved to common module
* Needed constants moved into collection/types/collection-lookup.ts files for import
* All API v4 controllers refactored to use common module pipes, removed individual versions

## Tests included

- [x] Passing? 

## Documentation


HayenNico:
One additional proposed change:
Right now during `OrigDatablock` creation any `ownerGroup`, `accessGroups` or `instrumentGroup` specified in the `CreateOrigDatablockDto` will always be overwritten by the fields of the `Dataset` referenced by `datasetId`. I propose changing that behavior to only use the `Dataset` values if the fields are not set in the DTO. Opinions?

HayenNico:
Plus one minor oddity:
Swagger does not add own tabs for the two OrigDatablock controllers like for Dataset, instead the public endpoints are listed under origdatablocks v4. Any idea why?
grafik

Junjiequan:
> Plus one minor oddity: Swagger does not add own tabs for the two OrigDatablock controllers like for Dataset, instead the public endpoints are listed under origdatablocks v4. Any idea why? 

The API tag for origdatablock public controller needs to be changed

HayenNico:
> The API tag for origdatablock public controller needs to be changed

Okay, will be fixed. Not sure how I managed to miss that...

src/casl/casl-ability.factory.ts
```diff
@@ -1364,7 +1372,7 @@ export class CaslAbilityFactory {
           accessGroups: { $in: user.currentGroups },
         });
         can(Action.OrigdatablockReadOneAccess, OrigDatablock, {
-          ownerGroup: { $in: user.currentGroups },
+          isPublished: true,
```
martin-trajanovski, line 1432: So as I understand with this change the `CREATE_DATASET_PRIVILEGED_GROUPS` can only read published `Origdatablock` and not all that are in their access group

src/datasets/dto/output-dataset.dto.ts
```diff
@@ -3,6 +3,14 @@ import { CreateDatasetDto } from "./create-dataset.dto";
 import { IsDateString, IsString } from "class-validator";
 
 export class OutputDatasetDto extends CreateDatasetDto {
+  @ApiProperty({
+    type: String,
+    required: true,
+    description: "Persistent identifier of the dataset.",
+  })
+  @IsString()
+  declare pid: string;
```
martin-trajanovski, line 12: I think we already include the `pid` as part of the `CreateDatasetDto` but maybe the idea here is to mark it as mandatory and not optional field.

src/origdatablocks/dto/update-origdatablock.dto.ts
```diff
@@ -55,6 +56,15 @@ export class UpdateOrigDatablockDto extends OwnableDto {
...
   declare readonly ownerGroup: string;
+
+  @ApiProperty({
+    type: Boolean,
+    required: false,
+    description: "Flag is true when data are made publicly available.",
+  })
+  @IsOptional()
+  @IsBoolean()
+  readonly isPublished?: boolean;
```
martin-trajanovski, line 67: I guess the idea behind adding the field in the update dto is to be able to control it from outside by allowing a user with proper access to update the `isPublished` flag

src/origdatablocks/origdatablocks-public.v4.controller.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { Test, TestingModule } from "@nestjs/testing";
+import { OrigDatablocksPublicV4Controller } from "./origdatablocks-public.v4.controller";
+import { OrigDatablocksService } from "src/origdatablocks/origdatablocks.service";
+import { ConfigModule } from "@nestjs/config";
+
+class OrigDatablocksServiceMock {}
+
+describe("DatasetsController", () => {
```
martin-trajanovski: I guess the name should be changed to OrigDatablocksController here or `OrigDatablocksPublicController`

src/origdatablocks/origdatablocks.controller.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+import { CaslAbilityFactory } from "src/casl/casl-ability.factory";
+import { ConfigModule } from "@nestjs/config";
+
+class OrigDatablocksServiceMock {}
+
+class DatasetsServiceMock {}
+
+class CaslAbilityFactoryMock {}
+
+describe("DatasetsController", () => {
```
martin-trajanovski: Naming here should be fixed as well

src/origdatablocks/origdatablocks.v4.controller.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+import { CaslAbilityFactory } from "src/casl/casl-ability.factory";
+import { ConfigModule } from "@nestjs/config";
+
+class OrigDatablocksServiceMock {}
+
+class DatasetsServiceMock {}
+
+class CaslAbilityFactoryMock {}
+
+describe("DatasetsController", () => {
```
martin-trajanovski: Here as well

src/common/pipes/filter-validation.pipe.ts
```diff
@@ -1,40 +1,12 @@
 import { PipeTransform, Injectable } from "@nestjs/common";
```
martin-trajanovski, line 1: I like the idea that you made this reusable and moved it to a common module

martin-trajanovski (commented):
The PR in general looks great. I have left few minor comments but otherwise I think you did a great job.

src/casl/casl-ability.factory.ts
HayenNico: The removed check for ownerGroup here was a duplicate, the exact same check is still in the code a few lines above. This one was a copy-paste error I assume, where there should always have been the isPublished check..

src/datasets/dto/output-dataset.dto.ts
HayenNico: Exactly. Since OrigDatablocks always point to a Dataset `pid` with `datasetId`, for a few functions in the controller I need a guarantee that the return value from a Dataset lookup will always have a `pid` field.

src/origdatablocks/dto/update-origdatablock.dto.ts
HayenNico: This is specifically in the update DTO because the base class `OwnableDto` does not include `isPublished`, so each subclass needs to add it manually. From #1757 I got the impression that long-term isPublished may not stay in the `Ownable` schema, so i opted to put it here

src/origdatablocks/origdatablocks-public.v4.controller.spec.ts
HayenNico: Fixed

src/origdatablocks/origdatablocks.controller.spec.ts
HayenNico: FIxed

src/origdatablocks/origdatablocks.v4.controller.spec.ts
HayenNico: FIxed

Junjiequan (approved):

martin-trajanovski (approved):

nitrosx (approved):

Issue #1937, opened 2025-05-23
Jobs: create separate controllers for v3 and v4


despadam:
## Summary
Currently there is only one `jobs.controller` containing methods for both V3 and V4. This causes a duplicated suffix in the methods in the SDK.

## Current Behaviour
For example these methods currently looks like this in the SDK:
```
jobsControllerCreateV3V3()
jobsControllerCreateV4V4()
```

## Expected Behaviour
But we want them to look like this:
```
jobsControllerCreateV3()
jobsControllerCreateV4()
```

## Todo
Split the methods into two separate controllers, depending on the version, similarly to how it is done for datasets.

PR #1954, opened 2025-06-02
fix: create separate v3, v4 Jobs controllers

despadam:
## Description

Previously there was only one `jobs.controller` containing all endpoints for both V3 and V4. This PR splits this into two separate files, and moves the common util functions in a separate file. This way not only do we fix the issue of the suffix duplication in the SDK methods, but also make the code more readable and easy to review.

## Motivation

https://github.com/SciCatProject/scicat-backend-next/issues/1937

## Tests included

- [ ] Passing? 

## Documentation


sofyalaski (approved):
looks good to me!

PR #1967, opened 2025-06-04
feat(datafile): add of datafile metadata field

MatthMig:
## Description
Add of optional metadata field for element in datafileList in OrigDatablock

## Motivation
This would open the possibility to add information specific to a datafile, for example duration, measurement type, sample type. This could later open possibilities for new views to show characteristics of datafiles compounding a dataset

## Changes:
- Add of metadata field in _datafile_ **dto** and **scheme** files and in _common interface_ file
- Add of metadata filed in tests of _origdatablocks_ and _datablocks_

## Documentation

I suppose that I should update the model at https://www.scicatproject.org/documentation/Development/v4.x/Data_Model.html ?


nitrosx (commented):
Can you please add the appropriate tests under datasets to make sure that if we add datasets with datafile metadata, we are able to retrieve them.
We can brainstorm on the use cases

HayenNico:
Basen on the examples provided in the Swagger API documentation for the new field, I'm opposed to this change as it conflates responsibilities of `Datasets` and `OrigDatablocks` / `DataFiles`.

In my view, there needs to be a distinction between the research data as an abstract concept and the files. Things like used sample or technique are not properties of the file, but the research data it represents/expresses.
In terms of SciCat models, `Datasets` are the fitting structure for such information, while `OrigDatablocks` and `DataFiles` should concern themselves only with the technical aspects of storage and file format.

MatthMig:
> Basen on the examples provided in the Swagger API documentation for the new field, I'm opposed to this change as it conflates responsibilities of `Datasets` and `OrigDatablocks` / `DataFiles`.
> 
> In my view, there needs to be a distinction between the research data as an abstract concept and the files. Things like used sample or technique are not properties of the file, but the research data it represents/expresses. In terms of SciCat models, `Datasets` are the fitting structure for such information, while `OrigDatablocks` and `DataFiles` should concern themselves only with the technical aspects of storage and file format.

This is because I think there is a small misunderstanding here, the purpose is not to add information about the sample itself, that we still want to store in the datasets, but about the sample type. In the use case at ILL facility, all files of a same dataset will still all have same proposal and same sample.

Each file in the dataset has a specific role regarding the data reduction. For example, in a dataset that have as technique Powder diffractometers,  files can have: 
* TypeOfMeasure
  * Diff Powder
  * Diff Single Crystal"
  * Flipper ratio
  * Other
* typeOfSample
  * EmptyCell
  * EmptyInst
  * SampleRef
  * SampleNormalisation
  * SampleAbsorption
  * Sample
* SubType
  * None
  * Ni
  * B4C
  * Vana
  * H2O
  * Cad
  * Gado
  * Other

So it is relevant to have still technique and the sample description in the dataset, but having the sample type and measurement type is proper to a run, so it is best to have it tied to the file itself, which are stored in the OrigDatablocks.

HayenNico:
Okay, I understand that you only want to add a few controlled terms on `DataFile` instances in your use case. My main concern is that this change enables arbitrarily complex metadata (`Record<string, unknown>`) in two locations, `Dataset.scientificMetadata` and `DataFile.metadata`, which leads to unnecessary ambiguity - users could in principle start putting all metadata at the DataFile level. For your case, would it not suffice to have metadata as `Record<string, string>` as an immediate fix?

More broadly, the problem seems to be that there's no rigorous way in `Dataset` to express which `OrigDatablock`/`DataFile` corresponds to which part of a series of measurements, yes?

MatthMig:
> Okay, I understand that you only want to add a few controlled terms on `DataFile` instances in your use case. My main concern is that this change enables arbitrarily complex metadata (`Record<string, unknown>`) in two locations, `Dataset.scientificMetadata` and `DataFile.metadata`, which leads to unnecessary ambiguity - users could in principle start putting all metadata at the DataFile level. For your case, would it not suffice to have metadata as `Record<string, string>` as an immediate fix?
> 
> More broadly, the problem seems to be that there's no rigorous way in `Dataset` to express which `OrigDatablock`/`DataFile` corresponds to which part of a series of measurements, yes?

Actually I see what your concern is, but it did not seem to me that it was too ambiguous, as currently if we want to add more metadata about certain files of a dataset, Record<string, unkown> do the trick but the thing is that in the frontend I'd like to implement a box to display more information about datafiles metadata of a dataset like this:  
image

Therefore, the idea is that we have to have some metadata stored somewhere about the datafiles, so to have something clean, we thought that best way would be to have metadata directly attached to the concerned datafile.

But in the end, we could definitely simply use Record<string, unkown> but we would therefore have to see how to cleanly integrate the datafiles metadata overview box.

sbliven:
@MatthMig Would archive systems be expected to copy over the metadata for files when converting an `OrigDataBlock` to a `DataBlock`? They normally duplicate the DataFile entries rather than sharing references.

I also feel like `DataFile` should keep the narrow information about the file itself, and metadata should live elsewhere.

MatthMig:
> @MatthMig Would archive systems be expected to copy over the metadata for files when converting an `OrigDataBlock` to a `DataBlock`? They normally duplicate the DataFile entries rather than sharing references.
> 
> I also feel like `DataFile` should keep the narrow information about the file itself, and metadata should live elsewhere.

I have no idea about the archive systems, I do not know anything about it as I did not need it for the ILL implementation of SciCat. Concerning **DataFile** I do agree that it should keep information about the file itself, but the tags _sampleType_, _sampleSubType_, _measurementType_ and _duration_ are specific to a file itself, so at ILL they are indeed a narrow information for a raw data file.

In the end we'd like to have the possibility to describe a file the way that is canonical in our facility, but from a technical perspective having the metadata of each file stored in the dataset rather than in **DataFile** will add difficulty for the ingestion and maybe to represent the dataset overview (even though with a dynamic display leveraging map it should be fine). 

Having the metadata with the file will just be more efficient in our case for the ingestion, by reducing number of call necessary to ensure we correctly ingest files by keeping integrity between the DataFileList and the metadata for each file in scientificMetadata.

I do not truly see where an ambiguity is generated, because from a facility point of view when i want to store metadata specific to a file, I want to store it with the other predefined metadata for the same file and not in another location. The only thing I could see is that one could put on each data file of a dataset the exact same metadata about something that is in fact common to all data file in the dataset. But in that case one should quickly see the problem as he will not be able to have a clean visualization in the frontend.

MatthMig:
I have noticed that proposal DTO has itself already a metadata field: https://github.com/SciCatProject/scicat-backend-next/blob/master/src/proposals/dto/update-proposal.dto.ts, I understand that it may raise concern to add metadata in DataFile, but if metadata field has been added to proposal, why wouldn't we also add metadata field for sample DTO and DataFile DTO ?

Maybe am i wrong, but maybe SciCat should put all metadata of form _Record<string, unknown>_ in scientificMetadata of dataset collection, **or** we should have the possibility to have metadata field for proposal, sample, instrument, dataset, and datafile.

To me, the latter is the best idea by far, otherwise every dataset that share common proposal, sample, instrument or dataFile would have useless redundancy for the metadata.

PR #1983, opened 2025-06-11
fix: proposals endpoint access fix and improvement

martin-trajanovski:
## Description
This PR aims to fix and improve the proposals endpoint access.

## Motivation
We have experienced some issues with the previous setup where for example if you define `ADMIN_GROUPS=admin, globalaccess` and `DELETE_GROUPS=admin, archivemanager` (admin is part of both) the `DELETE_GROUPS` condition in the `casl-ability.factory.ts` overwrites all other access that the user has as an admin.

## Fixes

* proposals endpoint access for admin and other logged in users

## Changes:

* changes are only in the `casl-ability.factory.ts` where we add the improvement for proposals endpoint access management.

## Tests included

- [ ] Passing? 

## Documentation


nitrosx (approved):

nitrosx:
@martin-trajanovski have you checked if the same issue is valid for datasets and the other sub-systems?

PR #1985, opened 2025-06-13
fix: proposal instance access fixes and improvements

martin-trajanovski:
## Description
This PR aims to fix and improve the logic in the proposal instance permissions.

## Motivation
The proposal instance permissions had some issues if a user is part of `ADMIN_GROUPS` and `DELETE_GROUPS` at the same time.

## Fixes

* proposal instance permissions

## Changes:

* changes are mainly in the `casl-ability.factory.ts`

## Tests included

- [ ] Passing? 

## Documentation


src/casl/casl-ability.factory.ts
```diff
@@ -1582,91 +1582,99 @@ export class CaslAbilityFactory {
...
-      });
-      can(Action.ProposalsAttachmentDeleteOwner, ProposalClass, {
-        ownerGroup: { $in: user.currentGroups },
-      });
-    } else if (user) {
-      /**
-       * authenticated users
-       */
+        can(Action.ProposalsCreateAny, ProposalClass);
+        can(Action.ProposalsUpdateAny, ProposalClass);
```
HayenNico, line 1610: Does it make sense to give PROPOSAL_GROUPS members CreateAny and UpdateAny rights, but access restricted Read rights? CreateOwner / UpdateOwner feels more appropriate here

HayenNico (approved):
These changes look good to me.

Added one comment on inconsistent permissions for PROPOSAL_GROUPS members for future reference, but that had been present before

rkweehinzmann (approved):
from what I can say: OK.

src/casl/casl-ability.factory.ts
martin-trajanovski: @HayenNico To be honest I haven't touched those rules and I guess they were set a while ago. Maybe we should review all the logic here once more for every endpoint and instance. I know Jay was thinking of refactoring and simplifying this. I don't want to change those as part of this PR as it might introduce some other issues but if you agree we can create a ticket and go trough this in a separate PR.

PR #2000, opened 2025-06-20
feat(api v3): add Datablock controller

### LINKED ISSUE #1964: Implement Datablocks controller
## Summary

The `/api/v3/Datablocks` endpoint and related endpoints are not implemented in scicat-backend-next.

## Details

At a minimum the following endpoints should be supported:

```
/api/v3/Datablocks (GET,POST)
/api/v3/Datablocks/count
/api/v3/Datablocks/{id} (GET,PATCH,DELETE)
```

These should be backwards compatible with the scicat-v3 data model (although I don't believe this changed).

It would make sense to also make the same endpoints available as `/api/v4/*`, but that is less urgent.

fpotier:
Needed by #1964


src/casl/casl-ability.factory.ts
```diff
@@ -638,6 +640,50 @@ export class CaslAbilityFactory {
...
+        can(Action.DatablockRead, Datablock);
+        can(Action.DatablockUpdate, Datablock);
+        can(Action.DatablockDelete, Datablock);
+      }
+
+      if (
+        user.currentGroups.some((g) => this.accessGroups?.delete.includes(g))
+      ) {
+        cannot(Action.DatablockCreate, Datablock);
+        cannot(Action.DatablockRead, Datablock);
```
omkar-ethz: This seems a bit counter-intuitive to remove permissions because a user belongs to certain group.
Let's say user belongs to both an admin group and a delete group. This will remove read / create permissions from the user (although the user should have it from the admin group).

fpotier: You're right! fixed it

src/datablocks/datablocks.controller.ts
```diff
@@ -0,0 +1,364 @@
...
+      const ability = this.caslAbilityFactory.datablockInstanceAccess(user);
+
+      if (!instance) {
+        throw new NotFoundException();
+      }
+
+      if (
+        ability.cannot(Action.DatablockUpdateInstance, instance) ||
+        ability.cannot(Action.DatablockUpdateAny, Datablock)
+      ) {
```
omkar-ethz: This should be an &&:
Don't throw exception if `(can DatablockUpdateInstance OR  can DatablockUpdateAny)`
=> throw exception  if `not(can DatablockUpdateInstance OR  can DatablockUpdateAny)`
=> throw exception  if `(cannot DatablockUpdateInstance AND  cannot DatablockUpdateAny)`

src/datablocks/datablocks.controller.ts
```diff
@@ -0,0 +1,364 @@
...
+    if (!instance) {
+      throw new NotFoundException();
+    }
+
+    if (
+      abilities.cannot(Action.DatablockReadInstance, instance) ||
+      abilities.cannot(Action.DatablockReadAny, Datablock)
+    ) {
+      throw new UnauthorizedException();
+    }
```
omkar-ethz: This should be && as well (see comment on update endpoint below)

omkar-ethz (approved):
LGTM!

src/casl/action.enum.ts
```diff
@@ -103,6 +103,22 @@ export enum Action {
...
+  DatablockDeleteEndpoint = "datablock_delete_endpoint",
+  // individual actions
+  DatablockCreateInstance = "datablock_create_instance",
+  DatablockReadInstance = "datablock_read_instance",
+  DatablockUpdateInstance = "datablock_update_instance",
+  // admin actions
+  DatablockReadAny = "datablock_read_any",
+  DatablockUpdateAny = "datablock_update_any",
+  DatablockDeleteAny = "datablock_delete_any",
+
```
Junjiequan, line 121: There's generic any permission for admin users on line 14, maybe that can be used instead of having more anys?
accessAny = "access_any",

fpotier: My understanding here is that `accessAny` would only replace `DatablockReadAny`, is that correct?

Junjiequan: My thoughts on `accessAny`is that `any` actions should be applied to admin only and it should not matter which CRUD operation.

Junjiequan: We can discuss it when we refactoring the permissions, this sdhould not be a blocker for the PR

sbliven (approved):
gltm

PR #2017, opened 2025-06-27
feat: refactor published data controller

martin-trajanovski:
## Description
Refactor of the published data controller and mapping of the published data fields to datacite.
**NOTE**: This is just a PR towards the main `published-data-refactor` branch. Once this is ready there will be a new PR that contains all the changes from the refactor towards the `master` branch

## Motivation
There was need for a change in the published data workflow.

## Fixes

* Bug fixed (#X)

## Changes:

* Changes are mostly related to the published data, its validation and mapping towards the datacite schema.

## Tests included

- [ ] Passing? 

## Documentation


src/casl/casl-ability.factory.ts
```diff
@@ -766,10 +766,11 @@ export class CaslAbilityFactory {
 
     if (
       user &&
-      user.currentGroups.some((g) => this.accessGroups?.delete.includes(g))
+      (user.currentGroups.some((g) => this.accessGroups?.delete.includes(g)) ||
+        user.currentGroups.some((g) => this.accessGroups?.admin.includes(g)))
     ) {
```
Junjiequan: Based on the previous permission managing logic, only delete groups users are allowed to delete anything.
Is this something has been discussed that allowing admin to delete

src/published-data/published-data.controller.ts
```diff
@@ -235,33 +302,240 @@ export class PublishedDataController {
...
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+  }
+
   // DELETE /publisheddata/:id
   @UseGuards(PoliciesGuard)
   @CheckPolicies("publisheddata", (ability: AppAbility) =>
-    ability.can(Action.Delete, PublishedData),
+    ability.can(Action.Update, PublishedData),
```
Junjiequan: why is it Action.Update here instead of Action.Delete?

src/published-data/published-data.controller.ts
```diff
@@ -235,33 +302,240 @@ export class PublishedDataController {
...
+        );
+      }
+    } else {
+      if (publishedData.status !== PublishedDataStatus.PRIVATE) {
+        throw new HttpException(
+          `Published data can only be resynced if it is in ${PublishedDataStatus.PRIVATE} state.`,
+          HttpStatus.BAD_REQUEST,
+        );
+      }
+    }
```
Junjiequan: I'd suggest keeping the permission logic in the `casl-ability.factory.ts` file to make it easy to maintain.
Currently we have two permission layers, endpoint and instance respectively. In this case, instead of check if `userIsAdmin` is true we can just pass the user to `const ability = this.caslAbilityFactory.publishedDataInstanceAccess(user);` which is reusable and easier to main, as we keep all the permission logic in the casl-ability factory

src/published-data/published-data.controller.ts
```diff
@@ -235,33 +302,240 @@ export class PublishedDataController {
...
+    ability.can(Action.Update, PublishedData),
+  )
+  @ApiResponse({
+    status: HttpStatus.OK,
+    type: PublishedData,
+    isArray: false,
+    description: "Return amended data with id specified",
+  })
+  @Post("/:id/amend")
+  async amend(
```
Junjiequan, line 443: Is there a specific reason to name it `amend` instead of `findOneAndUpdate`, which is a more common naming in general?

src/published-data/published-data.controller.ts
```diff
@@ -235,33 +302,240 @@ export class PublishedDataController {
...
+  @CheckPolicies("publisheddata", (ability: AppAbility) =>
+    ability.can(Action.Update, PublishedData),
+  )
+  @ApiResponse({
+    status: HttpStatus.OK,
+    type: PublishedData,
+    isArray: false,
+    description: "Return amended data with id specified",
+  })
+  @Post("/:id/amend")
```
Junjiequan, line 442: Is there any reason post is needed for update request here instead of patch or put

src/published-data/published-data.controller.ts
```diff
@@ -271,179 +545,86 @@ export class PublishedDataController {
...
-        }
-
-        console.warn(
-          "results not pushed to oaiProvider as oaiProviderRoute route is not specified in the env variables",
-        );
+    try {
+      await firstValueFrom(
+        this.httpService.request(registerDataciteDoiOptions),
+      );
+    } catch (err: any) {
```
Junjiequan: Try to avoid using `any`, unless it is just a proof of concept PR

src/published-data/published-data.service.ts
```diff
@@ -44,8 +45,17 @@ export class PublishedDataService {
     private readonly httpService: HttpService,
     @Inject(REQUEST)
     private request: Request,
+    private configService: ConfigService,
   ) {}
 
+  async getConfig(): Promise<Record<string, unknown> | null> {
+    const config =
+      this.configService.get<Record<string, unknown>>("publishedDataConfig") ||
```
Junjiequan, line 53: I might missed something here, I couldn't find anywhere the file being read. 
It seems to me the config variable will be assigned with just a string of `publishedDataConfig.json` instead of  actual object?

src/casl/casl-ability.factory.ts
martin-trajanovski: I was following the workflow that i got for the published data which was allowing the admins to delete published data but I double checked this with Nicolas and I think it's fine for now to only allow users within `DELETE_GROUPS`

src/published-data/published-data.service.ts
martin-trajanovski: You can see that it's being read here https://github.com/SciCatProject/scicat-backend-next/pull/2017/files#diff-8c7c24088419fc90f5f1bb6c6435c1b14bc21e2d61c3c7c3bf40e2e63c335caaR62

src/published-data/published-data.controller.ts
martin-trajanovski: As we are using POST for publishing and registering I think it would be nice to do it the same way here as well.

src/published-data/published-data.controller.ts
martin-trajanovski: As we have publish and register endpoints separately I think it is nice to have separate one for amend as well. That way I guess it will be easier to maintain the access and permissions.

src/published-data/published-data.controller.ts
martin-trajanovski: Yes you are right I will revert this one

src/published-data/published-data.controller.ts
martin-trajanovski: I will try to move these checks inside the casl factory itself and see if it can be successfully managed there. Thanks for the suggestion.

src/published-data/published-data.service.ts
Junjiequan: I think I wrote that function and completely forgot about it

Junjiequan (approved):

PR #2035, opened 2025-07-04
feat: published data refactor

martin-trajanovski:
## Description
This PR aims to refactor the published data following the new workflow shown in the picture.

## Motivation
The old published data workflow was outdated and wasn't using all the metadata fields that can be provided to DataCite

## Fixes

* Bug fixed (#X)

## Changes:

* Changes are mostly in the published data schema and controller
* There is JSON config file introduced which is used for the frontend form generation and validation of the published data.

## Tests included

- [x] Passing? 

## Documentation

published_data_workflow_1


src/published-data/schemas/published-data.schema.ts
```diff
@@ -12,7 +14,12 @@ export type PublishedDataDocument = PublishedData & Document;
...
-export class PublishedData {
+export class PublishedData extends QueryableClass {
+  @Prop({
+    type: String,
+  })
+  _id: string;
+
   @Prop({
     type: String,
     default: function genUUID(): string {
```
HayenNico, line 26: Minor fix needed: The default value here should use PID_PREFIX, not DOI_PREFIX

src/published-data/schemas/published-data.schema.ts
```diff
@@ -49,114 +56,45 @@ export class PublishedData {
   })
```
HayenNico, line 59: Default value of the DOI suffix currently is a uuidv4, which is quite long (DataCite recommends 6-10 character suffixes). Is there an easy way to use something shorter (for example: truncate the generated uuidv4)?

martin-trajanovski: I understand your concern and recommendations from DataCite. This was set to use the `uuid` for some time ago and I guess we can change it but at the same time maybe we need to think about data migration of all the existing published data and if that is a smart thing to do or not for already registered DOIs.

martin-trajanovski: @HayenNico I think I will introduce a new environment variable flag (something like DOI_UUID_SUFFIX) that will enable/disable 10 character suffix for the DOIs and can be configured by the site admin. If some sites want to continue with the `uuid` that will be still be possible.

HayenNico: I think that's a good compromise, but I'd want to raise the following question as well: Currently, the PublishedData._id and PublishedData.doi are two completely unrelated, randomly generated UUIDs. In cases where people want to continue to use uuid4 as their DOI suffix, would they not expect the suffix to match the _id field?

martin-trajanovski: As I have tested locally if you choose to use the uuid or even randomly generated 10 character string for the `PublishedData.doi` it always matches the `PublishedData._id` because we set the `_id` to be equal to the doi here https://github.com/SciCatProject/scicat-backend-next/blob/master/src/published-data/published-data.module.ts#L38

martin-trajanovski: FYI: I added `DOI_SHORT_SUFFIX` flag which if enabled with `true` value will make the app use shorter DOI suffix with randomly generated 10 characters. I added some description for it in the readme file as well.

src/published-data/published-data.controller.ts
```diff
@@ -253,24 +323,217 @@ export class PublishedDataController {
...
+    }
+
     return this.publishedDataService.update(
       { doi: id },
       updatePublishedDataDto,
     );
   }
 
+  // POST /publisheddata/:id/publish
+  @UseGuards(PoliciesGuard)
```
Junjiequan: `PoliciesGuard` allows unauthenticated users to access the endpoint.

`AuthenticatedPoliciesGuard` should be used here to prevent unauthenticated users from publishing

test/PublishedData.js
```diff
@@ -33,7 +38,7 @@ const nonpublictestdataset = {
   ownerGroup: "examplenonpublicgroup",
 };
 
-describe("1600: PublishedData: Test of access to published data", () => {
+describe.only("1600: PublishedData: Test of access to published data", () => {
```
Junjiequan: `only` needs to be removed

src/published-data/published-data.service.ts
```diff
@@ -44,8 +45,17 @@ export class PublishedDataService {
...
   ) {}
 
+  async getConfig(): Promise<Record<string, unknown> | null> {
+    const config =
+      this.configService.get<Record<string, unknown>>("publishedDataConfig") ||
+      null;
+
+    return config;
+  }
+
```
Junjiequan, line 58: It'd be better to generate one from example if no publishedDataConfig json is provided 
Currently it fails silently when no publishedDataConfig is provided

src/published-data/published-data.controller.ts
martin-trajanovski: I can update this by your suggestion but I think we already have:
`
@CheckPolicies("publisheddata", (ability: AppAbility) =>
    ability.can(Action.Update, PublishedData),
  )
 `
 after this line of code which will prevent unauthenticated users to update/publish

Junjiequan: Sorry i didn't explain it correctly. Either one of the decorator prevents user from accessing the endpoint, but `AuthenticatedPoliciesGuard ` prevents `@CheckPolicies` from being excuted. 
The idea is that, endpoints for authenticated users only, we dont need to check their permissions

README.md
```diff
@@ -148,14 +154,17 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
...
+| `DOI_PASSWORD` | string | | The facility DOI DataCite password. | |
 | `EXPRESS_SESSION_SECRET` | string | No | Secret used to set up express session. Required if using OIDC authentication | |
 | `EXPRESS_SESSION_STORE` | string | Yes | Where to store the express session. When "mongo" on mongo else in memory | |
 | `HTTP_MAX_REDIRECTS` | number | Yes | Max redirects for HTTP requests. | 5 |
 | `HTTP_TIMEOUT` | number | Yes | Timeout for HTTP requests in ms. | 5000 |
 | `JWT_SECRET` | string | | The secret for your JWT token, used for authorization. | |
 | `JWT_EXPIRES_IN` | number | Yes | How long, in seconds, the JWT token is valid. | 3600 |
 | `LDAP_URL` | string | Yes | The URL to your LDAP server. | |
-| `LDAP_BIND_DN` | string | Yes | Bind_DN for your LDAP server. | |
+| `LDAP_BIND_DN` | string | Yes | Bind*DN for your LDAP server. | |
```
nitrosx: `Bind*DN` should `Bind_DN`

README.md
```diff
@@ -166,7 +175,7 @@ Valid environment variables for the .env file. See [.env.example](/.env.example)
...
 | `OIDC_SUCCESS_URL` | string | Yes | SciCat Frontend auth-callback URL. Required to pass user credentials to SciCat Frontend after OIDC login. Example: https://myscicatfrontend/auth-callback. Must be `frontend-base-url/auth-callback` or `frontend-base-url/login` for the official SciCat frontend. | |
 | `OIDC_RETURN_URL` | string | Yes | The path segment within the SciCat Frontend to redirect to, passed as query param in `OIDC_SUCCESS_URL` and handled by frontend. Example: /datasets. | |
-| `OIDC_FRONTEND_CLIENTS` | string | Yes | Comma separated list of additional frontend OIDC clients for this backend. Example: scilog,maxiv. Their success and return URLs can be configured by setting `OIDC_${CLIENT}_SUCCESS_URL` (E.g. `OIDC_SCILOG_SUCCESS_URL`) and `OIDC_${CLIENT}\_RETURN_URL`| |
+| `OIDC_FRONTEND_CLIENTS` | string | Yes | Comma separated list of additional frontend OIDC clients for this backend. Example: scilog,maxiv. Their success and return URLs can be configured by setting `OIDC*${CLIENT}_SUCCESS_URL` (E.g. `OIDC_SCILOG_SUCCESS_URL`) and `OIDC_${CLIENT}\_RETURN_URL`| |
```
nitrosx: `OIDC*${CLIENT}` should be `OIDC_${CLIENT}`

migrations/20250521070910-published-data-migration.js
```diff
@@ -0,0 +1,106 @@
...
+          dataDescription: publishedData.dataDescription,
+          resourceType: publishedData.resourceType,
+          contributors:
+            publishedData.authors?.map((author) => ({
+              name: author.trim(),
+            })) || [],
+          relatedIdentifiers:
+            publishedData.relatedPublications?.map((relatedPublication) => ({
+              relatedIdentifier: relatedPublication,
+            })) || [],
```
nitrosx, line 31: Should we use the same metadata structure that we use in Datasets?
```
metadata: {
  affiliation: {
    value: publishedData.affiliation,
    humanName: ....
}
```

publishedDataConfig.example.json
nitrosx, line 1: Is this json file following json schema format?

src/published-data/published-data.controller.ts
```diff
@@ -483,80 +654,116 @@ export class PublishedDataController {
...
+    descriptions,
+    relatedItems,
+    relatedIdentifiers,
+    language,
+    dates,
+    sizes,
+    formats,
+    geoLocations,
+    fundingReferences,
+    landingPage,
```
nitrosx, line 728: Given that the metadata is arbitrary, in my opinion is too risky to assume that the fields that are needed to register a DOi are all there.

nitrosx (changes requested):
I left few comments.
Not sure if they will all lead to changes.

publishedDataConfig.example.json
martin-trajanovski: Yes the file is following the json schema format and it is example file that also contains all DataCite fields with correct requirements.

migrations/20250521070910-published-data-migration.js
martin-trajanovski: I don't think so as we are following a different structure here and the configuration provided should be json schema format to be able to validate it easy based on requirements from DataCite.

src/published-data/published-data.controller.ts
martin-trajanovski: Yes I understand your concern but in the register function before I call `doiRegistrationJSON` and build the final json sent to register a DOI there is validation on the metadata. This validation is based on the config that is provided and if all the required fields are there then you are able to register the DOI without any problem. I think this is enough for the first version and if there are some improvements or adjustments needed we can do them afterwards.

Junjiequan (approved):

Issue #1938, opened 2025-05-23
Implement Job/datasetDetails


sbliven:
## Summary

Backend-v3 provided endpoint `/api/v3/Jobs/datasetDetail` which provided information about a job and its datasets in one query. Similar functionality should be implemented in scicat-backend-next.

## Steps to Reproduce

We previously used the call `/api/v3/Jobs/datasetDetails?` with the following query params:
```js
 {
    jobId: id,
    datasetFields: JSON.stringify({
        pid: true, 
        sourceFolder: true, 
        sourceFolderHost: true,
        contactEmail: true, 
        owner: true, 
        ownerGroup: true, 
        classification: true, 
        type: true,
        datasetlifecycle: true,
        createdBy: true
    }),
    include: JSON.stringify({relation: "datablocks"}),
    includeFields: JSON.stringify({
        id: true,
        archiveId: true,
        size: true,
        datasetId: true
    })
}
```
## Current Behaviour

The `GET /api/v4/jobs` endpoint doesn't support any query parameters

## Expected Behaviour

It would be nice to have a `/api/v3/Jobs/datasetDetails` endpoint that replicated the old functionality (which is not well documented, so this would require looking at the old js code).

For v4 we could implement it similar to the `/api/v4/datasets` endpoint, which accepts `include`, and `fields` filters.

sbliven:
@nitrosx thought that the old /datasetDetails was overly specific to loopback. A better way to implement would be to follow the example of the filters for `GET /dataset` and use the "include" and "fields" properties to join the job to the dataset/datablocks.

Example (rough; we need to look up the correct syntax to pass to mongoose):
```
GET /jobs/:id
{
  "include": ["datasets","datablocks"],
  "fields": ["id","jobParams","datasets.pid","datablocks.id"]
}
```

This would require that 'datasets' and 'datablocks' be added as optional fields in the `JobOutputDto`.

sbliven:
I think we should implement `/api/v3/Jobs/datasetDetails` for backwards compatibility and an `include` field for `/api/v4/jobs`.

PR #2063, opened 2025-07-17
feat(Jobs): implements include filters for jobs

sofyalaski:
## Description
This PR is refactoring filters that can be provided for Jobs system. 
Previously, only `where`, `fields` and `limits` filters could be specified. This PR allows for an additional `include` filter, which would merge jobs with datasets collection based on ID. 
Additionally, another endpoint datasetDetails was introduced that does a merge with datasets collection and  a nested merge with Datablocks, Origdatablocks and Attachment collections. 

* When a simple `GET /api/v4/Jobs` endpoint is used with an `include: ["datasets"]` query, the output job will include a property "datasets" in addition to the ones defined in the `OutputJobDto`
* When  `GET /api/v4/Jobs/datasetDetails` endpoint is used, `include` filter is forbidden as it is hardcoded to include Datasets, Datablocks, Origdatablocks and Attachment collection. The output job will include a "datasetDetails" property ( instead of "datasets" in the previously explained endpoint). The value of this property is an object that includes dataset keys, datablocks, origdatablocks and attachments.

## Motivation
In legacy backend, [there was an endpoint](https://github.com/SciCatProject/scicat-backend-next/issues/1938) - `jobs/datasetDetails` that was doing a merge with datasets and a consequent one with datablocks. Similar functionality is missing in a new backend, both V3 and V4 APIs
## Fixes

A few fixes to the similar logic that merges Datasets with other collections existed already. 

* in `src/datasets/datasets-access.service.ts` when evaluating casl permissions for Datablocks, actions were evaluated based on `Datablock` class, while in `casl-factory.ts` it actually uses `DatasetClass`
 

## Changes:
in `src/datasets/datasets-access.service.ts` a new function `addDatasetAccess` was added that would evaluate user's access to the dataset itself. 

**job-lookup.ts**
* Introduces the main pipeline to join with datasets. For the "datasets" `$lookup` an empty pipeline array is defined. For the "datsetDetails" on the contrary, `$lookup` pipeline includes nested `$lookup`s with the predefined keys for datablocks, origdatablocks and attachments. 

**output-job-v4.dto.ts**
* defines `OutputJobDto` class, which is synonymous to an output of a `JobClass`.
* defines `PartialOutputDto` class, that doesn't require all the fields from `OutputJobDto` but includes new arrays of datasets and datasetDetails, which are both not required.
* defines `PartialIntermediateOutputDto` class, that extends `OutputJobDto` but requires such properties from `JobClass` as `_id`, `id`, `type`, `ownerGroup` and `ownerUser`. These fields are required in any instance of a job to evaluate casl permissions.

**jobs.controller.ts**
* new endpoint `datasetDetails` with three query options is introduced it reuses the `GET api/v4/datsets/:id` endpoint with refactored filters and then regroups the output.

**jobs.controller.utils.ts**
* `isFilterValid` was removed and substituted by pipes.
* `generateJobInstanceForPermissions` can now accept a type union of both `JobClass` or `PartialIntermediateOutputDto`, as the latter is guaranteed to have the necessary properties.
* functions related to getting jobs expect an instance of either `JobClass` or `PartialOutputDto`.
* `getJobById` was refactored into two small functions, the block that was evaluating casl permissions on job instance was separated into `getOneJob` function in the same file. `getJobById` is now only required by `/api/v3`
* `getJobByQuery` makes a call to job service to find a job based on the query and then on top casl permission are evaluated by calling `getOneJob`. 
* `removeFields` removes fields that were added for casl properties, unless they were not specifically required in request.

**jobs.service.ts** 
* now requires `DatasetsAccessService` in constructor
* `findJobComplete` implements an aggregation pipeline to merge jobs with datasets collection. It first introduces a new fields `datasetIds` as these are present in array structure inside `jobPrams:{datasetList[]}`
* `addLookupFields` that dynamically adds another join operations to the aggregation pipeline for collections that
can be joined with datasets.

**jobs.v4.controller.ts** 
* mainly refactors the swaggerUI interface inputs 
* refactors  Fullquery endpoint `/api/v4`. This part will be reverted and introduced in a new PR instead, as it might break existing pipelines. 

**include-validation.pipe.ts**
implements a check for the `include` filters, it cannot use the one from `common`, as here more restrictions are required

**jobs-filter-content.ts**
* templates for swaggerUI

## Tests included

In JobsV3 introduce a few tests. 
The rest will be updated
- [x] Passing? 

## Documentation


Junjiequan:
`"0202: should fetch different dataset if skip is used in limits filter"` test error will be resolved in #2065

src/datasets/datasets-access.service.ts
```diff
@@ -66,19 +65,19 @@ export class DatasetsAccessService {
...
         const canViewOwner = ability.can(
           Action.DatasetDatablockReadOwner,
-          Datablock,
+          DatasetClass,
         );
         const canViewPublic = ability.can(
           Action.DatasetDatablockReadPublic,
-          Datablock,
+          DatasetClass,
         );
```
HayenNico, line 81: These checks for the combined abilities DatasetDatablockRead should have Datablock passed in, no? And these combined abilities will be obsolete in API v4 and should be replaced with the general DatablockRead abilities

src/datasets/datasets-access.service.ts
```diff
@@ -185,4 +182,48 @@ export class DatasetsAccessService {
...
+
+  addDatasetAccess(fieldValue: PipelineStage.Lookup) {
+    const currentUser = this.request.user as JWTUser;
+    const ability = this.caslAbilityFactory.datasetInstanceAccess(currentUser);
+    const canViewAny = ability.can(Action.DatasetReadAny, DatasetClass);
+    const canViewAccess = ability.can(
+      Action.DatasetReadManyAccess,
+      DatasetClass,
+    );
+    const canViewOwner = ability.can(Action.DatasetReadManyOwner, DatasetClass);
```
HayenNico, line 194: DatasetReadManyAccess and DatasetReadManyOwner should require an instance to be passed instead of DatasetClass

src/jobs/dto/output-job-v4.dto.ts
```diff
@@ -0,0 +1,147 @@
...
+  datablocks: Datablock[];
+
+  @ApiProperty({ required: false })
+  @IsArray()
+  attachments: Attachment[];
+
+  @ApiProperty({ required: false })
+  @IsArray()
+  samples: SampleClass[];
+}
```
HayenNico: Need some clarification: What other resources can a job refer to directly? This links to #2068 - if a job has a dataset list, then only datasets should be valid for lookup, since chain lookups don't work

src/jobs/jobs.controller.ts
```diff
@@ -218,6 +224,135 @@ export class JobsController {
...
+      typeof parsedInclude === "object" &&
+      "relation" in parsedInclude &&
+      typeof parsedInclude.relation === "string"
+    ) {
+      includeArray.push(`datasets.${parsedInclude.relation}`);
+      for (const [k, v] of Object.entries(parsedIncludeFields ?? {})) {
+        if (v) {
+          fieldArray.push(`datasets.${parsedInclude.relation}.${k}`);
+        }
+      }
```
HayenNico: This seems to be the chained lookup Job->Dataset->Datablock,Proposal,... that will cause issues

src/jobs/types/job-lookup.ts
```diff
@@ -0,0 +1,39 @@
+import { PipelineStage } from "mongoose";
+
+export enum JobLookupKeysEnum {
+  datasets = "datasets",
```
HayenNico, line 5: I think `all = undefined` should be a valid lookup key for any collection in API v4, in this case it would just be the same as providing `datasets`.

src/jobs/jobs.v4.controller.ts
```diff
@@ -136,33 +137,38 @@ export class JobsV4Controller {
...
     description:
-      "Define further query parameters like skip, limit, order.\n" +
-      fullQueryDescriptionLimits,
+      "Database filters to apply when retrieving jobs. Use dot-delimited paths (e.g., 'datasets.attachments') to specify fields from related collections.",
     required: false,
     type: String,
-    example: fullQueryExampleLimits,
+    content: getSwaggerJobFilterContent({
+      where: true,
+      include: false,
```
HayenNico, line 148: Shouldn't include be true here if it's intended as a valid filter field?

HayenNico (commented):
Overall, I think we need to consider if allowing more complex include filters like `datasets.datablocks` makes sense. Datasets API v4 was not designed with this in mind, since there are no relations that would require more than one lookup from there. If this is how we want to handle include filters, the same functionality should be available for all API v4 controllers.

Is there a specific use case that requires the backend to return one giant aggregate document of, for example, all Datablocks for all Datasets in a Job? I imagine in most cases it would be easier to send additional queries to the backend instead.

src/jobs/dto/output-job-v4.dto.ts
sbliven: What's the reason for restricting chaining? This would be useful. Our archiver currently uses `/Jobs/datasetDetails?include={relation: "datablocks"}` (in v3) to get all datablocks in one request. We're using a workflow system that supports individual REST calls but not loops (or at least makes implementing loops harder), so it would be nice to also be able to aggregate all datablocks from a single job call.

HayenNico: I'm not against allowing chaining in principle, but that will require changes to the way API v4 is implemented. Datasets v4 doesn't include this functionality since there are no collections more than one relation step away from a Dataset. In #1953 we made the design decision not to allow chained lookups coming from OrigDatablocks (no use case for that), and refactored the validation pipes (`FilterValidationPipe` and `IncludeValidationPipe`) accordingly to the common module.

The logic in the common module versions could be changed to accommodate chaining, which would make it available for all collections in API v4. Would that be your preferred approach?

src/datasets/datasets-access.service.ts
sofyalaski: Yes, I agree that they should. 
In casl the permissions for the datablock actions were written for the `DatasetClass`. I assumed that the permissions model for datablocks is the same as for datasets.

src/datasets/datasets-access.service.ts
sofyalaski: The idea here is to create the query, on the `DatasetClass` type of subject. The query is created in the next lines. We use a similar approach in [`datasets.v4.controller.ts`](https://github.com/SciCatProject/scicat-backend-next/blob/771acc2706979bb5c58a66a7f5263c82de4bf011/src/datasets/datasets.v4.controller.ts#L216)

src/jobs/dto/output-job-v4.dto.ts
sofyalaski: @HayenNico, answering your first question, for now, it's really just the datasets that can be accessed directly from the job.
To the general problem with the chaining: you're right that there are no other collections that can be merged consequently, but is that really an issue? What other changes to the v4 api other than the filter and include pipes would it require, you think? Because the structure of the `DATASET_LOOKUP_FIELDS` instance offers a very convinient way to chain them.

src/jobs/jobs.v4.controller.ts
sofyalaski: This one is a bit confusing. 
Per se, this endpoint became redundant with the normal GET `/api/v4/jobs`. The `fullquery`in jobs endpoint has a similar functionality as `/api/v3/datasets/fullquery` has for datasets ( where "include" isn't part of the filter ). In v4 this functionality is substituted via the GET `/api/v4/datasets`. This PR implements proper filtering for jobs in v4, that's why the endpoint becomes redundant. We just didn't want to delete it straight away because this might be breaking change.

src/jobs/types/job-lookup.ts
sofyalaski: yes, I can add that

src/datasets/datasets-access.service.ts
HayenNico: The model should be the same, depending on the current implementation it may still be the case that access to a Dataset directly implies access to all attached Datablocks.

src/datasets/datasets-access.service.ts
HayenNico: What I mean here is that `ability.can(Action.DatasetReadManyOwner, DatasetClass)` (and equivalent for access) will always evaluate to `false` when `DatasetClass` is passed in as the second argument. The second argument must be an instance of `DatasetClass` with `ownerGroup` and `accessGroups` set (matching the requested dataset), so the service can check if the current user's `currentGroups` include either

sofyalaski: In the [docs](https://casl.js.org/v6/en/guide/intro) under the Basics point "2. Subject", they say about the second argument  "The subject or subject type which you want to check user action on." And in this case I pass the subject type, which is a valid argument. Then the authorization happens in the casl ability factory.
If we search through the codebase of casl factory by `Action.DatasetReadManyAccess`, all the hits have a form like this: `can(Action.DatasetReadManyAccess, DatasetClass);` - no conditional filtering on the instance of the Dataset class. So the `const canViewAccess` can evaluate to true. 

But you're right about the `Action.DatasetReadManyOwner`, there's no rule for anything with this action type in casl!

HayenNico: Sounds right to me, and I can see plenty of examples that match that description.
What confuses me is that in the dataset controller (v3 and v4) there's extra logic for the same type of operation to make sure an instance is passed instead of the class:
[https://github.com/SciCatProject/scicat-backend-next/blob/b96dd654edecfa5e85acfbeb9e7c9470021f5110/src/datasets/datasets.controller.ts#L263](https://github.com/SciCatProject/scicat-backend-next/blob/b96dd654edecfa5e85acfbeb9e7c9470021f5110/src/datasets/datasets.controller.ts#L263)
[https://github.com/SciCatProject/scicat-backend-next/blob/b96dd654edecfa5e85acfbeb9e7c9470021f5110/src/datasets/datasets.controller.ts#L392](https://github.com/SciCatProject/scicat-backend-next/blob/b96dd654edecfa5e85acfbeb9e7c9470021f5110/src/datasets/datasets.controller.ts#L392)

I think we need to clarify why that logic was put in place and if it's necessary here. @nitrosx Can you help?

sofyalaski: I am not sure if I understood it correctly, are you asking why we sometimes pass the DatasetClass and sometimes an instance of that class? Or why we wrap the `dataset` in `generateDatasetInstanceForPermissions` to explicitly copy it?

src/jobs/dto/output-job-v4.dto.ts
```diff
@@ -0,0 +1,147 @@
...
+} from "class-validator";
+import { PartialOutputDatasetDto } from "src/datasets/dto/output-dataset.dto";
+import { Instrument } from "src/instruments/schemas/instrument.schema";
+import { ProposalClass } from "src/proposals/schemas/proposal.schema";
+import { OrigDatablock } from "src/origdatablocks/schemas/origdatablock.schema";
+import { Datablock } from "src/datablocks/schemas/datablock.schema";
+import { Attachment } from "src/attachments/schemas/attachment.schema";
+import { SampleClass } from "src/samples/schemas/sample.schema";
+
+export class OutputJobDto extends CreateJobDto {
```
despadam, line 12: Based on https://github.com/SciCatProject/scicat-backend-next/issues/1590, the `@ApiProperty` could be omitted.

src/jobs/dto/output-job-v4.dto.ts
```diff
@@ -0,0 +1,147 @@
...
+  @ApiProperty({
+    type: Date,
+    required: true,
+    description: "Date/time when the job was last updated. ISO 8601 format.",
+  })
+  @IsDateString()
+  updatedAt: Date;
+}
+
+export class PartialOutputJobDto extends PartialType(OutputJobDto) {
```
despadam, line 70: Since it is extending the output, I do not think that the name `PartialOutputJobDto` is representative

sofyalaski: Since it has the "fields" filter it also can reduce the JobClass in output, and that naming is following the convention with other systems, but I get your point

src/datasets/datasets-access.service.ts
HayenNico: To document what we went over with Max, Matin and Jay:
My question was exactly about this inconsistency between passing a class or an instance to casl. It turns out that the `DatasetReadMany` action types don't actually do any instance checks (no conditions defined in casl-ability-factory), so passing the `DatasetClass` is completely equivalent.
But as I understand this also means that these checks don't fulfill their intended purpose either and are essentially redundant with the endpoint check if the user has `DatasetRead` access. Since `addAccessBasedFilters` is only ever called after the user already passed the endpoint access check, `canViewAccess` will always be true (`canViewOwner` is always false because of the missing rule definition in casl, but is also never reached).
For this PR this can stay as is, a change here would belong in a separate PR

sofyalaski:
After the discussion with SciCat team on August 13th, a decision was made to use a hardcoded pipeline for nested lookups instead of adding them dynamically. 
This might add additional cost to the operation, as it now always joins with three collections ( For instance at PSI, they only require datablocks but would still get all three - datablocks, origdatablocks and attachments). 
If facilities at a later time point argue that this adds too much operational cost, SciCat team should rethink the dynamic adding of nested lookups to the pipeline

despadam (approved):
Overall looks good to me. This was a lot of work so thank you @sofyalaski 🙏 

There are a few remaining topics I would like to bring up:
1. There is this `.finished` file included in the PR, could it be removed?
2. Since the tests were updated, the documentation needs to be updated as well.
3. Is there a reason why we need to keep the `@ApiProperty` in `OutputJobDto`? Apart from the `OutputJobV3Dto`, all other job-related DTOs do not have it anymore. Is it needed for the output ones or did we just forget to update those?

despadam:
Hi @sofyalaski , thanks for the latest changes.

The description from `ApiProperty` should be replaced with a docstring - see [create-job.dto.ts](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/dto/create-job.dto.ts) for example. Since we are doing that for the v4 output, would it be possible to also similarly update [output-job-v3.dto.ts](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/dto/output-job-v3.dto.ts) for consistency reasons? 

Also, what is the purpose of the `.finished` file? Could it be removed? 

Thanks

sofyalaski:
@despadam, thanks for the comments, 

I have updated DTO as required. Thanks for noticing. 
`.finished` comes from scicatlive, it shouldn't be here, I made a mistake by adding it. 

As for the documentation, I will prepare a separate PR in the next few days.

src/jobs/jobs.controller.utils.ts
```diff
@@ -842,30 +845,77 @@ export class JobsControllerUtils {
...
+  ): Promise<PartialOutputJobDto | null> {
+    const currentJob = await this.jobsService.findOne({ _id: id });
+    return await this.getOneJob(
+      request,
+      currentJob as unknown as PartialIntermediateOutputJobDto,
+    );
+  }
+
+  /**
+   * Remove fields added to the job to evaluate casel permission if they are not present in fields
```
martin-trajanovski: Just a small typo here in `casel` -> `casl`

test/JobsGet.js
```diff
@@ -0,0 +1,749 @@
...
+        ]);
+        j1ds1.datablocks.should.be.an("array").to.have.lengthOf(2);
+        j1ds1.datablocks
+          .map((db) => db._id)
+          .should.include.members([datablockId1, datablockId2]);
+        j1ds1.origdatablocks.should.be.an("array").to.have.lengthOf(1);
+        const [origdatablock] = j1ds1.origdatablocks;
+        origdatablock.should.have.property("_id");
+        origdatablock._id.should.be.eq(origDatablock1);
+        // j1ds1.attachments.should.be.an("array").to.have.lengthOf(0);
```
martin-trajanovski: If not needed these commented lines of code should be removed.

martin-trajanovski (approved):
Just a minor cleanup and changes needed. Otherwise it looks good. Great job Sofya and it's quite good test coverage as well. 💯

test/JobsGet.js
sofyalaski: Thank you, I enabled this check as well.

PR #2080, opened 2025-07-24
feat: add global interceptor to recursively mask emails

minottic:
## Description

We have the requirement to hide emails for GDPR. This PR recursively removes emails from all fields (as well as orcid) from HTTP responses

## Changes:

* global interceptor to recursively mask emails
* activated only when env is set
* email is not masked when === user's email

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/common/interceptors/mask-sensitive-data.interceptor.ts
+++ b/src/common/interceptors/mask-sensitive-data.interceptor.ts
@@ -12,7 +12,6 @@ import { map, Observable } from "rxjs";
 
 @Injectable()
 class MaskSensitiveDataInterceptor implements NestInterceptor {
-
   private maskSensitiveData<T>(data: T, ownEmail: string): T {
     if (Array.isArray(data)) {
       let allArrays = true;
@@ -29,9 +28,7 @@ class MaskSensitiveDataInterceptor implements NestInterceptor {
     if (!this.isPlainObject(data)) return data as T;
     const result: Record<string, unknown> = {};
     for (const [key, value] of Object.entries(data)) {
-      if (
-        this.isToMaskEmail(value, ownEmail)
-      ) {
+      if (this.isToMaskEmail(value, ownEmail)) {
         result[key] = this.maskValue();
         continue;
       }
--- a/src/common/interceptors/mask-sensitive-data.interceptor.ts
+++ b/src/common/interceptors/mask-sensitive-data.interceptor.ts
@@ -1,0 +1,99 @@
+import {
+  CallHandler,
+  ExecutionContext,
+  Injectable,
+  Module,
+  NestInterceptor,
+} from "@nestjs/common";
+import { APP_INTERCEPTOR } from "@nestjs/core";
+import { instanceToPlain } from "class-transformer";
+import { isEmail } from "class-validator";
+import { map, Observable } from "rxjs";
+
+@Injectable()
+class MaskSensitiveDataInterceptor implements NestInterceptor {
+  private readonly keysToMask = ["orcidOfOwner"];
+
+  private maskSensitiveData<T>(data: T, ownEmail: string): T {
+    if (Array.isArray(data)) {
+      let allArrays = true;
+      const maskedData = data.map((item) => {
+        if (this.isToMaskEmail(item, ownEmail)) {
+          return this.maskValue();
+        }
+        allArrays = false;
+        return this.maskSensitiveData(item, ownEmail);
+      });
+      return ((allArrays && [...new Set(maskedData)]) || maskedData) as T;
+    }
+
+    if (!this.isPlainObject(data)) return data as T;
+    const result: Record<string, unknown> = {};
+    for (const [key, value] of Object.entries(data)) {
+      if (
+        this.keysToMask.includes(key) ||
+        this.isToMaskEmail(value, ownEmail)
+      ) {
+        result[key] = this.maskValue();
+        continue;
+      }
+      result[key] = this.maskSensitiveData(value, ownEmail);
+    }
+    return result as T;
+  }
+
+  private isPlainObject(input: unknown): input is Record<string, unknown> {
+    return (
+      typeof input === "object" &&
+      input !== null &&
+      !Array.isArray(input) &&
+      Object.getPrototypeOf(input) === Object.prototype
+    );
+  }
+
+  private isToMaskEmail(value: string | unknown, ownEmail: string) {
+    return typeof value === "string" && isEmail(value) && value !== ownEmail;
+  }
+
+  private maskValue(): string {
+    return "";
+  }
+
+  private toPlain<T extends { toObject?: () => object }>(value: T): T {
+    if (Array.isArray(value)) {
+      return value.map(this.toPlain) as unknown as T;
+    }
+
+    if (value && typeof value === "object") {
+      if (typeof value.toObject === "function") {
+        return value.toObject() as T;
+      }
+
+      try {
+        return instanceToPlain(value) as T;
+      } catch {
+        return value;
+      }
+    }
+
+    return value;
+  }
+
+  intercept(context: ExecutionContext, next: CallHandler): Observable<unknown> {
+    const request = context.switchToHttp().getRequest();
+    const ownEmail = request.user?.email;
+    return next.handle().pipe(
+      map((data) => {
+        const plainData = this.toPlain(data);
+        return this.maskSensitiveData(plainData, ownEmail);
+      }),
+    );
+  }
+}
+
+@Module({
+  providers: [
+    { provide: APP_INTERCEPTOR, useClass: MaskSensitiveDataInterceptor },
+  ],
+})
+export class MaskSensitiveDataInterceptorModule {}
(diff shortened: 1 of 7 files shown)


nitrosx:
@minottic I assume that you need to hide personal information in the data returned in by the BE and not only in the FE, correct?
With this solution, will you be masking personal information for everybody? Also for admins?
We would like to hide some of the personal information to non authenticated users, but show them to authenticated users and/or admins. Is that also your use case?

test/jest-e2e-tests/accessGroups.e2e-spec.ts
nitrosx, line 1: Are we planning to move to Jest tests from mocha?

nitrosx: ...also why saving them in a nested folder?

minottic:
this is masking for everyone, also for admins as we don't have needs now to make them available to admins, and we want to apply it to BE so also HTTP calls are hidden. The interceptor can be easily modified though to only mask when no auth

src/common/interceptors/mask-sensitive-data.interceptor.ts
```diff
@@ -0,0 +1,99 @@
...
+    );
+  }
+
+  private isToMaskEmail(value: string, ownEmail: string) {
+    return isEmail(value) && value !== ownEmail;
+  }
+
+  private maskValue(): string {
+    return "";
+  }
```
Junjiequan: Is it allowed to return partial information under the regulation instead of just returning empty string? e.g, `ju*****@ess.eu` 
if it's not allowed then maybe we should respond with some unique values like `Not available` so that FE knows it's a processed data

minottic: > Is it allowed to return partial information under the regulation instead of just returning empty string? e.g, ju*****@ess.eu

I have to admit I also thought about it but I am unsure what the value is when returnig it (for example keeping the original domain)

> if it's not allowed then maybe we should respond with some unique values like Not available so that FE knows it's a processed data

good point, thanks! Any suggestion on what the hardcoded value should be?

minottic:
while discussing internally it was raised that maybe the orcid does not need to be masked as the user can choose to hide emails once on orcid. Thus, do you have use cases to hide anything else than emails? If not, maybe we should simplify the logic (I can do that) and only mask emails, without grepping for additional fields (orcid)

src/common/interceptors/mask-sensitive-data.interceptor.ts
Junjiequan: I don't have a strong opinion on the value. Something like `gdpr_hidden`,  `not available(gdpr)` would work maybe?

minottic: what if I use `*****`?

nitrosx: Here my 2 cents: if the GDPR allows it, we should just replace all the letters and number with character *.
Examples
- if my email is first.last@mycompany.com, we show `*****.****@*********.***`
- in case we are masking an ORCID, from 0000-0003-2818-0368 we show `****-****-****-****`

nitrosx:
Let's address your use case, and expand the functionalities later.
Either way works for me, meaning we can just add email or email and ORCID.

src/common/interceptors/mask-sensitive-data.interceptor.ts
```diff
@@ -0,0 +1,99 @@
...
+        return this.maskSensitiveData(item, ownEmail);
+      });
+      return ((allArrays && [...new Set(maskedData)]) || maskedData) as T;
+    }
+
+    if (!this.isPlainObject(data)) return data as T;
+    const result: Record<string, unknown> = {};
+    for (const [key, value] of Object.entries(data)) {
+      if (
+        this.keysToMask.includes(key) ||
```
nitrosx: In this function, you are masking the ORCID only by targeting the field, but you are masking any email no matter where is found.
This does not cover the case where the ORCID is saved in the description or any other field. In such case, the ORCID will be returned to the user.
I think you should change the ORCID masking to follow the same principle as the email.

test/jest-e2e-tests/accessGroups.e2e-spec.ts
nitrosx, line 1: Can we transfer this file in the main test folder? ...or should we address this in a separate PR?

test/jest-e2e-tests/datasets.e2e-spec.ts
nitrosx, line 1: Can we transfer this file in the main test folder? ...or should we address this in a separate PR?

test/jest-e2e-tests/oidc.e2e-spec.ts
nitrosx, line 1: Same comment as above

test/jest-e2e-tests/utlis.ts
nitrosx, line 1: Same comment as above

nitrosx (changes requested):

test/jest-e2e-tests/accessGroups.e2e-spec.ts
minottic: dunno, this was not implemented by me...

nitrosx: Let's save it for a future PR.

test/jest-e2e-tests/oidc.e2e-spec.ts
nitrosx: Future PR

test/jest-e2e-tests/utlis.ts
nitrosx: Future PR

test/jest-e2e-tests/datasets.e2e-spec.ts
nitrosx: Future PR

minottic: what I do here is set an env var different from the ones used in tests and run a self contained app. Since the tests in `tests` assume an app is running (with preset  envs), I could not do that in an easy manner. I could have had a .spec.ts file beside the dataset controller, but what I understood from the repo structure is that those files should contain unit tests only. I actually think this is NestCreateApp that is used here is a better testing strategy, as one has all capabilities to mock particular method (e.g. the oidcguard in the other test below, which is not possible with the tests in the `test` folder), but I simpatise with keeping both given that many were reused from the legacy be

test/jest-e2e-tests/oidc.e2e-spec.ts
minottic: I cannot do that because of the OidCMock

test/jest-e2e-tests/accessGroups.e2e-spec.ts
minottic: see comments below. I dunno why they are in that folder and using that framework, I added the dataset one there  (and the oidc one in an older PR) as they are not unit tests (which should go in *.spec.ts from what I understood), neither tests I can add directly as tests under `tests` because they need some mocking (see comments below)

src/common/interceptors/mask-sensitive-data.interceptor.ts
minottic: that's a valid point, but I am not sure there's a way to "understand" if a format follows the ORCID one. Let me ask internally if we really need to mask this, maybe we don't and we limit this interceptor to hide emails for now

minottic:
> Here my 2 cents: if the GDPR allows it, we should just replace all the letters and number with character *.
Examples

> if my email is [first.last@mycompany.com](mailto:first.last@mycompany.com), we show *****.****@*********.***
in case we are masking an ORCID, from 0000-0003-2818-0368 we show ****-****-****-****

I think this is a good suggestion, but I am unsure what we can profit from it. I understand Jay's point of having a way to make the FE understand if they were masked, but why adding this extra logic of having different masks per type? What do clients gain from it?

src/common/interceptors/mask-sensitive-data.interceptor.ts
Junjiequan: > I have to admit I also thought about it but I am unsure what the value is when returnig it (for example keeping the original domain)

Both `***@***` or `*****` looks fine for me

minottic: > Here my 2 cents: if the GDPR allows it, we should just replace all the letters and number with character *.
> Examples
> 
> if my email is [first.last@mycompany.com](mailto:first.last@mycompany.com), we show .@.***
> in case we are masking an ORCID, from 0000-0003-2818-0368 we show ---
> 
> 

I think this is a good suggestion, but I am unsure what we can profit from it. I understand Jay's point of having a way to make the FE understand if they were masked, but why adding this extra logic of having different masks per type? What do clients gain from it? This said, it would be a fairly easy change if we decide so

minottic:
I have run this internally and it's enough to mask emails (thus I have removed the orcid part). I have used `*****` whenever a value is masked. I think having a single "masking" value, no matter the field, will be easier for the FE

Junjiequan (approved):
looks fine to me

minottic:
@nitrosx can we merge this?

minottic:
@nitrosx any reason why this is blocked?

nitrosx (approved):

PR #2092, opened 2025-07-31
fix: encode metadata keys

abdimo101:
## Description
In `datasets.service.ts` the metadata keys are now encoded before datasets gets created and updated. 

## Motivation
To ensure correct formatting of metadata key queries for both Elastic Search and MongoDB. 

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [x] Passing? 

## Documentation


Junjiequan:
Can we create 2 utility functions for encode and decode that handles dot ? 
Maybe call it like `encodeURIComponentExtended` or `decodeURIComponentExtended`
so that if we need to add more special characters in the future we don't have to change them from different places

src/datasets/datasets.service.ts
```diff
@@ -95,16 +97,39 @@ export class DatasetsService {
...
+    const encoded: Record<string, unknown> = {};
+
+    Object.entries(metadata).forEach(([key, value]) => {
+      const decodedKey = decodeURIComponentExtended(key);
+      const encodedKey =
+        decodedKey === key ? encodeURIComponentExtended(key) : key;
+      encoded[encodedKey] = value;
+    });
+    return encoded;
+  }
```
nitrosx: Is this function recursive?
SciCat allows for nested metadata and we nee to  be able to encode metadata keys when metadata is nested.

nitrosx: I'm not sure to understand why we are using both encode and decode functions in this helper.
I think that line 107 and 108 can be rewritten as:
```
      const encodedKey = encodeURIComponentExtended(key);
```
Can you clarify why you  decode the metadata key that you receive from the user?

src/datasets/datasets.service.ts
```diff
@@ -406,6 +438,11 @@ export class DatasetsService {
       throw new NotFoundException(error);
     }
   }
+
+  decodeMetadataKeys(keys: string[]): string[] {
+    return keys.map((key) => decodeURIComponent(key.replace(/%2e/g, ".")));
```
nitrosx: Please use `decodeURIComponentExtended` function

nitrosx (changes requested):

nitrosx:
We should also add testing for this feature

src/datasets/datasets.service.ts
abdimo101: Currently the function is not recursive, but I will update it to handle nested metadata soon. About the decode before encoding is because without decoding, if a user renames a field that has already been encoded, it would result in it being encoded twice or more

nitrosx: Apologies but it is still not clear to me... must be the jet-leg.
Can you elaborate further on the decoding and provide an example?

abdimo101: Yes so for example, if a field "experiment name" is already encoded as "experiment%20name" and the user renames it, encoding it again without decoding would result in "experiment%25%20name". So decoding first normalizes the field name back to "experiment name", ensuring it is only encoded once each time. Let me know if you need a better explanation

migrations/20250908132500-encode-metadatakeys-datasets.js
```diff
@@ -0,0 +1,46 @@
...
+        await db
+          .collection("Dataset")
+          .updateOne(
+            { _id: dataset._id },
+            { $set: { scientificMetadata: encodedMetadata } },
+          );
+      });
+  },
+
+  async down(db, client) {},
```
minottic: is the down migration missing on purpose?

test/RawDataset.js
```diff
@@ -253,7 +253,7 @@ describe("1900: RawDataset: Raw Datasets", () => {
           res.body.should.have
             .property("pid")
             .and.equal(decodeURIComponent(pid));
-          isEqualWithAny(res.body, TestData.RawCorrectGet).should.be.true;
+          //isEqualWithAny(res.body, TestData.RawCorrectGet).should.be.true; // this line would fail the test because the metadata keys are now encoded which would no longer match the TestData.RawCorrectGet
```
minottic: I would discourage removing the test... is there a way to make it pass and modify the value it is checked against?

migrations/20250908132500-encode-metadatakeys-datasets.js
```diff
@@ -0,0 +1,46 @@
+module.exports = {
+  async up(db, client) {
+    function encodeKey(key) {
+      return encodeURIComponent(key).replace(/\./g, "%2E");
+    }
+
+    function encodeMetadataKeys(metadata) {
```
minottic: I think you can import functions from the node app here, see the history migration

minottic (changes requested):
LGTM, a part from a couple of comments, especially the commented test I would change.

Would it make sense to decode the keys in the response returned from the API?

migrations/20250908132500-encode-metadatakeys-datasets.js
Junjiequan: Good point. we should have a rollback for this breaking change

abdimo101: down migration added in [1a2ca44](https://github.com/SciCatProject/scicat-backend-next/pull/2092/commits/1a2ca44805b2955b348195c4762fb998a2e21ef6)

test/RawDataset.js
abdimo101: the test has been uncommented since it now returns the decoded metadata keys in [1a2ca44](https://github.com/SciCatProject/scicat-backend-next/pull/2092/commits/1a2ca44805b2955b348195c4762fb998a2e21ef6)

migrations/20250908132500-encode-metadatakeys-datasets.js
abdimo101: Fixed in [1a2ca44](https://github.com/SciCatProject/scicat-backend-next/pull/2092/commits/1a2ca44805b2955b348195c4762fb998a2e21ef6)

minottic (approved):
very nice thanks! 

a very minor comment (but I guess it's a matter of preferences). If you wanted to stick to nestjs even closer, you could use [this](https://docs.nestjs.com/techniques/serialization#transform-plain-objects) rather than plainToInstance.

Maybe it's worth squashing the commits together on merge

test/DatasetFilter.js
```diff
@@ -963,4 +1013,140 @@ describe("0400: DatasetFilter: Test retrieving datasets using filtering capabili
...
+  });
+
+  it("0530: should delete dataset with special characters in metadata keys", async () => {
+    return request(appUrl)
+      .delete("/api/v3/datasets/" + encodedDatasetPidSpecial)
+      .set("Accept", "application/json")
+      .set({ Authorization: `Bearer ${accessTokenArchiveManager}` })
+      .expect(TestData.SuccessfulDeleteStatusCode)
+      .expect("Content-Type", /json/);
+  });
```
Junjiequan, line 1084: 0530 and 0620 tests are not so relevant, but I guess it's fine to leave it

Junjiequan (approved):
Looks good to me.
Please merge it after testing the migration script locally

nitrosx (approved):

Issue #689, opened 2023-08-25
Add field for capturing dataset links


dylanmcreynolds:
# Background
SciCat allows users to add thumbnails (stored in mongo) with each data set that is ingested. This allows facilities to provide some very rudimentary visualization alongside datasets. 

There is also a "Jupyter" button on datasets that can only be configured to take the user to a particular Jupyter(Hub) instance. This cannot be coded to take users to a particular notebook, nor is there a method to associate a notebook with a particular dataset once opened.

We get a lot of requests from users and researchers to provide a variety of different types of web-based tools that related directly to datasets. These can be pages that display visualization, pages that provide analysis tools (including AI/ML) and jupyter notebooks on particular JupyterHub instance. In all cases, we would want to take the user to the page with the dataset in context.

# Proposal
I can imagine a lot of solutions in SciCat. We could start maintaining visualizations on a technique/instrument basis within SciCat, but the wide variety of instruments, techniques and tools used across the Scicat facilities makes this a truly daunting task.

Instead, I propose moving the task of creating links for datasets ingestors. I proposed adding a new field to the dataset models (raw and derived) called `links`. This would be a list of link object that contain urls, display and type, something like:

```json

"links": [
  {"type": "analysis", "display": "Reconstruction Notebook", "link": "http://magrathea.org/jupyter/a/b/c/notebook.ipynb?scicat_dataset=abc123"},
  {"type": "visualization", "display": "Raw Data Visualization", "link": "http://magrathea.org/planet_viz?scicat_dataset=abc123"},
  {"type": "segmentation", "display": "Fjord Segmentation Application", "link": "http://magrathea.org/fjord_seg?scicat_dataset=abc123"},
]

```

The frontend could then display these items as href links. We might even think about adding another optional `embed` field to each link that lets the frontend decide wether to embed the web page in an `iframe` tag so that the visitation opens when you view the dataset.

If this were approved by the community, I could put it into the backend now, then follow up with a PR to the front end to display the links.

nitrosx:
@dylanmcreynolds this would be really useful, thumbs up from me. 
Let's discuss it at the next meeting.

I would also include an icon to be shown.
The equivalent json would be:
```
"links": [
  {
    "type": "analysis", 
    "user_name": "Reconstruction Notebook", 
    "url" : "http://magrathea.org/jupyter/a/b/c/notebook.ipynb?scicat_dataset=<dataset.pid>",
    "icon" : "<path_to_the_icon_file>",
  },
[...]

PR #2097, opened 2025-08-04
feat: templated external links (draft)

GBirkel:
## Description

Add a field for generating dataset links based on admin-configured templates.

See https://github.com/SciCatProject/scicat-backend-next/issues/689 for details.

This has a companion PR for the front end: https://github.com/SciCatProject/frontend/pull/1910 .

## Motivation

We get a lot of requests from users and researchers to provide a variety of different types of web-based tools that related directly to datasets. These can be pages that display visualization, pages that provide analysis tools (including AI/ML) and jupyter notebooks on particular JupyterHub instance. In all cases, we would want to take the user to the page with the dataset in context.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation


src/datasets/schemas/externallink.class.ts
```diff
@@ -0,0 +1,31 @@
+import { ApiProperty } from "@nestjs/swagger";
+import { IsString } from "class-validator";
+
+// This data is not represented in Mongoose.
+// It is generated internally based on the user-defined link templates.
```
nitrosx: Not sure what this comments means?
Can it be removed?

nitrosx (changes requested):
IT can be approved after fixing the linting and adding some API testing.
Also I would check if dataset schema to obsolete dto works correctly

GBirkel:
> Also I would check if dataset schema to obsolete dto works correctly

It does, though I had to add `toObject: { virtuals: true, }`
 https://github.com/SciCatProject/scicat-backend-next/pull/2097/files#diff-4b9a5b8e6eef69831951cba2b15a48dd5fd9b05d06b5fe10d223171039448199R18

Looks like the current-to-obsolete conversion is the only code that uses `toObject` on datasets.

src/datasets/schemas/externallink.class.ts
GBirkel: I've modified the comment somewhat.  It explains why a class without schema annotations is in the schema folder.

GBirkel: I moved away from using a virtual field and now have a separate API endpoint for fetching links.

Junjiequan:
Duplication PR with https://github.com/SciCatProject/backend/pull/2194/changes

PR #2109, opened 2025-08-05
fix: remove measurementPeriod inheritance

minottic:
## Description
This was problematic because the "[timestamps](https://mongoosejs.com/docs/guide.html#timestamps)" option is missing in the measurementsPeriod schema. The createAt and updatedAt are not defaulted when left empty. This then complains whenever using the SDK as they are checked by pydantic and missing.

I believe his could have been fixed by setting the timestamps and adding a migration script (as legacy measurements have no createdAt and updatedAt so any GET with the SDK would return error), but I don't think we require these extra properties in measurements

## Fixes

* remove extends QueryableClass

## Tests included

- [ ] Passing? 

## Documentation


Junjiequan (approved):
looks fine to me, but please check with @nitrosx before merging it

minottic:
@nitrosx any opinion?

PR #2110, opened 2025-08-07
fix: hide pending_registration to non auth

minottic:
## Description

Unauth users are able to see "pending_registration" data. I think this should not happen, and I think it was also the intention of [this interceptor](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/published-data/interceptors/registered.interceptor.ts), which does not work though. This is a problem especially in the landing page where people browsing see pending_reg (that they should not see IMO) and which do not yet have a DOI registered on doi.org.

⚠️ I have noticed request interceptors that modify requests.query do not behave as intended (at least for me, and this was also the reason of this fix). I believe this happens because request.query is a shallow copy thus does not modify the original object that is then passed to controllers. So for example the [interceptors here](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/interceptors/datasets-public.interceptor.ts) do not modify the request.query fields (but it seems they are not really used, as the same logic happens somewhere else in the controllers)

## Fixes

* do not return published_data in pending_registration
* interceptor refactored to pipe

## Changes:

* Pipe to transform from id to doi. This can be applied to all other controllers in later PRs
* Pipe to add status: registered to fields as plain obj
* Pipe to add status: registered to fields in filters: {filters: string, fields: string, limits: string}

## Tests included

- [x] Passing? 

## Documentation


minottic:
sorry, I did not see [this PR](https://github.com/SciCatProject/scicat-backend-next/pull/2035/files#diff-7969e20db7a6d27fc718560d0ba12ed71d184a6d369d5aa47e455627cc8e747e) before. Could it be that this will be fixed there already @martin-trajanovski ?

martin-trajanovski:
> sorry, I did not see [this PR](https://github.com/SciCatProject/scicat-backend-next/pull/2035/files#diff-7969e20db7a6d27fc718560d0ba12ed71d184a6d369d5aa47e455627cc8e747e) before. Could it be that this will be fixed there already @martin-trajanovski ?

@minottic yes you are right this should be fixed in my PR. We changed the status names a bit there but when the data is in `private` (which was the actual `pending_registration` before) status only the creators and admins can see it.

minottic:
> @minottic yes you are right this should be fixed in my PR. We changed the status names a bit there but when the data is in private (which was the actual pending_registration before) status only the creators and admins can see it.


mmm but will this require a migration for old data then?

martin-trajanovski:
> > @minottic yes you are right this should be fixed in my PR. We changed the status names a bit there but when the data is in private (which was the actual pending_registration before) status only the creators and admins can see it.
> 
> mmm but will this require a migration for old data then?

There is a migration here https://github.com/SciCatProject/scicat-backend-next/pull/2035/files#diff-b683df9498a94d26ac6a796216aed274736814ac8bb9917fec399e126404f52aR34. Let me know if you think that we need to do some improvements there.

minottic:
ah ok. I believe though migration scripts should be avoided whenever possible. They are often heavy for the DB and require extra care + set up to be run. What's actually the value here of renaming `pending_reg` to `private`?

minottic:
ah I see it's not the only change done by the migration script, happy with the renaming then

minottic:
@martin-trajanovski do you have an estimate for when that PR will be merged? Should I close this one? Or shall we merge it as a temporary fix to the problem, before your solution gets merged?

minottic:
@martin-trajanovski do you think we could merge this before #2035 is complete? It's a hot fix that should change much your PR

martin-trajanovski:
> @martin-trajanovski do you think we could merge this before #2035 is complete? It's a hot fix that should change much your PR

I think I can merge the #2035 today.

minottic:
superseed by #2035, thanks @martin-trajanovski !

Issue #2078, opened 2025-07-23
Replace /Datasets/resetArchiveStatus endpoint


sbliven:
## Summary
backend-v3 has a `/Dataset/resetArchiveStatus?datasetId` endpoint. 

> [PUT /api/v3/Dataset/resetArchiveStatus](https://dacat.psi.ch/explorer/#!/Dataset/Dataset_reset)
> Reset archive status by deleting created datablocks of dataset and resetting archiveStatusMessage and archivable flag.

This is a bit of an odd endpoint that combines a few operations needed for our 'reset' workflow. This is called by the archive system to recover from errors with the tape system. It deletes all datablocks associated with a dataset and also updates the datasetlifecycle. Afterwards the dataset is in an 'archivable' state, ready for a fresh archive job.

## Current Behaviour

- The endpoint is missing in scicat-backend-next from both v3 and v4 dataset controllers.
- `DELETE /api/v3/Datasets/{id}/datablocks`, which does half of this endpoint, is also missing in v3 and v4 controllers

## Current workaround

Currently this functionality could be implemented with multiple calls:
1. `GET /api/v3/datasets/{pid}/datablocks` to list all datablock IDs
2. Loop over datablocks, calling `DELETE /api/v3/datasets/{pid}/datablocks/{did}` for each
3. `PATCH /api/v3/datasets/{pid}/datasetlifecycle` to update the message and flags:
```json
{
  archivable: true,
  retrievable: false,
  publishable: false,
  archiveStatusMessage: "datasetCreated",
  retrieveStatusMessage: "",
  retrieveIntegrityCheck: false
}
```

## Details

How should this functionality be implemented in v4? The previous implementation was rigid (eg hard-coded messages) but it was convenient to make only a single request. This is ensures atomicity.

I could also imagine this as two requests:
1. `DELETE /api/v4/datasets/{pid}/datablocks` (not implemented yet)
2. `PATCH /api/v4/datasets/{pid}/datasetlifecycle`

Or, we could keep the single request `PUT /api/v4/Dataset/resetArchiveStatus` (is put the right verb?) but add parameters for the messages.

Should we implement backwards compatibility for v3? Or do we remove this endpoint from `/api/v3`?

HayenNico:
[...]

PR #2112, opened 2025-08-11
feat: implement DELETE datasets/{pid}/datablocks

minottic:
## Description
Implement DELETE datasets/{pid}/datablocks

## Motivation
In our archive/retrieve scripts we sometimes need to clean space on tape and remove files. To keep scicat in sync, we remove the dataset and the attached datablocks which can be many (sometimes >1000). This PR adds an endpoint to delete all datablocks attached to a PID, in v3. In v4, this will need to be rethought, maybe leveraging this PR #1939 and providing a deleteMany and a recovery mechanism.

Follows from #2078 

## Changes:

* DELETE datasets/{pid}/datablocks
* removeMany in datablocks service

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -2569,6 +2569,58 @@ export class DatasetsController {
     return null;
   }
 
+  // DELETE /datasets/:id/datablocks
+  @UseGuards(PoliciesGuard)
+  @CheckPolicies("datasets", (ability: AppAbility) =>
+    ability.can(Action.DatasetDatablockDelete, DatasetClass),
+  )
+  @Delete("/:pid/datablocks")
+  @ApiOperation({
+    summary: "It deletes all datablock from the dataset.",
+    description:
+      "It deletes all datablocks from the dataset.<br>This endpoint is obsolete and will be dropped in future versions.<br>Deleting datablocks will be done only from the datablocks endpoint.",
+  })
+  @ApiParam({
+    name: "pid",
+    description:
+      "Persistent identifier of the dataset for which we would like to delete the datablock specified",
+    type: String,
+  })
+  @ApiResponse({
+    status: HttpStatus.OK,
+    schema: {
+      type: "object",
+      properties: {
+        deletedCount: { type: "integer" },
+      },
+    },
+    description:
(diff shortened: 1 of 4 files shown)


src/datasets/datasets.controller.ts
```diff
@@ -2569,6 +2569,58 @@ export class DatasetsController {
     return null;
   }
 
+  // DELETE /datasets/:id/datablocks
+  @UseGuards(PoliciesGuard)
+  @CheckPolicies("datasets", (ability: AppAbility) =>
+    ability.can(Action.DatasetDatablockDelete, DatasetClass),
```
sbliven, line 2575: Is this different from `DatablockDeleteEndpoint`? I'm confused by the casl logic. There's also `DatasetDatablockDeleteAny`, how does that fit in?

sbliven (commented):
The implementation looks good, but I don't understand the casl permissions.

src/datasets/datasets.controller.ts
minottic: I have reused the same CASL as [here](https://github.com/SciCatProject/scicat-backend-next/pull/2112/files#diff-b03fa945c1274c82dae201ae4333a5bb728fb5ce9813b5e0a89121b22875ed61L2504-L2505)

Junjiequan: @sbliven 
`DatablockDeleteEndpoint` is endpoint level check for @Delete("/datablocks/:id") - if user do not have permission `@CheckPolicies` will throw Forbidden right away without proceeding.
`DatasetDatablockDeleteAny `is an instance-level check. The endpoint first fetches the dataset, then compares the user’s groups with the groups in the dataset. If there’s a match, the delete action is allowed.

And I agree it's confusing : D There will be refactor soonish

sbliven: @Junjiequan Permissions for `DELETE /datablocks/:blockid` should be the same as `DELETE /dataset/:pid/datablocks`. We should delete one of the policies during the refactor.

sbliven (approved):
It seems like the permissions are correct, although they should be simplified in the future.

Issue #2122, opened 2025-08-14
SciCat authentication refactor plan - phase 1


Junjiequan:
This outlines how the authentication refactor should be carried out and the standards it will follow, without going into in-depth technical details. A table showing the current authentication logic is needed before proceeding with the refactor proposal, so we can better understand the differences between the current logic and the planned approach. The implementation will break elasticsearch, ignore elasticsearch for now, and we can fix elasticsearch in phase 2.

## Why this refactor ##

- Hard-coded, duplicated, and inconsistent auth logic: ~2000 lines of
can/cannot rules across multiple services plus ~300 lines per controller to inject manual
filters (e.g. inject isPublished OR ownerGroup to mongo query by permission), which diverge between
services and are hard to maintain.
- Inefficient auth check for single data instance: fetch data instance-> check data instance -> postfilter,
causing irrelevant fetch and throw away if user do not have access.
- Current CASL rules hard to read – complaints from the SciCat community
- Inflexible CASL rules - cannot be configured to match facility-specific requirements

## Objectives / Non-goals ##

- Keep refactor scoped to auth
- Make auth configurable from JSON (one file), validated by JSON Schema validator.
- Reduce auth code and collapse to 2 layers: endpoint (guard) + instance (DB filter)
- Use **@casl/mongoose** to push instance conditions into Mongo queries (accessibleBy) to
avoid post-filtering.
- Remove manual filter injectors (e.g. modify filters via canViewAny, canViewOwner etc).
- Move CASL authentication business logic from controller to service file.
  - [Controller to handle I/O](https://docs.nestjs.com/controllers)
  - [Service to handle complex business logic](https://docs.nestjs.com/providers)
- No ElasticSearch changes in phase 1 (refactor ES in a follow-up)
- No new dependency beyond what you already have other than @casl/mongoose (reuse
existing JSON Schema validator)

### Two layers casl ability check per request ###

- Endpoint layer: route guard – Controller level, no conditions
- Instance layer: Specific data access – Service level, with conditions (except admin)

### Casl Rule precedence ###

[**_The rule of thumb is to define general rules first and more specific after general ones._**](https://casl.js.org/v4/en/guide/define-rules)

- **Groups** rule should win over **Defaults** rule – always generate Group permission after
defaults permission (defaults rule is for unauthen3cated).
- **cannot** should win over **can** for the same rule– always generate cannot after can.

### Casl Actions ###

For each module, only 8 or 10 actions should be applied:

[...]

PR #2123, opened 2025-08-14
refactor: proof-of-concept refactor of casl permissions

HayenNico:
## Description
This PR tries to refactor the permission system in casl by merging endpoint- and instance-level access control functions and reducing the number of action keys. The current proof-of concept covers `Attachments`, `Datablocks` and `Datasets`.

## Motivation
The current casl models include a lot of redundancies and are explicitly separated in endpoint access control and instance access control for each collection. Using the casl condition mechanism ([docs](https://casl.js.org/v6/en/guide/intro#basics)) these two cases can be expressed in a combined function in `casl-ability-factory.ts`.

Additionally, with this merge a lot of the endoint- and instance-level keys in `action.enum.ts` become redundant and can be removed. At instance-level, the separation between different access levels (e.g. `CreateAny` vs. `CreateOwner`) as different actions is also unnecessary and can be deferred to the casl conditions.

## Fixes

* Removed an illegal limit filter on /api/v4/datasets/count in test 2500-0400

## Changes:

* merge endpoint and instance access functions in `casl-ability-factory.ts` for each colection
* remove action keys in `action.enum.ts` that are implied by casl conditions

## Tests included

- [ ] Passing? 

## Documentation


HayenNico:
I'd like some comments on whether this approach aligns with what we want from the refactor of casl permissions. My approach here is fairly heavy-handed in removing redundancies, so the concern is that this may go too far.

Junjiequan:
I’ve created a draft access-refactoring plan: https://github.com/SciCatProject/scicat-backend-next/issues/2122.
The approach is similar to yours but goes further by using [@casl/mongoose](https://casl.js.org/v6/en/package/casl-mongoose) to reduce manual query filter injection and post-filtering.
There should be a table or diagram comparing the current access logic (how actions are used now) with the expected refactored logic, so whoever works on the refactor has a solid understanding of the current access logic, can avoid the mistakes we made before.

HayenNico:
@Junjiequan I've read through the plan #2122 and the docs for @casl/mongoose, and I think that plan is better than the approach in this PR. As I understand according to #2122 the aim is to eliminate the need for hardcoded casl abilities entirely by making them configurable.
I won't continue on this version then, but I'll leave the PR up for a bit to let others comment.

One thing I would add though: Endpoint and instance abilities are generally the same, since each instance rule implies a coresponding endpoint rule, so in the JSON schema I don't think there's a need to maintain both. With the casl check `ability.can(action, subject)` one can generally pass the `subjectType` (= the class of the subject instance) for the endpoint-level check.

PR #2124, opened 2025-08-14
feat: add instrumentIds property to Proposal schema

Junjiequan:
## Description
Added new property to the proposal schema
```ts
  @Prop({
    type: [String],
    default: [],
    required: false,
  })
  instrumentIds?: string[];
```

## Motivation
Enable filtering proposals by instrumentId

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/migrations/20250814143105-add-instrumentids-field-to-proposal.js
+++ b/migrations/20250814143105-add-instrumentids-field-to-proposal.js
@@ -1,0 +1,14 @@
+module.exports = {
+  async up(db, client) {
+    await db
+      .collection("Proposal")
+      .updateMany(
+        { instrumentIds: { $exists: false } },
+        { $set: { instrumentIds: [] } },
+      );
+  },
+
+  async down(db, client) {
+    // No path backward
+  },
+};
--- a/src/proposals/schemas/proposal.schema.ts
+++ b/src/proposals/schemas/proposal.schema.ts
@@ -161,6 +161,16 @@ export class ProposalClass extends OwnableClass {
     default: DEFAULT_PROPOSAL_TYPE,
   })
   type: string = DEFAULT_PROPOSAL_TYPE;
+
+  /**
+   * List of instrument IDs associated with the proposal.
+   */
+  @Prop({
+    type: [String],
+    default: [],
+    required: false,
+  })
+  instrumentIds?: string[];
 }
 
 export const ProposalSchema = SchemaFactory.createForClass(ProposalClass);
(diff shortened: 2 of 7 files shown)


migrations/20250814143105-add-instrumentids-field-to-proposal.js
```diff
@@ -0,0 +1,14 @@
...
+    await db
+      .collection("Proposal")
+      .updateMany(
+        { instrumentIds: { $exists: false } },
+        { $set: { instrumentIds: [] } },
+      );
+  },
+
+  async down(db, client) {
+    // No path backward
```
martin-trajanovski, line 12: Maybe down migration should be removing the `instrumentIds` field as we are adding it in this up migration.

src/proposals/schemas/proposal.schema.ts
```diff
@@ -161,6 +161,16 @@ export class ProposalClass extends OwnableClass {
...
+
+  /**
+   * List of instrument IDs associated with the proposal.
+   */
+  @Prop({
+    type: [String],
+    default: [],
+    required: false,
+  })
+  instrumentIds?: string[];
```
martin-trajanovski, line 173: I see the `instrumentIds` is added here but I am wondering where do we set this value. I cannot see it in the create or update DTOs.

migrations/20250814143105-add-instrumentids-field-to-proposal.js
Junjiequan: I'm not sure if we should allow removing instrumentIds with down migration...
For empty instrumentIds field its fine to remove the field, but what about instrumentIds that contains values? if we accidentally down migration in the future, there's no way to retrieve back lost values.

src/proposals/schemas/proposal.schema.ts
Junjiequan: Yes you are right it should be in the updateProposal dto

martin-trajanovski: We can leave it like this I guess. It's not a big deal even if we have it

martin-trajanovski (approved):

migrations/20250814143105-add-instrumentids-field-to-proposal.js
nitrosx: I agree with @Junjiequan
I think we should not offer down migration, although we can offer it with the explicit note that there will be data loss

nitrosx (approved):

Issue #2130, opened 2025-08-19
Including attachments when fetching datasets doesn't work


sofyalaski:
## Summary

For the Datasets collection V4 API, GET method set to include attachments never returns any attachments, but an empty array.

## Steps to Reproduce

use a GET method on Dataset V4 endpoint with filter set to `include : ["attachments"]` 

## Current Behaviour

always returns an empty array of attachments

## Expected Behaviour

including the information on the attachments

## Details

This happens because the lookup is not defined correctly for attachments. [It's expecting a field](https://github.com/SciCatProject/scicat-backend-next/blob/ba638c1ee29115466e683a9912c13e1a8d221f3d/src/datasets/types/dataset-lookup.ts#L50-L57) `datasetId` in the attachments collection, while the attachment schema doesn't include that property. [It instead has](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/attachments/schemas/attachment.schema.ts) a `relationships` property that is an object where `targetId` property lives. 
We should be looking for `relationships.targetId` and not the `datasetId` ( and `relationships.targetType` equal to "dataset" )  when we perform a lookup with the attachments collection

PR #2131, opened 2025-08-19
fix: correct way to join attachments with datasets

sofyalaski:
## Description

Fixes the way a lookup on Attachments from Datasets is done, issue #2130 
## Fixes

* Bug fixed (#2130)
the lookup structure is now reflecting the attachment's true output dto
## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation


martin-trajanovski (approved):
Looks good to me

PR #2132, opened 2025-08-19
fix: make updatedby optional

minottic:
## Description
As updatedBy is only added on update and the QueryableClass schema is shared on create and update, the property has been marked as optional

## Fixes

* make updatedBy optional

## Tests included

- [ ] Passing? 

## Documentation


minottic:
superseeded by #2134

Issue #1969, opened 2025-06-06
Support for conditional requests/Strategy for not missing updates?


cchndl:
Hi!

I was wondering how to make sure in SciCat that updates don't get lost when multiple clients access the same resource?
What I wanted to do uses the origdatablocks endpoint, but probably extends to the other PATCH endpoints.

I noticed the Etag header was set, and I tried supplying [If-Match](https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/If-Match) and [If-Unmodified-Since](https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/If-Unmodified-Since), but as far as I now understand after having a look, the default behaviour in the etag library is to only hash the response value for caching, and you have to implement the logic for these headers yourself (or I'm doing something wrong).

Is there another way that I have missed? The search didn't turn up issues/pr's.

Best
cchndl

Here is a python test script, using the instruments endpoint since it's more compact than origdatablocks/datasets:

```python
import requests
# setup
url = ""
instrument = {"uniqueName": "test-instrument", "name": "TBD"}
credentials = {"username": "", "password": ""}

r = requests.post(f"{url}/api/v3/auth/login", json=credentials)

headers = {"Authorization": "Bearer " + r.json().get("access_token")}

r = requests.post(f"{url}/api/v3/instruments/", headers=headers, json=instrument)
instr_id = r.json()["pid"]

# get the initial instrument state
r = requests.get(f"{url}/api/v3/instruments/{instr_id}", headers=headers)

not_modified_headers = headers.copy()
not_modified_headers["If-Match"] = r.headers.get("etag")
not_modified_headers["If-Unmodified-Since"] = r.headers.get("date")

intermediate_update = {"name": "intermediate"}
conditional_update = {"name": "conditional"}

# do an unconditional update
requests.patch(
    f"{url}/api/v3/instruments/{instr_id}",
    headers=headers,
    json=intermediate_update,
)
# this should only happen if the resource has not changed since the GET request, no?
r = requests.patch(
    f"{url}/api/v3/instruments/{instr_id}",
    headers=not_modified_headers,
    json=conditional_update,
)

r = requests.get(f"{url}/api/v3/instruments/{instr_id}", headers=headers)
print(r.json().get("name"))  # will print "conditional"
```

sbliven:
[...]

PR #2138, opened 2025-08-22
fix: considering header field if-unmodified-since for different patch requests

joe-baudisch:
## Description
"_if-unmodified-since_" header field is considered for patch requests for the following endpoints:

- datasets (v3/v4)
- attachments (v4)
- instruments
- origdatablocks (v3/v4)
- samples
- proposals

If the "_if-unmodified-since_" header field is missing, the patch request will still be excuted.

## Fixes

*  fixes main aspect of #1969

### DIFF
--- a/src/attachments/attachments.v4.controller.ts
+++ b/src/attachments/attachments.v4.controller.ts
@@ -15,6 +15,8 @@ import {
   Patch,
   Put,
   HttpCode,
+  Headers,
+  HttpException,
 } from "@nestjs/common";
 import {
   ApiBearerAuth,
@@ -370,21 +372,36 @@ Set \`content-type\` header to \`application/merge-patch+json\` if you would lik
   async findOneAndUpdate(
     @Req() request: Request,
     @Param("aid") aid: string,
+    @Headers() headers: Record<string, string>,
     @Body() updateAttachmentDto: PartialUpdateAttachmentV4Dto,
   ): Promise<OutputAttachmentV4Dto | null> {
-    const foundAattachment = await this.checkPermissionsForAttachment(
+    const headerDateString = headers["if-unmodified-since"];
+    const headerDate =
+      headerDateString && !isNaN(new Date(headerDateString).getTime())
+        ? new Date(headerDateString)
+        : null;
+
+    const foundAttachment = await this.checkPermissionsForAttachment(
       request,
       aid,
       Action.AttachmentUpdateEndpoint,
     );
     const updateAttachmentDtoForservice =
       request.headers["content-type"] === "application/merge-patch+json"
-        ? jmp.apply(foundAattachment, updateAttachmentDto)
+        ? jmp.apply(foundAttachment, updateAttachmentDto)
         : updateAttachmentDto;
-    return this.attachmentsService.findOneAndUpdate(
-      { _id: aid },
-      updateAttachmentDtoForservice,
-    );
+
+    if (headerDate && headerDate <= foundAttachment.updatedAt) {
+      throw new HttpException(
+        "Update error due to failed if-modified-since condition",
+        HttpStatus.PRECONDITION_FAILED,
+      );
+    } else {
+      return this.attachmentsService.findOneAndUpdate(
+        { _id: aid },
+        updateAttachmentDtoForservice,
+      );
+    }
   }
 
   // PUT /attachments/:aid
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -6,6 +6,7 @@ import {
   Delete,
   ForbiddenException,
   Get,
+  Headers,
   HttpCode,
   HttpException,
   HttpStatus,
@@ -1343,11 +1344,19 @@ export class DatasetsController {
     @Req() request: Request,
     @Param("pid") pid: string,
     @Body()
+    @Headers()
+    headers: Record<string, string>,
     updateDatasetObsoleteDto:
       | PartialUpdateRawDatasetObsoleteDto
       | PartialUpdateDerivedDatasetObsoleteDto
       | PartialUpdateDatasetDto,
   ): Promise<OutputDatasetObsoleteDto | null> {
+    const headerDateString = headers["if-unmodified-since"];
+    const headerDate =
+      headerDateString && !isNaN(new Date(headerDateString).getTime())
+        ? new Date(headerDateString)
+        : null;
+
     const foundDataset = await this.datasetsService.findOne({
       where: { pid },
     });
@@ -1356,53 +1365,59 @@ export class DatasetsController {
       throw new NotFoundException();
     }
 
-    // NOTE: Default validation pipe does not validate union types. So we need custom validation.
-    let dtoType;
-    switch (foundDataset.type) {
-      case DatasetType.Raw:
-        dtoType = PartialUpdateRawDatasetObsoleteDto;
-        break;
-      case DatasetType.Derived:
-        dtoType = PartialUpdateDerivedDatasetObsoleteDto;
-        break;
-      default:
-        dtoType = PartialUpdateDatasetDto;
-        break;
+    if (headerDate && headerDate <= foundDataset.updatedAt) {
+      throw new HttpException(
+        "Update error due to failed if-modified-since condition",
+        HttpStatus.PRECONDITION_FAILED,
+      );
+    } else {
+      // NOTE: Default validation pipe does not validate union types. So we need custom validation.
+      let dtoType;
+      switch (foundDataset.type) {
+        case DatasetType.Raw:
+          dtoType = PartialUpdateRawDatasetObsoleteDto;
+          break;
+        case DatasetType.Derived:
+          dtoType = PartialUpdateDerivedDatasetObsoleteDto;
+          break;
+        default:
+          dtoType = PartialUpdateDatasetDto;
+          break;
+      }
+      const validatedUpdateDatasetObsoleteDto =
+        (await this.validateDatasetObsolete(
+          updateDatasetObsoleteDto,
+          dtoType,
+        )) as
+          | PartialUpdateRawDatasetObsoleteDto
+          | PartialUpdateDerivedDatasetObsoleteDto
+          | PartialUpdateDatasetDto;
+
+      // NOTE: We need DatasetClass instance because casl module can not recognize the type from dataset mongo database model. If other fields are needed can be added later.
+      const datasetInstance =
+        await this.generateDatasetInstanceForPermissions(foundDataset);
+
+      // instantiate the casl matrix for the user
+      const user: JWTUser = request.user as JWTUser;
+      const ability = this.caslAbilityFactory.datasetInstanceAccess(user);
+      // check if he/she can create this dataset
+      const canUpdate =
+        ability.can(Action.DatasetUpdateAny, DatasetClass) ||
+        ability.can(Action.DatasetUpdateOwner, datasetInstance);
+
+      if (!canUpdate) {
+        throw new ForbiddenException("Unauthorized to update this dataset");
+      }
+
+      const updateDatasetDto = this.convertObsoleteToCurrentSchema(
+        validatedUpdateDatasetObsoleteDto,
+      ) as UpdateDatasetDto;
+
+      const res = this.convertCurrentToObsoleteSchema(
+        await this.datasetsService.findByIdAndUpdate(pid, updateDatasetDto),
+      );
+      return res;
     }
-
-    const validatedUpdateDatasetObsoleteDto =
-      (await this.validateDatasetObsolete(
-        updateDatasetObsoleteDto,
-        dtoType,
-      )) as
-        | PartialUpdateRawDatasetObsoleteDto
-        | PartialUpdateDerivedDatasetObsoleteDto
-        | PartialUpdateDatasetDto;
-
-    // NOTE: We need DatasetClass instance because casl module can not recognize the type from dataset mongo database model. If other fields are needed can be added later.
-    const datasetInstance =
-      await this.generateDatasetInstanceForPermissions(foundDataset);
-
-    // instantiate the casl matrix for the user
-    const user: JWTUser = request.user as JWTUser;
-    const ability = this.caslAbilityFactory.datasetInstanceAccess(user);
-    // check if he/she can create this dataset
-    const canUpdate =
-      ability.can(Action.DatasetUpdateAny, DatasetClass) ||
-      ability.can(Action.DatasetUpdateOwner, datasetInstance);
-
-    if (!canUpdate) {
-      throw new ForbiddenException("Unauthorized to update this dataset");
-    }
-
-    const updateDatasetDto = this.convertObsoleteToCurrentSchema(
-      validatedUpdateDatasetObsoleteDto,
-    ) as UpdateDatasetDto;
-
-    const res = await this.convertCurrentToObsoleteSchema(
-      await this.datasetsService.findByIdAndUpdate(pid, updateDatasetDto),
-    );
-    return res;
   }
 
   // PUT /datasets/:id
--- a/src/datasets/datasets.v4.controller.ts
+++ b/src/datasets/datasets.v4.controller.ts
@@ -1,24 +1,26 @@
 import {
-  BadRequestException,
   Body,
-  ConflictException,
   Controller,
   Delete,
-  ForbiddenException,
   Get,
-  HttpCode,
-  HttpStatus,
-  InternalServerErrorException,
-  NotFoundException,
   Param,
   Patch,
   Post,
   Put,
   Query,
-  Req,
   UseGuards,
   UseInterceptors,
+  HttpCode,
+  HttpStatus,
+  NotFoundException,
+  Req,
+  BadRequestException,
+  ForbiddenException,
+  InternalServerErrorException,
+  ConflictException,
   UsePipes,
+  Headers,
+  HttpException,
 } from "@nestjs/common";
 import {
   ApiBearerAuth,
@@ -784,9 +786,31 @@ Set \`content-type\` header to \`application/merge-patch+json\` if you would lik
   async findByIdAndUpdate(
     @Req() request: Request,
     @Param("pid") pid: string,
+    @Headers() headers: Record<string, string>,
     @Body()
     updateDatasetDto: PartialUpdateDatasetDto,
   ): Promise<OutputDatasetDto | null> {
+    return this.findByIdAndUpdateInternal(
+      request,
+      pid,
+      headers,
+      updateDatasetDto,
+    );
+  }
+
+  async findByIdAndUpdateInternal(
+    @Req() request: Request,
+    @Param("pid") pid: string,
+    @Headers() headers: Record<string, string>,
+    @Body()
+    updateDatasetDto: PartialUpdateDatasetDto,
+  ): Promise<OutputDatasetDto | null> {
+    const headerDateString = headers["if-unmodified-since"];
+    const headerDate =
+      headerDateString && !isNaN(new Date(headerDateString).getTime())
+        ? new Date(headerDateString)
+        : null;
+
     const foundDataset = await this.datasetsService.findOne({
       where: { pid },
     });
@@ -813,15 +837,22 @@ Set \`content-type\` header to \`application/merge-patch+json\` if you would lik
       );
     }
 
-    const updateDatasetDtoForService =
-      request.headers["content-type"] === "application/merge-patch+json"
-        ? jmp.apply(foundDataset, updateDatasetDto)
-        : updateDatasetDto;
-    const updatedDataset = await this.datasetsService.findByIdAndUpdate(
-      pid,
-      updateDatasetDtoForService,
-    );
-    return updatedDataset;
+    if (headerDate && headerDate <= foundDataset.updatedAt) {
+      throw new HttpException(
+        "Update error due to failed if-modified-since condition",
+        HttpStatus.PRECONDITION_FAILED,
+      );
+    } else {
+      const updateDatasetDtoForService =
+        request.headers["content-type"] === "application/merge-patch+json"
+          ? jmp.apply(foundDataset, updateDatasetDto)
+          : updateDatasetDto;
+      const updatedDataset = await this.datasetsService.findByIdAndUpdate(
+        pid,
+        updateDatasetDtoForService,
+      );
+      return updatedDataset;
+    }
   }
 
   // GET /datasets/:id/datasetlifecycle
--- a/src/instruments/instruments.controller.ts
+++ b/src/instruments/instruments.controller.ts
@@ -1,16 +1,20 @@
 import {
-  Controller,
-  Get,
-  Post,
   Body,
-  Patch,
-  Param,
-  Delete,
-  UseGuards,
-  Query,
-  UseInterceptors,
-  InternalServerErrorException,
   ConflictException,
+  Controller,
+  Delete,
+  Get,
+  Headers,
+  HttpException,
+  HttpStatus,
+  InternalServerErrorException,
+  NotFoundException,
+  Param,
+  Patch,
+  Post,
+  Query,
+  UseGuards,
+  UseInterceptors,
 } from "@nestjs/common";
 import { MongoError } from "mongodb";
 import { InstrumentsService } from "./instruments.service";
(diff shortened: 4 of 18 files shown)


Junjiequan:
Isn't `if-unmodified-since` an optional field in the header? that it has to be set from the client side to make it actually useful. 
I see the issues you are trying to resolve here, but it would be nice to do concurrency control on the server side instead of relying on client

joe-baudisch:
> Isn't `if-unmodified-since` an optional field in the header? that it has to be set from the client side to make it actually useful. I see the issues you are trying to resolve here, but it would be nice to do concurrency control on the server side instead of relying on client

In my opinion this approach is entirely server-side: the server tracks timestamps and controls concurrency. 
This is a form of optimistic concurrency control, where the client assumes the resource hasn't changed and wants to avoid overwriting newer data. Clients still need to send If-Unmodified-Since, but they don’t manage timestamp-based versioning logic. If clients do not want to use the server-side management of this versioning logic, they omit if-unmodified-since.
Solving this issue is a special request from MLZ.

cchndl:
Looks good! I think its a good compromise to use modified-since only for now, as using Etags would require computing/storing them and be quite a lot more effort to implement as it is right now. 

@Junjiequan to your concerns:

> I see the issues you are trying to resolve here, but it would be nice to do concurrency control on the server side instead of relying on client

Well the problem is that the time between these requests may be quite long.

Doing locking in the backend is not a good Idea:
In my opinion you don't want to give out locks over resources on a REST endpoint, since its stateless and you are not guaranteed to have it released in a timely manner. And enforcing some timeouts would be bad for expensive operations on the client side and for performance. In the first case, the client would never be able to finish its computation and make the request before the lock is released again and in the latter one, waiting for these timeouts prevents other clients to do anything else.

For doing some locking directly on the database, I think that has the same issues. I don't think you want to hold a database lock while waiting for a client to make a follow-up post/patch request.

Optimistic concurrency with the [conditional requests](https://www.rfc-editor.org/rfc/rfc7232#section-1) is the standard way to do this, at least to my knowledge.

> Isn't if-unmodified-since an optional field in the header? that it has to be set from the client side to make it actually useful.

Yes, the client has to set it. For our case at MLZ at least, the clients we write internally for the automatic ingestion will use it. If its merged, I want to have a look at scitacean to implement support there and then most use-cases should be covered. As we probably won't require clients to send it with every request, there will of course still be the option around it, but well behaved clients can use it, which is a better situation than now.

src/attachments/attachments.v4.controller.ts
```diff
@@ -376,10 +383,16 @@ Set \`content-type\` header to \`application/merge-patch+json\` if you would lik
...
-    );
+
+
+    if (!headerDate || headerDate > foundAattachment.updatedAt) {
+      return this.attachmentsService.findOneAndUpdate(
+        {_id: aid},
+        updateAttachmentDtoForservice,
+      );
+    } else {
+      throw new HttpException("Precondition Failed", HttpStatus.PRECONDITION_FAILED);
```
nitrosx: Can we make the error message more detailed?
Something like `Update error due to failed if-modified-since condition`

src/datasets/datasets.controller.ts
```diff
@@ -1410,52 +1417,57 @@ export class DatasetsController {
...
-        dtoType,
-      )) as
-        | PartialUpdateRawDatasetObsoleteDto
-        | PartialUpdateDerivedDatasetObsoleteDto
-        | PartialUpdateDatasetDto;
-
-    // NOTE: We need DatasetClass instance because casl module can not recognize the type from dataset mongo database model. If other fields are needed can be added later.
-    const datasetInstance =
-      await this.generateDatasetInstanceForPermissions(foundDataset);
+    if (!headerDate || headerDate > foundDataset.updatedAt) {
```
nitrosx: I would negate the condition and return the exception in the `then` branch.
 If everything is fine, continue with the execution of the function.

src/datasets/datasets.v4.controller.ts
```diff
@@ -809,15 +816,20 @@ Set \`content-type\` header to \`application/merge-patch+json\` if you would lik
...
-    const updateDatasetDtoForService =
-      request.headers["content-type"] === "application/merge-patch+json"
-        ? jmp.apply(foundDataset, updateDatasetDto)
-        : updateDatasetDto;
-    const updatedDataset = await this.datasetsService.findByIdAndUpdate(
-      pid,
-      updateDatasetDtoForService,
-    );
-    return updatedDataset;
+    if (!headerDate || headerDate > foundDataset.updatedAt) {
```
nitrosx: same comment as in dataset v3.
negate the if condition and return the exception.

src/instruments/instruments.controller.ts
```diff
@@ -157,26 +163,41 @@ export class InstrumentsController {
...
+      if (!headerDate) {
+        return this.instrumentsService.update({_id: id}, updateInstrumentDto);
+      }
+
+      // If header is present, compare with updatedAt
+      if (!instrument.updatedAt || headerDate > instrument.updatedAt) {
+        return this.instrumentsService.update({_id: id}, updateInstrumentDto);
+      } else {
+        throw new HttpException("Precondition Failed", HttpStatus.PRECONDITION_FAILED);
+      }
```
nitrosx: Can we implement it as follow:
```
// If header is missing, always update
      if ( headerDate && instrument.updatedAt && headerDate <= instrument.updatedAt ) {
        throw new HttpException("Precondition Failed", HttpStatus.PRECONDITION_FAILED);
      }

      return this.instrumentsService.update({_id: id}, updateInstrumentDto);
```

src/instruments/schemas/instrument.schema.ts
```diff
@@ -74,6 +74,10 @@ export class Instrument {
     default: {},
   })
   customMetadata: Record<string, unknown>;
+
+  createdAt?: Date;
+  updatedAt?: Date;
```
nitrosx: This fields are added by extending `QueryableClass`.
Please do not add them locally.

src/origdatablocks/origdatablocks.controller.ts
```diff
@@ -651,14 +652,30 @@ export class OrigDatablocksController {
...
+        updateOrigDatablockDto,
+      );
+      if (!origdatablock) {
+        throw new NotFoundException("Datablock not found");
+      }
+      await this.updateDatasetSizeAndFiles(origdatablock.datasetId);
+      return origdatablock;
+    } else {
+      throw new HttpException("Precondition Failed", HttpStatus.PRECONDITION_FAILED);
+    }
```
nitrosx: Same comment as above. Can you negate the condition?

src/origdatablocks/origdatablocks.v4.controller.ts
```diff
@@ -773,16 +778,32 @@ export class OrigDatablocksV4Controller {
...
+        updateOrigDatablockDto,
+      );
+
+      if (!origdatablock) {
+        throw new NotFoundException("Datablock not found");
+      }
       await this.updateDatasetSizeAndFiles(origdatablock.datasetId);
+      return origdatablock;
+    } else {
+      throw new HttpException("Precondition Failed", HttpStatus.PRECONDITION_FAILED);
```
nitrosx: Same comments as above. Please negate the condition and change the if statement

src/proposals/proposals.controller.ts
```diff
@@ -720,17 +720,38 @@ export class ProposalsController {
...
+        throw new NotFoundException("Proposal not found");
+      }
+      if (!headerDate || headerDate > proposal.updatedAt) {
+        return this.proposalsService.update(
+          {proposalId: proposalId},
+          updateProposalDto,
+        );
+      } else {
+        throw new HttpException("Precondition Failed", HttpStatus.PRECONDITION_FAILED);
+      }
```
nitrosx: Same comment on the condition and if statement

src/samples/samples.controller.ts
```diff
@@ -697,9 +697,36 @@ export class SamplesController {
...
+      }
+
+      // If header is present, compare with updatedAt
+      if (!sample.updatedAt || headerDate > sample.updatedAt) {
+        console.log("Header date is newer — proceeding with update");
+        return this.samplesService.update({sampleId: id}, updateSampleDto);
+      } else {
+        console.log("Header date is older — skipping update");
+        throw new HttpException("Precondition Failed", HttpStatus.PRECONDITION_FAILED);
+      }
```
nitrosx: Same comment as above on the if statement

nitrosx (changes requested):
Please switch around the if statements as suggested.
Also make sure to include some API test to test backward compatibility and the functionality itself.

nitrosx:
As long as that the feature is backward compatible and it does not effect the updates if it is not used, I'm open to accept it.

@joe-baudisch does this applies only to updates, correct?

joe-baudisch:
> As long as that the feature is backward compatible and it does not effect the updates if it is not used, I'm open to accept it.
> 
> @joe-baudisch does this applies only to updates, correct?

@nitrosx : this only applies to this updates.

Junjiequan:
@cchndl 
Apologies for my misinterpretation of optimistic concurrency control.
What I wanted to say is actually using ETag with versions to control concurrency requests instead of relying on the date to control patches. For a temporary solution it’s fine, but it is not a very clean solution compared to version ETag. First, there might be timing issues (rare, but still). Second, it’s a fix only for clients that are aware of concurrency issues. I could imagine concurrency requests causing data loss if not handled carefully. That being said, I agree that ETag takes more effort to implement, which is also best practice if I’m not mistaken.

nitrosx:
Jumping in as a complete ignorant here.
how does an ETag be more or strict than a if-modified-before?
It all depends how the information is used by the client.

cchndl:
> Apologies for my misinterpretation of optimistic concurrency control.

I mean you didn't misinterpreted it, I just wanted to clarify why this would be preferable to other options. I hope I was not overbearing!

> What I wanted to say is actually using ETag with versions to control concurrency requests instead of relying on the date to control patches. For a temporary solution it’s fine, but it is not a very clean solution compared to version ETag. First, there might be timing issues (rare, but still). Second, it’s a fix only for clients that are aware of concurrency issues. I could imagine concurrency requests causing data loss if not handled carefully. That being said, I agree that ETag takes more effort to implement, which is also best practice if I’m not mistaken.

Full agreement with you there. Its good to have both supported in the end.

cchndl:
> Jumping in as a complete ignorant here. how does an ETag be more or strict than a if-modified-before? It all depends how the information is used by the client.

@nitrosx 
Maybe "more strict" is a little bit too strong, the good thing about Etags is that they are really a version "distinguisher", if that makes sense. If its not the same etag, something changed. 

For the last-modified, if you read and afterwards in the same second an update comes, you wont see that. Its not likely but possible. The specification only says second-accuracy. For example, a  last-modified-header that the etag libraty gives could be:
last-modified Wed, 30 Apr 2025 08:37:54 GMT
This should be fine for most resources but if you want to be sure for writes, the Etag is in that sense "better" to check for.

In the end, i believe having both would be ideal. The last-modified is "free" in the sense that most of the objects already track it, so it can't hurt. Depending on how we do it, it may also be cheaper to first check the timestamp and then the etag if both are given, but that I can't say now, thats something one would have to check when we build it.

As the etag library generates the Etag by hashing the response body (not the headers I think?), and not the object in the database, we would have to do some work there. This is a nice default behaviour for caching page reads and so on, but not for the objects themselves.  We would have to hook Etag generation into there somewhere or save the Etag in the db in or next to the object.

Hope this helps.

nitrosx:
@cchndl thank you for the explanation.
I think there are already ETag functionality in nest.js.
I found the following post that mention ETag in nest.js:
- https://stackoverflow.com/questions/68080302/disabling-auto-generated-etags-in-nestjs-app
Although I'm not sure it is relevant to our use case

cfelder:
If you do not allow sub-second updates on a resource there is no concurrency issue using ``If-Unmodified-Since``.

Example: Assuming we have two clients A and B requesting the same resource R concurrently
* Client A and B get the same last modified time t0
* Client A sends an update with If-Unmodified-Since: t0
* Client B sends an update with If-Unmodified-Since: t0
* One Client will win, let's assume Client B wins b/c of a shorter round trip time
* The Application receives update B first and updates its last modified time to t1 (server based timestamp)
* The Application receives request from A and compares the last modified time t0 with t1 and responds with HTTP 412 Precondition Failed.

For our use case this simple implementation is good enough and clients can handle http 412 accordingly. I currently do not see a need for sub-second concurrent updates on our end.

nitrosx:
I agree with @cfelder

joe-baudisch:
> Please switch around the if statements as suggested. Also make sure to include some API test to test backward compatibility and the functionality itself.
 
Added a test that can be applied in a similar way to the other affected controllers.

joe-baudisch:
Hello @cchndl, @cfelder, dear maintainers,

Do you need further testing?
Otherwise, I suggest merging the pull request into the master so I don't run into issues like the one with  [fdfd2de](https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/fdfd2deab4976a346f6f3aaf476665780d4254e5)

The master has evolved, but it hasn't taken into account that the API call this.datasetsController.findByIdAndUpdate in
src/published-data/published-data.v4.controller.ts requires an additional parameter, the headers.
It could  be solved with a non decorated call.

By the way:
If you're calling a decorated method just for its logic, it's usually better to extract that logic into a separate function.


Now  with  https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/de6c79a6e19de04651dd2445c75a5b864f3e00ae all tests are present and all tests were passed by running `npx jest`,
especially taking into account [10a639e](https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/10a639ed9b4e099efedb2f777a4cc1381cade44f) .
Therefore, I again suggest merging the pull request into the master.

joe-baudisch:
This one https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/03a1e8abce7be60685926bdfeecf542cd663512b
was hard

joe-baudisch:
Should I copy my tests to https://github.com/SciCatProject/scicat-backend-next/tree/master/test/jest-e2e-tests ?

cchndl:
From my testing, it looks good. Are there any other points, or do you think we can merge this @nitrosx?

joe-baudisch:
It is not my fault that some  tests in https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/8384ba28d30e39c856072bd6f26534c5398408f9  failed.

### Failing **API tests with ElasticSearch enabled yes** in https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/8384ba28d30e39c856072bd6f26534c5398408f9 because of:


1. **InstrumentFilter Tests** (e.g. test/InstrumentsFilter.js:265, 307, 350, 375, 384, 393, 402):
   - Error: expected "Content-Type" header field
   - Solution: Ensure that your API responses for instrument filtering endpoints always set the "Content-Type" header, especially for JSON responses. In your controllers/middleware, add:
   ```js
   res.set('Content-Type', 'application/json');
   ```
   before sending a response.

2. **OrigDatablockForRawDataset & OrigDatablockV4 Tests** (e.g. test/OrigDatablockForRawDataset.js:499, 558, test/OrigDatablockV4.js:866, 886):
   - Error: expected 200 "OK", got 403 "Forbidden"
   - Solution: These tests are being denied permission. Check your authorization logic for the endpoints handling origdatablock updates. Ensure the test users have the required roles/permissions in your test setup or mock data.

3. **Proposal and RawDataset Tests** (e.g. test/Proposal.js:359, test/RawDataset.js:310, 326, 342, 358, 371, 384, 397, test/RawDatasetDatablock.js:71):
   - Errors: 404 "Not Found", 500 "Internal Server Error"
   - Solution: For 404s, make sure the test data is seeded correctly and the IDs used exist. For 500s, check your update logic for null references or missing required fields, and add error handling/logging to clarify root causes.

4. **Sample Authorization Tests** (e.g. test/SampleAuthorization.js:3093, 3117, 3141, 3165, 3206, 3247, 3288, 3329, 3353):
   - Error: expected 200 "OK", got 404 "Not Found"
   - Solution: The sample being updated does not exist. Make sure your test database is seeded with all required samples before running these tests.


### Failing **API tests with ElasticSearch enabled no** in https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/8384ba28d30e39c856072bd6f26534c5398408f9 because of:



1. "expected 'Content-Type' header field" errors
- **Files & Lines:**
  - test/InstrumentsFilter.js:265, 307, 350, 375, 384, 393, 402 ([examples](https://github.com/SciCatProject/scicat-backend-next/blob/8384ba28d30e39c856072bd6f26534c5398408f9/test/InstrumentsFilter.js#L265))
- **Root Cause:** The API responses are missing the `Content-Type` header, likely for JSON payloads.
- **Solution:** In your API controller (where these endpoints are defined), ensure you set this header, for example:
  ```js
  res.set('Content-Type', 'application/json');
  ```
  Add this line before sending responses related to instrument filtering.



 2. "expected 200 'OK', got 403 'Forbidden'" errors
- **Files & Lines:**
  - test/OrigDatablockForRawDataset.js:499, 558
  - test/OrigDatablockV4.js:866, 886
- **Root Cause:** The tests are trying to update resources but lack the required permissions, or the authentication/authorization logic is incorrect.
- **Solution:** 
  - Ensure test users have the right roles/permissions.
  - Check your authorization middleware to confirm that users (admin or ownerGroup) are correctly allowed to perform updates.
  - Review the related endpoint logic to ensure it respects the test setup.



 3. "expected 200 'OK', got 404 'Not Found'" errors
- **Files & Lines:**
  - test/Proposal.js:359
  - test/SampleAuthorization.js:3093, 3117, 3141, 3165, 3206, 3247, 3288, 3329, 3353
- **Root Cause:** The test is attempting to update or reference entities that do not exist, possibly due to missing setup steps or cleanup from previous tests.
- **Solution:** 
  - Make sure the test data (proposals, samples) are created before being updated.
  - In your test files, verify that setup hooks (`beforeEach`, etc.) are actually creating the required entities and not failing silently.
  - Example: 
    ```js
    beforeEach(async () => {
      await createSample({ id: 1, ... });
    });
    ```
    If the entity creation fails, the update will return 404.


 4. "expected 200 'OK', got 500 'Internal Server Error'" and "expected 400 'Bad Request', got 500 'Internal Server Error'"
- **Files & Lines:**
  - test/RawDataset.js:310, 326, 342, 358, 371, 384, 397
  - test/RawDatasetDatablock.js:71
- **Root Cause:** The server is crashing during update attempts, likely due to missing properties, null references, or bad data.
- **Solution:** 
  - Add error handling and input validation in the update endpoints for RawDataset and Datablock.
  - In the failing tests, log the request data and the error messages to pinpoint what is going wrong.
  - Example: Check for `null` or `undefined` values before accessing properties or saving to the database.

joe-baudisch:
It's not my fault that some tests failed in https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/8384ba28d30e39c856072bd6f26534c5398408f9  and  https://github.com/SciCatProject/scicat-backend-next/pull/2138/commits/19908229fdc1462f0e6f1aa61f6cff111d2d5ff1

src/datasets/datasets.v4.controller.ts
```diff
@@ -778,9 +780,31 @@ Set \`content-type\` header to \`application/merge-patch+json\` if you would lik
...
+    );
+  }
+
+  async findByIdAndUpdateInternal(
+    @Req() request: Request,
+    @Param("pid") pid: string,
+    @Headers() headers: Record<string, string>,
+    @Body()
+    updateDatasetDto: PartialUpdateDatasetDto,
+  ): Promise<OutputDatasetDto | null> {
```
nitrosx, line 807: Why do we define function `findByIdAndUpdateInternal` and call it only from `findByIdAndUpdate`?
Can we combine `findByIdAndUpdate` and `findByIdAndUpdateInternal`?

src/instruments/schemas/instrument.schema.ts
```diff
@@ -13,7 +14,7 @@ export type InstrumentDocument = Instrument & Document;
     getters: true,
   },
 })
-export class Instrument {
+export class Instrument extends QueryableClass {
```
nitrosx, line 17: Have we updated the instrument output dto to provide the additional fields introduced by `QueryableClass`?

src/published-data/published-data.v4.controller.ts
nitrosx, line 1: I'm not sure that I would introduce this change here!!!
If you are creating a published data record the datasets needs to be public, no exception.

That said, I might not be knowledgeable about the background and context that you are operating in.
Can you explain better why you need the changes in this file?

nitrosx (changes requested):
I left few comments which might become change requests.

Also, we should think about submitting a following-up PR introducing all the API tests covering all the use cases.
I will be happy to help you defining the use cases.

bpedersen2 (changes requested):
I have some change requests here:

 1)  We should cleanup this long string of commits by rebasing and possibly squashing some commits.
2)  I think the if-modifed check should be implemented as a Pipe or Interceptor, so we do not get so much code doubling. It also would make it easier to later implement a hash check

src/attachments/attachments.v4.controller.ts
```diff
@@ -370,21 +372,36 @@ Set \`content-type\` header to \`application/merge-patch+json\` if you would lik
...
+        ? jmp.apply(foundAttachment, updateAttachmentDto)
         : updateAttachmentDto;
-    return this.attachmentsService.findOneAndUpdate(
-      { _id: aid },
-      updateAttachmentDtoForservice,
-    );
+
+    if (headerDate && headerDate <= foundAttachment.updatedAt) {
+      throw new HttpException(
+        "Update error due to failed if-modified-since condition",
```
minottic, line 396: maybe we should raise something more talkative?

minottic (commented):
I very much agree with @bpedersen2 comment, if this was a pipe or an interceptor a lot of duplication could be avoided

src/attachments/attachments.v4.controller.ts
bpedersen2: There are alreay the ConflictException and the PreconditionFailed exceptions available from nest. 
How about `PreconditionFailed('Resource has been changed on server')`

See also https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/412 for an ETAG example

bpedersen2:
As for using ETags, the hash should be done over the resource, so we could add that to all GET requests and also return it on POST request via an interceptor. 

Another question, why is  done only for patch ( which should only send fields that you want to change anyway), but not for PUT, which would overwrite the complete entry?

cfelder:
We have already discussed that ETag implementation can be handled separately. I Strongly suggest to concentrate on the if-unmodified-since header implementation here to not further delay reaching a state where we can merge this.

joe-baudisch:
I agree with https://github.com/SciCatProject/scicat-backend-next/pull/2138#issuecomment-3484939122

cchndl:
> Another question, why is done only for patch ( which should only send fields that you want to change anyway), but not for PUT, which would overwrite the complete entry?

We wanted to keep the scope small while still solving our problems. That's why it includes only the patch endpoints.

> ( which should only send fields that you want to change anyway), 

Of course if these fields are distinct sets it works already with PATCH. But as discussed in the mentioned issue, (linking it [here](1969) again), the problem is with nested structures like the list of data files. There the patch request will override the changes made by other clients, since there is no merging of lists. And if the changes touch the same fields, the same problem applies anyway.

src/datasets/datasets.controller.ts
```diff
@@ -1420,13 +1420,14 @@ export class DatasetsController {
...
     const foundDataset = await this.datasetsService.findOne({
-      where: { pid },
+      where: {pid},
     });
 
     if (!foundDataset) {
       throw new NotFoundException();
     }
 
+<<<<<<< HEAD
```
cfelder: unresolved merge conflict

src/datasets/datasets.v4.controller.ts
```diff
@@ -813,6 +805,7 @@ Set \`content-type\` header to \`application/merge-patch+json\` if you would lik
       );
     }
 
+<<<<<<< HEAD
```
cfelder: unresolved merge conflict

src/origdatablocks/origdatablocks.v4.controller.ts
```diff
@@ -783,8 +783,28 @@ export class OrigDatablocksV4Controller {
       updateOrigDatablockDto,
     );
 
+<<<<<<< HEAD
```
cfelder: unresolved merge conflict

joe-baudisch: I have to check why this happened

nitrosx:
Although I thought that we had already addressed all the comments on this PR, I proposed to discuss the issue once more at the next SciCat meeting.
My two cents: I know that this PR might address just a specific use case, I think we should accept it once the change requests are done and the tests are passing.
This will help MLZ to go in production, with the understanding that this implementation is a just a step in the right direction and needs to be refined and made more elegant (aka adopting nestjs solution and reducing code duplication), which I hold the authors to maintain this promise.

nitrosx:
@joe-baudisch could you please check the feasibility to implement the functionality as interceptor?

FloKo2669:
@nitrosx I checked the possibility of implementing this as an interceptor. The problem is that, if I’m not wrong, a double fetch of the resource would be unavoidable (once in the interceptor and once in the actual controller). In theory, the interceptor could cache the resource for later use, but I think this could lead to confusion and is not very transparent. Additionally, in some cases (for example, Attachments), the fetch happens in the checkPermissions function, so caching the resource wouldn’t help.

My current solution, to reduce at least some code duplication and avoid double fetching, is to add a helper function that handles the date validation and comparison. In the end, this change would result in adding only a call to the helper function in each controller function.

nitrosx:
@FloKo2669 thank you for your quick reply and for checking on this. 
If you could implement this helper function and merge it in this PR, I would greatly appreciate it and it would speed up the acceptance of the PR and Production for MLZ.
We can than take a nice deep breath, sit down, discuss at length and find a more elegant solution.

minottic:
If I understand right this PR, the additional added logic is this: for every controller on PATCH (with the relevant dataservice). 

```TS
    const headerDateString = headers["if-unmodified-since"];
    const headerDate =
      headerDateString && !isNaN(new Date(headerDateString).getTime())
        ? new Date(headerDateString)
        : null;

    const datablock = await this.origDatablocksService.findOne({
      where: { _id: id },
    });
    if (!datablock) {
      throw new NotFoundException("Datablock not found");
    }

    if (headerDate && headerDate <= datablock.updatedAt) {
      throw new HttpException(
        "Update error due to failed if-modified-since condition",
        HttpStatus.PRECONDITION_FAILED,
      );
```
IMO, this is the perfect candidate for an interceptor. 

> @nitrosx I checked the possibility of implementing this as an interceptor. The problem is that, if I’m not wrong, a double fetch of the resource would be unavoidable (once in the interceptor and once in the actual controller). In theory, the interceptor could cache the resource for later use, but I think this could lead to confusion and is not very transparent. Additionally, in some cases (for example, Attachments), the fetch happens in the checkPermissions function, so caching the resource wouldn’t help.

If the concern is doing `await this.origDatablocksService.findOne` both in the interceptor and in the controller, I would honestly favour code readability, i.e. using the interceptor. Doing the findOne for non-huge DBs is usually very efficient O(logN) (and also mongoDB itself has a cache-like mechanism to speed up equivalent queries) , and we do it already in other controllers, e.g. [here](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/pipes/scientific-metadata-validation.pipe.ts#L51-L55).

joe-baudisch:
To come back to https://github.com/SciCatProject/scicat-backend-next/pull/2138#issuecomment-3484939122:

We've once again run into merge conflicts on and now I have to fix the tests because the merge was delayed. It's really frustrating to have someone else merge changes and then be left chasing down bugs afterward. This process needs to be more coordinated to avoid unnecessary rework.

joe-baudisch:
https://github.com/SciCatProject/scicat-backend-next/pull/2138#issuecomment-3484939122
I slipped on the keyboard.

joe-baudisch:
I will close my pull request, because  [#2331](https://github.com/SciCatProject/scicat-backend-next/pull/2331)

PR #2153, opened 2025-08-27
fix: improving logbook error

abdimo101:
## Description
Improving the error message when there is no logbooks
image

## Fixes
* https://jira.ess.eu/browse/SWAP-4960

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation


src/logbooks/logbooks.service.ts
```diff
@@ -58,16 +58,23 @@ export class LogbooksService {
...
+            .pipe(
+              catchError((error) => {
+                return of({ data: null });
+              }),
+            ),
         );
 
         if (!res.data) {
-          Logger.log("Logbook not found", { name });
+          Logger.log("Logbook not found");
```
Junjiequan: It would be nice to keep the name as it is for easier debug

Junjiequan (approved):

PR #2157, opened 2025-08-27
fix: add migration for jobs

minottic:
## Description
The logic is that after up all jobs in the Jobs collection are shaped as a V4 job. The mapping comes from [this function](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/jobs.controller.utils.ts#L610-L632).

This is the Job collection before running the migration. The 1st Job is a v3 Job (thus needs migration), the 2nd is a V4 one (thus no migration)
image
**after migration up** (note that 2nd hasn't changed), is 1st correct @sbliven? (a bunch of fields are now missing, I guess I should add those as well - these are the missing fields even after migration:   
```
  'ownerGroup',
  'accessGroups',
  'isPublished',
  'ownerUser',
  'statusMessage',
  'configVersion',
  'jobResultObject',
  '__v'
```
):
image

For the missing fields, I could suggest:

```
  'ownerGroup',: jobParams.username
  'accessGroups': [jobParams.username] or []
  'isPublished': true if "type: public"
  'ownerUser',: jobParams.username
  'statusMessage': ''
  'configVersion': do not add
  'jobResultObject': do not add
  '__v': do not add
```

**after migration down:**
image

An input from JOBS developers would help greatly

## Fixes

* add jobs up migration
* add jobs down migration

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/migrations/20250827150710-jobs-v3-to-v4-ids.js
+++ b/migrations/20250827150710-jobs-v3-to-v4-ids.js
@@ -17,6 +17,7 @@ module.exports = {
             { isPublished: { $exists: false } },
             { configVersion: { $exists: false } },
             { jobResultObject: { $exists: false } },
+            { statusMessage: { $exists: false } },
           ]
         },
         [
@@ -41,6 +42,7 @@ module.exports = {
               isPublished: { $ifNull: ["$isPublished", false] },
               configVersion: { $ifNull: ["$configVersion", ""] },
               jobResultObject: { $ifNull: ["$jobResultObject", {}] },
+              statusMessage: { $ifNull: ["$statusMessage", ""] },
             }
           },
           {
--- a/migrations/20250827150710-jobs-v3-to-v4-ids.js
+++ b/migrations/20250827150710-jobs-v3-to-v4-ids.js
@@ -11,6 +11,12 @@ module.exports = {
             { $and: [{ statusCode: { $exists: false } }, { jobStatusMessage: { $exists: true } }] },
             { $and: [{ "jobParams.datasetList": { $exists: false } }, { datasetList: { $exists: true } }] },
             { $and: [{ "jobParams.executionTime": { $exists: false } }, { executionTime: { $exists: true } }] },
+            { $and: [{ ownerUser: { $exists: false } }, { "jobParams.username": { $exists: true } }] },
+            { ownerGroup: { $exists: false } },
+            { accessGroups: { $exists: false } },
+            { isPublished: { $exists: false } },
+            { configVersion: { $exists: false } },
+            { jobResultObject: { $exists: false } },
           ]
         },
         [
@@ -28,7 +34,13 @@ module.exports = {
                     executionTime: { $ifNull: ["$executionTime", "$jobParams.executionTime"] }
                   }
                 ]
-              }
+              },
+              ownerUser: { $ifNull: ["$jobParams.username", "$ownerUser"] },
+              ownerGroup: { $ifNull: ["admin", "ownerGroup"] },
+              accessGroups: { $ifNull: [[], "$accessGroups"] },
+              isPublished: { $ifNull: [false, "$isPublished"] },
+              configVersion: { $ifNull: ["", "$configVersion"] },
+              jobResultObject: { $ifNull: [{}, "$jobResultObject"] },
             }
           },
           {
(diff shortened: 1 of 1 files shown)


despadam:
Hi @minottic,

It seems that in your suggestion you are assigning usernames to groups, which is not guaranteed to work. I am referring to these two lines:
```
'ownerGroup': jobParams.username
'accessGroups': [jobParams.username] or []
```
where we do not know if the username is the same as the name of the group, and apart from that, since a user can belong in more than one groups, we do not know whether the group you are trying to assign would be the correct one.

In this [mapping](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/interceptors/create-job-v3-mapping.interceptor.ts) you can see how we tried to handle that for v3 jobs that are created via the frontend. More specifically, in [these lines](https://github.com/SciCatProject/scicat-backend-next/blob/202fe76cfded135f4f63dc9048911f7ab1b47f44/src/jobs/interceptors/create-job-v3-mapping.interceptor.ts#L98-L131) we use the groups of the dataset(s) to assign groups to jobs. For `#datasetAccess` for example, we find all common groups between the `accessGroups` and `ownerGroup` of all datasets in the job's `jobParams.datasetList` and intersect it with the user's group. At least one common group should be found, otherwise it means that the user does not have access to the dataset(s) and cannot create the job anyway, unless the dataset is public.

This requires more logic on your side, but hopefully guarantees a more correct mapping.

minottic:
thanks! I'll have a look. I am a bit afraid of the impact on the DB if heavy logic, but I'll give it a try. If I understand right, worst case we can leave it empty, right, since what you mention will add the fields on the fly?

Would the rest be correct otherwise, in particular this one?
```
'isPublished': true if "type: public"
```

minottic:
> In this [mapping](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/interceptors/create-job-v3-mapping.interceptor.ts) you can see how we tried to handle that for v3 jobs that are created via the frontend. More specifically, in [these lines](https://github.com/SciCatProject/scicat-backend-next/blob/202fe76cfded135f4f63dc9048911f7ab1b47f44/src/jobs/interceptors/create-job-v3-mapping.interceptor.ts#L98-L131) we use the groups of the dataset(s) to assign groups to jobs. For #datasetAccess for example, we find all common groups between the accessGroups and ownerGroup of all datasets in the job's jobParams.datasetList and intersect it with the user's group. At least one common group should be found, otherwise it means that the user does not have access to the dataset(s) and cannot create the job anyway, unless the dataset is public.

if I get this right, I could use the intersection of datasets' groups to create the accessGroups of the job. But how do I create the ownerGroup? Do I just pick one randomly? I don't have the "currentUser" info during migration

minottic:
after discussing offline with @despadam , a good compromise seems to be:

```
  ownerUser: jobParams.username
  ownerGroup: "admin",
  accessGroups: [],
  isPublished: false,
  configVersion: '',
  jobResultObject: {},
```

this would allow when myself.username === ownerUser to see the JOB

ownerGroup = 'admin', hardcoded, since it's a mandatory property in the schema and should have a value, and will not be used for old data, since anyway users don't update jobs

minottic:
now migration is from (1st is V3, 2nd is V4)
image

migration up:
image
migration down:
image

despadam:
Hi @minottic, in the first job you have assigned `ownerGroup: 'ownerGroup'` but should be `ownerGroup: 'admin'` instead.

minottic:
> Hi @minottic, in the first job you have assigned ownerGroup: 'ownerGroup' but should be ownerGroup: 'admin' instead.

thanks, changed (the comment had old screenshots)

despadam (approved):
In general looks good to me, the only field missing from V4 is `statusMessage`, which is not a mandatory field in the [schema](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/schemas/job.schema.ts), however would be good to add it for completeness reasons and to avoid errors just in case someone is using it.
It could be either an empty string or take the `jobStatusMessage` value as well.

minottic:
> In general looks good to me, the only field missing from V4 is statusMessage, which is not a mandatory field in the [schema](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/jobs/schemas/job.schema.ts), however would be good to add it for completeness reasons and to avoid errors just in case someone is using it.
> It could be either an empty string or take the jobStatusMessage value as well.

Thanks. I had omitted that as it was optional. I will add it then

despadam:
Thanks! From my side looks ok, unless you want to wait for reviews from others as well.

PR #2159, opened 2025-08-27
fix(APIv4): enable dataset subdocument keys for where filters

HayenNico:
## Description
The `ALLOWED_DATASET_KEYS` list is expanded to include the keys of dataset subdocuments like lifecycle and relationship. This makes filters of the form `{ "where" : { "datasetlifecycle.archivable" : true } }` valid, to match API v3 behavior.

* Added keys for lifecycle, relationship, technique and history to valid keys in dataset where filters
* Added keys for datafile to valid keys in origdatablock where filters
* Removed some constraints in `FilterValidationPipe`, will permit arbitrary keys after `dataset.scientificMetadata`, `dataset.jobParameters`, `job.jobParams` and `job.jobResultObject`

## Tests included

- [x] Passing? 

## Documentation


sofyalaski:
Wouldn't the issue remain for scientific metadata? 
Also, since scientificMetadata is not necessarily following a schema, for it to be searchable, we need filters to be quite permissive.. This breaks the left-hand side filters on the main datasets view in the frontend, I expect.

Another change that would be required is to do the same expansion of allowed keys to other collections using the `FilterValidationPipe` pipe: 
`AttachmentRelationTargetType` for attachments,  `DataFile` for origdatablocks and  `jobResultObject` and `jobParams` for jobs

HayenNico:
ScientificMetadata is a remaining problem, yes.
The easiest solution would be to use the metadataKeys function in some way to determine the legal keywords, but that function is to my knowledge still very inefficient and will be rewritten at some point.

What I could do for now is to soften the requirements in `FilterValidationPipe`. Right now every sub-key in a dot-separated expression must match one of the `ALLOWED_DATASET_KEYS`. For scientificMetadata filters to work, any string needs to be permitted as valid - so maybe I could add a condition that detects "scientificMetadata" as a sub-key and allows anything thereafter? `jobResultObject` and `jobParams` would require similar handling

sofyalaski:
I agree that this is the fastest fix
> condition that detects "scientificMetadata" as a sub-key and allows anything thereafter

martin-trajanovski (approved):

PR #2163, opened 2025-08-28
fix: returns 404 for non-existent proposal delete

abdimo101:
## Description
Ensuring a 404 status code is returned for non existing proposals instead of 403. Cause of the bug is that the `PoliciesGuard` was checking for delete permission before the existence of the proposal could be verified.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation


Junjiequan (changes requested):
The endpoint permission should remain `ProposalsDelete` instead of replacing it to `ProposalsRead` 
The fix should happen in the instance permission scope instead of endpoint permission

PR #2164, opened 2025-08-29
fix: add check for unauthenticated users in origdatablock v3 controller

HayenNico:
## Description
Small fix to check for unauthenticated users in all readMany functions in `OrigDatablocksController`.

## Motivation
Unauthenticated users have `OrigdatablockRead` permissions, but will cause a 500 error in all readMany API functions since user.currentGroups is read when building permission-based filters.

## Fixes

* Unauthenticated users will no longer cause 500 errors, instead only get access to public ODBs

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/origdatablocks/origdatablocks.controller.ts
+++ b/src/origdatablocks/origdatablocks.controller.ts
@@ -321,7 +321,9 @@ export class OrigDatablocksController {
         OrigDatablock,
       );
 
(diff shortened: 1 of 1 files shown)


martin-trajanovski (approved):
Looks good to me

Junjiequan (approved):

Issue #2172, opened 2025-09-03
Uninformative error message when passing invalid json to datasets?filter=


sbliven:
## Summary

The `GET datasets` endpoint takes a JSON-format `filter` query parameter. The error message for typos in this json is very uninformative.

## Steps to Reproduce

Here is an example call with the invalid filter `{ "limits": {"limit": 1},}` (note the extra trailing comma):
```sh
curl -X 'GET' \
  'https://scicat.development.psi.ch/api/v3/datasets?filter=%7B%20%22limits%22%3A%20%7B%22limit%22%3A%201%7D%2C%7D' \
  -H 'accept: application/json' \
  -H "Authorization: Bearer $SCICAT_TOKEN"
```

It also occurs with `/api/v4/datasets?filter=`. I wouldn't be surprised if this was a general issue with all calls that take JSON query parameters.

## Current Behaviour

Currently this produces a 500 error with the generic error message "Internal server error".

## Expected Behaviour

It should give a more informative error message, eg "Unable to parse filter JSON"

## Details

This is somewhat mitigated on `/api/v4/datasets` when submitting through swagger because the swagger API validates the schema, so it won't let you submit invalid json. Unfortunately `/api/v3/datasets` doesn't validate the schema (should this be a separate bug?) so it's easy to make typos in long filter queries.

PR #1926, opened 2025-05-20
feat: parse config files as yaml

sbliven:
## Description

Use yaml to parse all JSON files (fix #1120)

This change is backwards compatible. No changes to existing json files are needed (although yaml features like comments can be added if desired). YAML features like comments can be added in the existing JSON files without changing the filename. If you want editor support for yaml it might be desirable to change the config file suffixes to `.yaml`; this is not done by default but is supported through environmental variables (`FUNCTIONAL_ACCOUNT_FILE` and similar).

## Motivation
Use of yaml for the jobConfig.yaml file has been convenient. This PR extends that out to other JSON config files.

## Fixes

* #1120

## Tests included

- [ ] Included for each change/fix?
- [ ] Passing? 

## Documentation
- [ ] swagger documentation updated (required for API changes)
- [ ] official documentation updated

PR #2175, opened 2025-09-04
fix(pipe): throw an error when malformed json is passed

sofyalaski:
## Description

Introduces a try-catch block for parsing passed filter as a json string, and when that fails it throws an error . Before it was failing with 500 "internal server error"

## Motivation

#2172 
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation


HayenNico (commented):
This looks good, but will only cover API v4 right now. #2172 indicates a similar fix is needed on the v3 controller. I think the same logic could be inserted in [DatasetsController.getFilters](https://github.com/SciCatProject/scicat-backend-next/blob/f2c27347a537a69143584d4e4770b3ebb3b21db3/src/datasets/datasets.controller.ts#L154)

nitrosx (approved):

nitrosx:
@HayenNico you are correct

HayenNico (approved):

bpedersen2:
And keep in mind possible interactions with #1926

bpedersen2 (approved):

PR #2176, opened 2025-09-04
fix: avoid changing the object types masking

minottic:
## Description
This fixes the bug reported [here](https://github.com/paulscherrerinstitute/scicat-ci/issues/336). It was due to instanceToPlain applying the schema definition of lifecycle class which misses the _id def, which converted it to a buffer. This rewrite, being more generic, avoids changing the object with "toPlain" and replaces the emails inplace, checking the recursion with a weakSet. Without the weakSet, mongoose goes into recursion error as mongoose references itself infenitely 

## Fixes

* avoid running toPlain that has side effects
* simplify recursion

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/common/interceptors/mask-sensitive-data.interceptor.ts
+++ b/src/common/interceptors/mask-sensitive-data.interceptor.ts
@@ -6,44 +6,43 @@ import {
   NestInterceptor,
 } from "@nestjs/common";
 import { APP_INTERCEPTOR } from "@nestjs/core";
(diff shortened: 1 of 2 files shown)


bpedersen2 (commented):
Looks reasonable

minottic:
> Looks reasonable

thanks! Happy to approve?

bpedersen2 (approved):

PR #2194, opened 2025-09-09
feat: templated external links

GBirkel:
This is a cleaned-up version of an earlier draft PR ( https://github.com/SciCatProject/scicat-backend-next/pull/2097 ).

## Description

Add a field for generating dataset links based on admin-configured templates.

See https://github.com/SciCatProject/scicat-backend-next/issues/689 for details.

This has a companion PR for the front-end, which is still in draft form: https://github.com/SciCatProject/frontend/pull/1910

## Motivation

We get a lot of requests from users and researchers to provide a variety of different types of web-based tools that related directly to datasets. These can be pages that display visualization, pages that provide analysis tools (including AI/ML) and jupyter notebooks on particular JupyterHub instance. In all cases, we would want to take the user to the page with the dataset in context.

## Changes:

* New API endpoint: Datasets/[pid]/externallinks

This returns an array of objects, `{ url: string, title: string, description: string }`.  See `ExternalLinkClass` in the code for more detail.

* New configuration entry, parsed from file `datasetExternalLinkTemplates.json` with an example template file.

Example of the content of this file:

```json
[
  {
    "title": "Franzviewer II",
    "url_template": "https://franz.site.com/franzviewer?id=${dataset.pid}",
    "description_template": "View ${dataset.numberOfFiles} files in Franz' own personal viewer",
    "filter": "(dataset.type == 'derived') && dataset.owner.includes('Franz')"
  },
  {
    "title": "High Beam-Energy View",
    "url_template": "https://beamviewer.beamline.net/highenergy?id=${dataset.pid}",
    "description_template": "The high-energy beamviewer (value ${dataset.scientificMetadata?.beamEnergy?.value}) at beamCo",
    "filter": "(dataset.scientificMetadata?.beamEnergy?.value > 20)"
  }
]
```

The `filter` field is JavaScript, parsed into a function at run time, which is called with a `dataset` argument containing a full dataset record.  If the function returns true, the given link applies to the given dataset.

The `url_template` and `description_template` fields are JavaScript-style format strings, which are passed the same dataset argument.  The strings they emit go into the `url` and `description` objects returned by the API.


src/datasets/datasets.controller.ts
```diff
@@ -1243,35 +1239,30 @@ export class DatasetsController {
...
-              );
-              break;
-            }
-            case "origdatablocks": {
-              outputDataset.origdatablocks =
-                await this.origDatablocksService.findAll({
-                  where: { datasetId: outputDataset.pid },
-                });
-              break;
+      if (includeFilters) {
```
emigun: Would you care to elaborate on this change a bit?

src/datasets/datasets.controller.ts
```diff
@@ -1823,6 +1814,46 @@ export class DatasetsController {
     return await this.convertCurrentToObsoleteSchema(outputDatasetDto);
   }
 
+  // GET /datasets/:id/externallinks
```
emigun, line 1754: Should we add new endpoints to the old v3 controller?

src/datasets/datasets.service.ts
```diff
@@ -395,6 +396,53 @@ export class DatasetsService {
...
+      where: { pid: id },
+      include: [DatasetLookupKeysEnum.all],
+    });
+
+    if (!thisDataSet) {
+      // no luck. we need to create a new dataset
+      throw new NotFoundException(`Dataset #${id} not found`);
+    }
+
+    interface ExternalLinkTemplateConfig {
```
emigun, line 501: I think it would be nice to have interfaces/types in separate file, but that it just my subjective view.

src/datasets/datasets.service.ts
```diff
@@ -395,6 +396,53 @@ export class DatasetsService {
...
+    }
+
+    const templates: ExternalLinkTemplateConfig[] | undefined =
+      this.configService.get("datasetExternalLinkTemplates");
+    if (!templates) {
+      return [];
+    }
+
+    return templates
+      .filter((d) => {
```
emigun: Perhaps a more descriptive variable name here?

src/datasets/schemas/externallink.class.ts
```diff
@@ -0,0 +1,32 @@
+import { ApiProperty } from "@nestjs/swagger";
+import { IsString } from "class-validator";
+
+// This class defines the externalLinks field in a dataset.
+// That field is not represented in the Mongoose data store,
+// so there is no equivalent schema representation for it.
+
+export class ExternalLinkClass {
+  @ApiProperty({
```
emigun, line 9: You could consider having these api properties automatically generated as we for example did in the samples module (if I remember correctly)

let me know if you want more information

emigun:
I think tests are missing?

minottic:
@GBirkel I have converted this to draft, as it did not have commits for the past month. Feel free to move it back to "ready for review" once you feel so or if I shouldn't have moved to draft

src/datasets/datasets.controller.ts
GBirkel: I'm not sure what the right move is, honestly.
I'd be happy to add them but it would of course be a change to the v3 API (though only an additive one).  Are we still changing that API or is it in a static state now that we have v4?

src/datasets/datasets.service.ts
GBirkel: In general I agree.  Right now this type is only used once, and in this spot.  But I think we should be using types like this when we're parsing/validating all the JSON that this config data comes from.  That may be a task best done en-masse though, as a separate change...

src/datasets/datasets.controller.ts
GBirkel: I came across this code when I was implementing an early version of this feature that included the external links in the dataset record itself, and found it confusing.  The change I made is logically equivalent but easier to parse.

src/datasets/datasets.service.ts
GBirkel: Good idea.  Addressed in https://github.com/SciCatProject/scicat-backend-next/pull/2194/commits/47838feffe64204b0a7c087878510fd6caead9d9 .

src/datasets/schemas/externallink.class.ts
GBirkel: Do you mean by using `SchemaFactory`?  I don't think we can do that here because these properties have no representations in MongoDb ... But I could be wrong.  I'd love more information!

GBirkel:
> @GBirkel I have converted this to draft, as it did not have commits for the past month. Feel free to move it back to "ready for review" once you feel so or if I shouldn't have moved to draft

Thanks. Yeah, I got sidetracked by other activity at the ALS.  I think you're right that it needs tests before I can call it ready.

PR #2195, opened 2025-09-09
fix(test): skip DataCite intergartion tests until properly mocked

sofyalaski:
## Description
When PR's a coming from the forks ( or run locally) secrets to datacite aren't available to all of collaborators and some of the PublishedDataV4 tests fail.
This PR skips the tests that depend on this integration (secrets aren't available in the .env).

This is a temporary fix and should be mocked properly with jest.

## Tests included

- [x] Passing? 

## Documentation


minottic (approved):
thanks!

PR #2196, opened 2025-09-09
feat: default datablocks owner and access to DS

minottic:
## Description
This replicates the logic from the old [BE](https://github.com/SciCatProject/backend-v3/blob/master/common/models/datablock.js#L34) that allowed for avoiding specifying the owner and accessGroups.
I believe PSI would prefer to have something similar to [origs](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/origdatablocks/origdatablocks.controller.ts#L199-L202), where they are always overridden, but this keeps backwards compatibility

## Changes:

* default to DS owner, access

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/datablocks/datablocks.controller.ts
+++ b/src/datablocks/datablocks.controller.ts
@@ -95,9 +95,14 @@ export class DatablocksController {
     );
 
     try {
-      const datablock = await this.datablocksService.create(createDatablockDto);
       const dataset = await this.datasetsService.findOne({
-        pid: datablock.datasetId,
+        pid: createDatablockDto.datasetId,
+      });
+      const datablock = await this.datablocksService.create({
+        ...createDatablockDto,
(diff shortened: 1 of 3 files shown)


HayenNico (approved):
Looks good to me.

Regarding origDatablocks: 
In my opinion we should also have this logic for the origDatablocks, or remove Ownable as the base from the DTOs for both datablock types. The current origDatablocks will allow specification of owners in the DTO and then silently drop that data in favor of the dataset owners, which may be confusing on the users' side.

minottic:
> Regarding origDatablocks:
> In my opinion we should also have this logic for the origDatablocks, or remove Ownable as the base from the DTOs for both datablock types. The current origDatablocks will allow specification of owners in the DTO and then silently drop that data in favor of the dataset owners, which may be confusing on the users' side.
> 

that's also what I thought when seeing the code. I'll open an issue and approach in another PR

PR #2197, opened 2025-09-09
fix: mock datacite calls and avoid in CI

minottic:
## Description
Enables mocking datacite calls with jest and removes tests from acceptance

## Fixes

* remove CI vars
* add tests to jest
* remove from .js

## Tests included

- [ ] Passing? 

## Documentation


minottic:
@nitrosx , @Junjiequan , if this succeeds, could you please remove the GITHUB secret?

test/PublishedDataV4.js
```diff
@@ -154,71 +148,6 @@ describe("1600: PublishedDataV4: Test of access to published data v4 endpoints",
...
-        .get("/api/v4/PublishedData/" + doi)
-        .set("Accept", "application/json")
-        .set({ Authorization: `Bearer ${accessTokenAdminIngestor}` })
-        .expect(TestData.SuccessfulGetStatusCode)
-        .expect("Content-Type", /json/)
-        .then((res) => {
-          res.body.should.have.property("status").and.equal("registered");
-        });
-    },
-  );
```
HayenNico, line 182: We could leave a variation of this test in that expects status "public" instead of "registered" after 0040

HayenNico (approved):
Looks good

test/PublishedDataV4.js
minottic: makes sense, I would say maybe in an another PR, this is only refactoring of existing tests, I wouldn't add additional logic

test/jest-e2e-tests/publishedData.e2e-spec.ts
```diff
@@ -0,0 +1,139 @@
...
+  it("Should fetch this amended published data", async () => {
+    await request(app.getHttpServer())
+      .get("/api/v4/PublishedData/" + doi)
+      .set("Accept", "application/json")
+      .set({ Authorization: `Bearer ${token}` })
+      .expect(200)
+      .expect("Content-Type", /json/)
+      .then((res) => {
+        expect(res.body.status).toEqual("amended");
+      });
```
Junjiequan, line 137: I'm trying to understand how it returns body or 201 without mocked response

minottic: mmm, good point thanks! let me check

minottic: these tests actually run the nestjs app, what is mocked is only the httpService to avoid the datacite call, so there should be no need to mock in this test. But maybe you meant something differetn?

Junjiequan: I might be wrong, but to my understanding, when httpServer/service is mocked the response also needs to be mocked

minottic: but httpServer is mocked only in 1 test, not for all of them, so the others will not be affected

Junjiequan: You are right, the endpoint just handles DB interaction, i thought it was reaching out external service

Junjiequan (approved):

Issue #2174, opened 2025-09-04
Make CreateJobDtoV3.jobStatusMessage optional


sbliven:
## Summary

In v3, `Job.jobStatusMessage` was optional and defaulted to "jobSubmitted" if not set. `CreateJobDtoV3.jobStatusMessage` should also be optional and default to `JOB_DEFAULT_STATUS_CODE`.

## Steps to Reproduce

In v3 calls to `POST /Jobs` omitted `jobStatusMessage`. For example, here's an archive job from the old frontend:

```json
{
  "jobParams": {
    "username": "***",
    "destinationPath": "/archive/retrieve",
    "option": "PSI"
  },
  "emailJobInitiator": "***@psi.ch",
  "datasetList": [
    {
      "pid": "20.500.11935/1e1edfaa-b425-4205-8ede-a7aef3963511",
      "files": []
    }
  ],
  "type": "retrieve"
}
```

Interestingly, this *is* accepted by the current backend (`POST /api/v3/jobs`), despite not conforming to the type definition. So this issue is primarily about having the correct typescript type definition; it shouldn't change any client behavior.

## Current Behaviour

Typescript clients creating a `CreateJobDtoV3` object are required to include `jobStatusMessage`.

## Expected Behaviour

`CreateJobDtoV3.jobStatusMessage` should be optional. If unset, the backend will fill it with the default.

PR #2204, opened 2025-09-12
fix: make CreateJobDtoV3.jobStatusMessage optional

despadam:
## Description

Made `CreateJobDtoV3.jobStatusMessage` optional and then also noticed that its value was always getting overridden by the default one. Made the appropriate changes to ensure the value from the dto persists, when specified, otherwise falls back to default.

## Motivation

See https://github.com/SciCatProject/scicat-backend-next/issues/2174

## Fixes

* `create-job-v3-mapping.interceptor`: stores `jobStatusMessage` temporarily in `jobParams` to ensure its value does not get overridden
* `update-job-v3-mapping.interceptor`: ensures that the previously stored values do not get overridden by the default ones
* `jobs.controller.utils`: prioritizes the `jobStatusMessage` that we stored in `jobParams`, if it exists, and then removes it from `jobParams`
* `jobs.service`: prioritizes the values coming from the dto instead of the default ones

## Changes:

* Marked `CreateJobDtoV3.jobStatusMessage` as optional

## Tests included

- [x] Passing? 

## Documentation

https://github.com/SciCatProject/documentation/pull/87


sbliven (approved):
This was a bigger change than I expected. Looks good.

PR #2212, opened 2025-09-16
fix(dataset): ensure that fields required for the lookups are included

abdimo101:
## Description
Fixes dataset V4 findOne endpoint to preserve fields required for relationship lookups.

## Motivation
When users request specific fields (e.g. only `datasetName`) while including relationships like `attachments` in their query, the system was filtering out required fields (like `pid`) needed to establish these relationships.

This resulted in empty relationship arrays being returned even when relationships existed in the database.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/attachments/attachments.service.ts
+++ b/src/attachments/attachments.service.ts
@@ -65,7 +65,7 @@ export class AttachmentsService {
 
     const pipeline: PipelineStage[] = [{ $match: whereFilter }];
     if (!isEmpty(fieldsProjection)) {
-      const projection = parsePipelineProjection(fieldsProjection);
+      const projection = parsePipelineProjection(fieldsProjection, filter);
       pipeline.push({ $project: projection });
     }
 
--- a/src/attachments/attachments.service.ts
+++ b/src/attachments/attachments.service.ts
@@ -65,7 +65,7 @@ export class AttachmentsService {
 
     const pipeline: PipelineStage[] = [{ $match: whereFilter }];
     if (!isEmpty(fieldsProjection)) {
-      const projection = parsePipelineProjection(fieldsProjection, filter.include);
+      const projection = parsePipelineProjection(fieldsProjection);
       pipeline.push({ $project: projection });
     }
 
--- a/src/attachments/attachments.service.ts
+++ b/src/attachments/attachments.service.ts
@@ -65,7 +65,7 @@ export class AttachmentsService {
 
     const pipeline: PipelineStage[] = [{ $match: whereFilter }];
     if (!isEmpty(fieldsProjection)) {
-      const projection = parsePipelineProjection(fieldsProjection, filter);
+      const projection = parsePipelineProjection(fieldsProjection, filter.include);
       pipeline.push({ $project: projection });
     }
 
--- a/src/datasets/datasets.service.ts
+++ b/src/datasets/datasets.service.ts
@@ -133,7 +133,7 @@ export class DatasetsService {
     this.addLookupFields(pipeline, filter.include);
 
     if (!isEmpty(fieldsProjection)) {
-      const projection = parsePipelineProjection(fieldsProjection);
+      const projection = parsePipelineProjection(fieldsProjection, filter);
       pipeline.push({ $project: projection });
     }
 
@@ -258,32 +258,10 @@ export class DatasetsService {
     };
 
     const pipeline: PipelineStage[] = [{ $match: whereFilter }];
-
-    // Map of lookup fields to their required join fields in the Dataset schema
-    // These are the fields that must be preserved for each relationship lookup to work correctly
-    const lookupFields: Record<string, string> = {
-      [DatasetLookupKeysEnum.attachments]: "pid",
-      [DatasetLookupKeysEnum.samples]: "sampleIds",
-      [DatasetLookupKeysEnum.proposals]: "proposalIds",
-      [DatasetLookupKeysEnum.origdatablocks]: "pid",
-      [DatasetLookupKeysEnum.datablocks]: "pid",
-      [DatasetLookupKeysEnum.instruments]: "instrumentIds",
-    };
+    this.addLookupFields(pipeline, filter.include);
 
     if (!isEmpty(fieldsProjection)) {
-      const projection = parsePipelineProjection(fieldsProjection);
-
-      // When specific fields are requested, we need to ensure that the fields required for lookups
-      // are preserved (e.g. pid for attachments). Otherwise relationship arrays would be empty
-      // since MongoDB removes all fields not explicitly requested in projection.
-
-      if (filter.include) {
-        filter.include.forEach((field: DatasetLookupKeysEnum) => {
-          const requiredField = lookupFields[field];
-          projection[requiredField] = true;
-        });
-      }
-
+      const projection = parsePipelineProjection(fieldsProjection, filter);
       pipeline.push({ $project: projection });
     }
 
@@ -294,7 +272,6 @@ export class DatasetsService {
 
     pipeline.push({ $skip: limits.skip || 0 });
 
-    this.addLookupFields(pipeline, filter.include);
 
     const [data] = await this.datasetModel
       .aggregate<OutputDatasetDto | undefined>(pipeline)
--- a/src/datasets/datasets.service.ts
+++ b/src/datasets/datasets.service.ts
@@ -259,8 +259,8 @@ export class DatasetsService {
 
     const pipeline: PipelineStage[] = [{ $match: whereFilter }];
 
-
-    // Lookup fields and their corresponding local fields in Dataset schema
+    // Map of lookup fields to their required join fields in the Dataset schema
+    // These are the fields that must be preserved for each relationship lookup to work correctly
     const lookupFields: Record<string, string> = {
       [DatasetLookupKeysEnum.attachments]: "pid",
       [DatasetLookupKeysEnum.samples]: "sampleIds",
@@ -268,19 +268,20 @@ export class DatasetsService {
       [DatasetLookupKeysEnum.origdatablocks]: "pid",
       [DatasetLookupKeysEnum.datablocks]: "pid",
       [DatasetLookupKeysEnum.instruments]: "instrumentIds",
-    }
+    };
 
     if (!isEmpty(fieldsProjection)) {
       const projection = parsePipelineProjection(fieldsProjection);
-      
-      // when specific fields are requested, we need to ensure that the fields required for the lookups are included.
-      // For example if attachments are included, we need to ensure that "pid" field is included in the projection to avoid missing data.
+
+      // When specific fields are requested, we need to ensure that the fields required for lookups
+      // are preserved (e.g. pid for attachments). Otherwise relationship arrays would be empty
+      // since MongoDB removes all fields not explicitly requested in projection.
 
       if (filter.include) {
         filter.include.forEach((field: DatasetLookupKeysEnum) => {
           const requiredField = lookupFields[field];
           projection[requiredField] = true;
-        })
+        });
       }
 
       pipeline.push({ $project: projection });
--- a/src/datasets/datasets.service.ts
+++ b/src/datasets/datasets.service.ts
@@ -191,7 +191,7 @@ export class DatasetsService {
     const pipeline: PipelineStage[] = [{ $match: whereFilter }];
     const addedRelations = this.addLookupFields(pipeline, filter.include);
 
-    if (Array.isArray(fieldsProjection)) {
+    if (Array.isArray(fieldsProjection) && fieldsProjection.length > 0) {
       fieldsProjection = Array.from(
         new Set([...fieldsProjection, ...addedRelations]),
       );
--- a/src/datasets/datasets.service.ts
+++ b/src/datasets/datasets.service.ts
@@ -258,8 +258,31 @@ export class DatasetsService {
     };
 
     const pipeline: PipelineStage[] = [{ $match: whereFilter }];
+
+
+    // Lookup fields and their corresponding local fields in Dataset schema
+    const lookupFields: Record<string, string> = {
+      [DatasetLookupKeysEnum.attachments]: "pid",
+      [DatasetLookupKeysEnum.samples]: "sampleIds",
+      [DatasetLookupKeysEnum.proposals]: "proposalIds",
+      [DatasetLookupKeysEnum.origdatablocks]: "pid",
+      [DatasetLookupKeysEnum.datablocks]: "pid",
+      [DatasetLookupKeysEnum.instruments]: "instrumentIds",
+    }
+
     if (!isEmpty(fieldsProjection)) {
       const projection = parsePipelineProjection(fieldsProjection);
+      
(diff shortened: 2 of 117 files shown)


emigun:
Looks good to me! (but I feel like someone with more knowledge on MongoDB lookups should be approving)

src/datasets/datasets.service.ts
```diff
@@ -258,8 +258,32 @@ export class DatasetsService {
...
+      // When specific fields are requested, we need to ensure that the fields required for lookups
+      // are preserved (e.g. pid for attachments). Otherwise relationship arrays would be empty
+      // since MongoDB removes all fields not explicitly requested in projection.
+
+      if (filter.include) {
+        filter.include.forEach((field: DatasetLookupKeysEnum) => {
+          const requiredField = lookupFields[field];
+          projection[requiredField] = true;
+        });
+      }
```
Junjiequan: I believe this change needs to be applied to all controllers, not just datasets. So it'd be nice to have a more generic solution to reduce duplication

Junjiequan (approved):
I think it'd be better if the solution were more generic so that other controllers can use it, but I think it works, and maybe we can refactor it later

HayenNico:
To me it looks like this issue comes from a messed up order of operations in the mongo pipeline in findOneComplete in datasets.service. What needs to change is that the fieldsProjection comes after the lookup in the pipeline.
If you compare the current findOneComplete to findAllComplete, the latter uses a different order that should always work

I don't think we need to (or should) introduce this exception for required fields

Junjiequan:
> To me it looks like this issue comes from a messed up order of operations in the mongo pipeline in findOneComplete in datasets.service. What needs to change is that the fieldsProjection comes after the lookup in the pipeline.
If you compare the current findOneComplete to findAllComplete, the latter uses a different order that should always work

@HayenNico Yes, if we change the order of fieldsProjection this will solve the issue, as long as whoever calls this endpoint knows that in order to get the lookup data they must provide the lookup document name in the $project.

For example, after changing the order like this:
```JS
const pipeline: PipelineStage[] = [{ $match: whereFilter }];

this.addLookupFields(pipeline, filter.include);

if (!isEmpty(fieldsProjection)) {
  const projection = parsePipelineProjection(fieldsProjection);
  pipeline.push({ $project: projection });
}

if (!isEmpty(limits.sort)) {
  const sort = parsePipelineSort(limits.sort);
  pipeline.push({ $sort: sort });
}
```

The following query returns the expected results:
```Json
{
  "where": {
    "datasetName": {
      "$regex": "test22222233333334",
      "$options": "i"
    }
  },
  "include": [
    "attachments"
  ],
  "fields": [
    "datasetName",
    "attachments"
  ]
}
```
But this one do not return attachments:
```Json
{
  "where": {
    "datasetName": {
      "$regex": "test22222233333334",
      "$options": "i"
    }
  },
  "include": [
    "attachments"
  ],
  "fields": [
    "datasetName"
  ]
}
```
It’s because [lookup fields/embedded documents require the embedded field to be in the $project](https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/#embedded-document-fields), otherwise it only returns the root doc.

It’s kinda against the first instinct of how it supposed to work. For example, the general use case of $project is that if you don’t provide a projection, **Mongo** returns the entire doc. So, the idea behind this change is that we keep $project after $lookup because we know what lookup fields list is provided, what corresponding IDs are required, and that the required ids field will not be changed. This way people who don’t know Mongo query logic won’t get confused.

Junjiequan:
That being said, I agree with your point. As a maintainer, changing the order is the simplest and cleanest solution.

HayenNico:
@Junjiequan Thanks for the explanation. Assuming we put the lookup pipeline step in the correct place, what we could add is that any filter under `"include"` is automatically put into the `"fields"` filter as well. I think this implicit assumption that a client who specified `"include"` would also want to see that in the projection is valid.

HayenNico:
Side note: This will require some adjustments to the tests as well. Right now [2500-0207](https://github.com/SciCatProject/scicat-backend-next/blob/4a85eac60b31ec64e4cda3712c211aabfb0dac49/test/DatasetV4.js#L565) on findAll and [2500-0308](https://github.com/SciCatProject/scicat-backend-next/blob/4a85eac60b31ec64e4cda3712c211aabfb0dac49/test/DatasetV4.js#L792) on findOne for using all query filters have exactly opposite logic with regard to the include fields, which is why the problem with empty includes wasn't picked up in testing.

Junjiequan:
> @Junjiequan Thanks for the explanation. Assuming we put the lookup pipeline step in the correct place, what we could add is that any filter under `"include"` is automatically put into the `"fields"` filter as well. I think this implicit assumption that a client who specified `"include"` would also want to see that in the projection is valid.

That sounds like a better solution. Just one additional thing needs to be clarified, if user provide the query below, there will be an error
```json
  "include": [
    "attachments"
  ],
  "fields": [
    "datasetName",
    "attachments.thumbnail"
  ]
```
Because `You cannot specify both an embedded document and a field within that embedded document in the same projection.` - [doc](https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/#path-collision-errors-in-embedded-fields), so if a user provided a field with embedded document then the document name should not be automatically put into the `fields`

@abdimo101 How do you think about this suggestion?

nitrosx:
My two cents, 

- we should do lookup before projection.
- if somebody provides the following input:
```
{
  "where": {
    "datasetName": {
      "$regex": "test22222233333334",
      "$options": "i"
    }
  },
  "include": [
    "attachments"
  ],
  "fields": [
    "datasetName",
    "attachments"
  ]
}
```
or 
```
{
  "where": {
    "datasetName": {
      "$regex": "test22222233333334",
      "$options": "i"
    }
  },
  "include": [
    "attachments"
  ],
  "fields": [
    "datasetName",
    "attachments.thumbnail"
  ]
}
```
we know that what to do.
- if a user provides the following input:
```
{
  "where": {
    "datasetName": {
      "$regex": "test22222233333334",
      "$options": "i"
    }
  },
  "include": [
    "attachments"
  ],
  "fields": [
    "datasetName"
  ]
}
```
we can infer that they have forgot to ask for the attachments and internally we should insert `attachments` in the projection as the input passed was like:
```
{
  "where": {
    "datasetName": {
      "$regex": "test22222233333334",
      "$options": "i"
    }
  },
  "include": [
    "attachments"
  ],
  "fields": [
    "datasetName",
    "attachments"
  ]
}
```

f course everything should be documented in the swagger endpoint, including the reference to the path collision mentioned by @Junjiequan

abdimo101:
> > @Junjiequan Thanks for the explanation. Assuming we put the lookup pipeline step in the correct place, what we could add is that any filter under `"include"` is automatically put into the `"fields"` filter as well. I think this implicit assumption that a client who specified `"include"` would also want to see that in the projection is valid.
> 
> That sounds like a better solution. Just one additional thing needs to be clarified, if user provide the query below, there will be an error
> 
> ```json
>   "include": [
>     "attachments"
>   ],
>   "fields": [
>     "datasetName",
>     "attachments.thumbnail"
>   ]
> ```
> 
> Because `You cannot specify both an embedded document and a field within that embedded document in the same projection.` - [doc](https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/#path-collision-errors-in-embedded-fields), so if a user provided a field with embedded document then the document name should not be automatically put into the `fields`
> 
> @abdimo101 How do you think about this suggestion?

Yes to avoid the path collision, we have to make a check to ensure that when a field within an embedded document(e.g `attachments.thumbnail`) has been added, then the embedded document should automatically not be added as well if it's under `include`.

HayenNico (approved):
Looks good

test/DatasetV4.js
```diff
@@ -843,6 +869,54 @@ describe("2500: Datasets v4 tests", () => {
         .expect(TestData.BadRequestStatusCode)
         .expect("Content-Type", /json/);
     });
+
+    it("0309: should only return the field and its subfields within the embedded documents even when 'all' is used under include", async () => {
+      const filter = {
+        include: ["all"],
+        fields: [
+          "datasetName",
+          "attachments.thumbnail",
```
minottic, line 1063: I think this should not be supported, it can already be done with scopes and it will be additional code to maintain

src/attachments/attachments.service.ts
```diff
@@ -65,7 +65,7 @@ export class AttachmentsService {
 
     const pipeline: PipelineStage[] = [{ $match: whereFilter }];
     if (!isEmpty(fieldsProjection)) {
-      const projection = parsePipelineProjection(fieldsProjection);
+      const projection = parsePipelineProjection(fieldsProjection, filter);
```
minottic: I think passing the `filter.include` here would make what the function does clearer

minottic: if we do this (or similar), I think the logic of the function can be greatly simplified. Essentially, every service can pass to it the intersection between its enum (e.g. DatasetLookupKeysEnum) and the relations that are in the filter

minottic (commented):
sorry, I overlooked this. I think now that the "scope" is supported in the dataset endpoint, a user would filter and project using that (see this [test](https://github.com/SciCatProject/scicat-backend-next/blob/master/test/DatasetV4.js#L668-L678)), so the relation fields in the parent I believe should not be supported (also because `scope` enables more than that, e.g. filters and limits on relations). I think all `findOneComplete` should simply reuse the `findAllComplete`, see [here](https://github.com/SciCatProject/scicat-backend-next/pull/2236). My solution started with datasest, I agree this concept should later be applied everywhere. I simpatise defaulting to include relations when in include but not in field, but then I think this is the only thing this PR should do, on top of the scopes and reusing findAllComplete

abdimo101:
Hi @minottic, I had a discussion with @nitrosx & @Junjiequan about your changes. We concluded that the `fields` inside `scope` are not really necessary, as they only apply to fields in lookup documents and not the parent (i.e dataset). The only way to filter parent fields like `datasetName` is to use the `fields` array outside the scope, which means the users would need to specify in two places. 

This becomes even more challenging when multiple relations are involved, as users would have to update several different `fields` arrays. For example:

```
{
  "where": {
    "datasetName": {
      "$regex": "Dataset",
      "$options": "i"
    }
  },
  "include": [
    "origdatablock",
    {
      "relation": "attachments",
      "scope": {
        "fields": [
          "filename",
          "mimetype"
        ],
        "limits": {
          "limit": 5,
          "skip": 0,
          "sort": {
            "filename": "asc"
          }
        },
        "where": {
          "filename": {
            "$regex": "data",
            "$options": "i"
          }
        }
      }
    },
    {
      "relation": "proposals",
      "scope": {
        "fields": [
          "title",
          "abstract"
        ],
        "limits": {
          "limit": 1,
          "skip": 0,
          "sort": {
            "title": "asc"
          }
        },
        "where": {
          "title": {
            "$regex": "ESS",
            "$options": "i"
          }
        }
      }
    }
  ],
  "fields": [
    "datasetName"
  ],
  "limits": {
    "limit": 10,
    "skip": 0,
    "sort": {
      "datasetName": "asc | desc"
    }
  }
} 

``` 

This approach requires users to handle multiple `fields` arrays, we believe it would be simpler for users to only handle one. 

FIrst improvement: 
``` 
{
  "where": {
    "datasetName": {
      "$regex": "Dataset",
      "$options": "i"
    }
  },
  "include": [
    "origdatablock",
    {
      "relation": "attachments",
      "scope": {
        "limits": {
          "limit": 5,
          "skip": 0,
          "sort": {
            "filename": "asc"
          }
        },
        "where": {
          "filename": {
            "$regex": "data",
            "$options": "i"
          }
        }
      }
    },
    {
      "relation": "proposals",
      "scope": {
        "limits": {
          "limit": 1,
          "skip": 0,
          "sort": {
            "title": "asc"
          }
        },
        "where": {
          "title": {
            "$regex": "ESS",
            "$options": "i"
          }
        }
      }
    }
  ],
  "fields": [
    "datasetName"
    "attachments.filename",
    "attachments.mimetype",
    "proposals.title",
    "proposals.abstract"
 ],
  "limits": {
    "limit": 10,
    "skip": 0,
    "sort": {
      "datasetName": "asc | desc"
    }
  }
}
```
And we also talked about implementing this format not in this PR, but in the future : 
```
{
  "where": {
    "datasetName": {
      "$regex": "Dataset",
      "$options": "i"
    }
  },
  "include": [
    "origdatablock",
    {
      "collection": "attachments",
      "limits": {
        "limit": 5,
        "skip": 0,
        "sort": {
          "filename": "asc"
        }
      },
      "where": {
        "filename": {
          "$regex": "data",
          "$options": "i"
        }
      }
    },
    {
      "collection": "proposals",
      "limits": {
        "limit": 1,
        "skip": 0,
        "sort": {
          "title": "asc"
        }
      },
      "where": {
        "title": {
          "$regex": "ESS",
          "$options": "i"
        }
      }
    }
  ],
  "fields": [
    "datasetName"
    "attachments.filename",
    "attachments.mimetype",
    "proposals.title",
    "proposals.abstract"
 ],
  "limits": {
    "limit": 10,
    "skip": 0,
    "sort": {
      "datasetName": "asc | desc"
    }
  }
}
```

minottic:
@abdimo101 fields will be both in scope and in the upper level in my (merged) PR, so the user will need a more complex json, but this will be way closer to the V3 implementation (and backward compatible with the old backend). But I simpatise with having the ability to add fields in the uppermost level as well, and I have to take back that it should not be supported. I think it would be a bit easier if we split the features. 

1. we support already scopes with where and fields, as [loopback 3](https://loopback.io/doc/en/lb3/Include-filter.html) did and as v3 should do (thus the easy extension to v4). See PR #2231 . This includes adding where, limits and fields on relations for datasets
2. we need to implicitely add relations to fields when used in the include (and I thikn this is the only thing this PR should do)
3. we should avoid code duplication, and this is what #2236 does
4. we need to extend parent fields and use `parent.children` in fields. This could either be here or in another PR (I think it's better in another PR, but I am also fine having it here)

btw, this 

> And we also talked about implementing this format not in this PR, but in the future :

is already supported by this PR for datasets #2231. But if you mean here just making the filter syntax a bit easiear, that could work even though I am not sure it's worth asking users to change all their syntax when migrating from v3 to v4 for such a cosmetic change (you are essentially "stripping" the "scope" and renaming "relation" to "collection". For example, see how simple it becomes to extend to v3 here #2237 ). I agree it should be extended to other collections, but as I commented in this PR #2236 I thikn a more complete change is to refactor all find in all controllers to use a shared findAllComplete and findOneComplete that will reuse the pipelines. 

I will be happy to discuss this in a meeting next week

abdimo101:
Hi @minottic, thanks for the quick response. I pushed my last commit before seeing your response, but I’m open to discussing this further next week.

nitrosx:
@minottic I understand that you would like to maintain the same filter syntax in v4. 
I would need more details to understand the context for this decision.
That said, I see dataset v4 as the right chance to move away completely from loopback syntax, which is not intuitive, complex and does not allow complex queries.

minottic (approved):
sorry, I missed the refactor of the PR. Looks good to me, thanks for the change!

PR #2213, opened 2025-09-16
fix(dataset): proper error handling for field projections

abdimo101:
## Description
Fixes the dataset API to throw an error message if users try to use object format.

## Motivation
When users requested fields using object format `({"pid": true, "owner": true})`, it would throw a 500 error because it expected array format only.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -417,13 +417,12 @@ export const parsePipelineSort = (sort: Record<string, "asc" | "desc">) => {
 export const parsePipelineProjection = (fieldsProjection: string[]) => {
   const pipelineProjection: Record<string, boolean> = {};
 
-  if (Array.isArray(fieldsProjection)) {
-    fieldsProjection.forEach((field) => {
-      pipelineProjection[field] = true;
-    });
-  } else {
-    return fieldsProjection;
+  if (!Array.isArray(fieldsProjection)) {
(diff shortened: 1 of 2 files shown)


emigun (approved):
:medal_sports:

src/common/utils.ts
```diff
@@ -416,9 +416,14 @@ export const parsePipelineSort = (sort: Record<string, "asc" | "desc">) => {
...
-    pipelineProjection[field] = true;
-  });
+
+  if (Array.isArray(fieldsProjection)) {
+    fieldsProjection.forEach((field) => {
+      pipelineProjection[field] = true;
+    });
+  } else {
+    return fieldsProjection;
+  }
```
Junjiequan: I'm not sure why we need to provide two methods for the same thing? Can we just keep array and throw proper message to keep it simple

Junjiequan (changes requested):

src/common/utils.ts
abdimo101: Fixed in [bb2b8cd](https://github.com/SciCatProject/scicat-backend-next/pull/2213/commits/bb2b8cda16d38fc971df5b7437d760b47d430b50)

Junjiequan (approved):

PR #2214, opened 2025-09-17
fix: remove 409 on same lifecycle patch

minottic:
## Description
Our archivers people reported 409 makes the lifecycle update more complex. They need to add additional logic on subsequent patches and ues try catch. This PR removes the 409 and runs the same logic on already done updates.

## Changes:

* avoid 409 on same content

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -36,7 +36,6 @@ import {
 import { ClassConstructor, plainToInstance } from "class-transformer";
 import { validate, ValidationError, ValidatorOptions } from "class-validator";
 import { Request } from "express";
-import { isEqual } from "lodash";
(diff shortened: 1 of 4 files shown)


minottic:
@sofyalaski I have just removed the logic, as I understood removing has no side effects. If this is not true, I will be happy to change the logic more

sofyalaski (changes requested):
Looks good, but we also need to do the same in v4 controller:
https://github.com/SciCatProject/scicat-backend-next/blob/f7034d42f5ff99172907b9c72f4dbbcb2f625a78/src/datasets/datasets.v4.controller.ts#L912-L937

sofyalaski (approved):

PR #2215, opened 2025-09-17
refactor(test): move auth from beforeEach to before in api testing

HayenNico:
## Description
This PR moves the authorization logic used in many tests to grab access tokens from the beforeEach to the before context. This significantly accelerates the api testing suites.
In addition, it reformats the heading lines in the API test code to make them uniform and easier to read, and enforce strict mode everywhere.

## Fixes

* Test 1170-0180 (JobAdmin Tests) would fail for the wrong reason

## Changes:

* Test file headers: remove all eslint disable instructions, add `"use strict";` everywhere
* Replace all `var` declarations with `let`
* Group all variable declarations at the top of the file for readability
* Remove some unused variables
* Reduce beforeEach content as much as possible to accelerate tests (except User/Auth tests themselves for safety)

## Tests included

- [ ] Passing? 

## Documentation


minottic (approved):
nice, thanks! FYI I think one can also have multiple subsequent `before` that are executed in order, that might have helped in the refactor

Junjiequan (approved):
That was a long needed cleanup/refactor, thanks!

sofyalaski (approved):
that's an amazing speed up!

PR #2230, opened 2025-09-24
feat(proposal): add numberOfDatasets field and sync on dataset changes with migration script

abdimo101:
## Description
This PR adds a `numberOfDatasets` field to the proposal schema and ensures it is incremented or decremented whenever datasets are created or deleted. It also includes a migration script to calculate and populate the `numberOfDatasets` field for existing proposals.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.service.ts
+++ b/src/datasets/datasets.service.ts
@@ -104,8 +104,10 @@ export class DatasetsService {
 
     const savedDataset = await createdDataset.save();
 
-    if(savedDataset.proposalIds){
-      await this.proposalService.incrementNumberOfDatasets(savedDataset.proposalIds);
+    if (savedDataset.proposalIds) {
+      await this.proposalService.incrementNumberOfDatasets(
+        savedDataset.proposalIds,
+      );
     }
 
     return savedDataset;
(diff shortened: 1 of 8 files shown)


src/datasets/datasets.service.ts
```diff
@@ -99,7 +101,16 @@ export class DatasetsService {
...
     }
-    return createdDataset.save();
+
+    const savedDataset = await createdDataset.save();
+
+    if (savedDataset.proposalIds) {
+      await this.proposalService.incrementNumberOfDatasets(
+        savedDataset.proposalIds,
+      );
+    }
```
Junjiequan, line 161: `savedDataset.proposalIds` is true when its an empty array

test/DatasetV4.js
```diff
@@ -353,6 +352,29 @@ describe("2500: Datasets v4 tests", () => {
...
+      };
+      await request(appUrl)
+        .post("/api/v4/datasets")
+        .send(dataset)
+        .auth(accessTokenAdminIngestor, { type: "bearer" });
+
+      const proposal = await request(appUrl)
+        .get(`/api/v3/Proposals/${encodeURIComponent(proposalId)}`)
+        .auth(accessTokenAdminIngestor, { type: "bearer" });
+      console.log("numberOfDatasets: ", JSON.stringify(proposal.body.numberOfDatasets, null, 2));
```
Junjiequan: Can we add 2 more tests here, one for decrement, and the other to make sure increment do not happen when dataset has no or empty proposals

Junjiequan (changes requested):

Junjiequan (approved):

PR #2231, opened 2025-09-24
feat: enable scoped search on relations

minottic:
## Description
This makes possible to pass a relation with scopes in the V4 endpoint, in addition to a list of relations as before

## Changes:

* openapi descriptor extended to support include: [{relation: abc, scope: {where: {...}, fields: [...]...}}]
* include validation pipe extended to check array of objects 
* addRelation extended to add pipeline stages for scope.where (match), scope.field (project), and limits

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/datasets/interfaces/dataset-filters.interface.ts
+++ b/src/datasets/interfaces/dataset-filters.interface.ts
@@ -1,8 +1,5 @@
 import { FilterQuery } from "mongoose";
-import {
-  ILimitsFilter,
-  IScientificFilter,
-} from "src/common/interfaces/common.interface";
+import { IScientificFilter } from "src/common/interfaces/common.interface";
 import { DatasetLookupKeysEnum } from "../types/dataset-lookup";
 
 export interface IDatasetFields {
@@ -35,5 +32,11 @@ export interface IDatasetFiltersV4<T, Y = null> {
   where?: FilterQuery<T>;
   include?: DatasetLookupKeysEnum[] | IDatasetRelation<T, Y>[];
   fields?: (keyof Y)[];
-  limits?: ILimitsFilter;
+  limits?: IDatasetLimitsV4<T>;
+}
+
+interface IDatasetLimitsV4<T> {
+  limit: number;
+  skip: number;
+  sort: Record<keyof T, "asc" | "desc">;
 }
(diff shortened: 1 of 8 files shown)


src/datasets/interfaces/dataset-filters.interface.ts
```diff
@@ -26,9 +26,14 @@ export interface IDatasetFields {
   [key: string]: unknown;
 }
 
+export interface IDatasetRelation<T, Y = null> {
+  relation: DatasetLookupKeysEnum;
+  scope: IDatasetFiltersV4<T, Y>;
```
HayenNico: Shouldn't the `scope` field match the v4 filters for all possible **non-dataset** collections in an include query - so `IInstrumentFiltersV4`, `IProposalFiltersV4`, `IOrigDatablockFiltersV4`, `IDatablockFiltersV4`, `IAttachmentFiltersV4` and `ISampleFiltersV4` (only some of which currently exist)?

src/datasets/types/dataset-filter-content.ts
```diff
@@ -14,8 +14,29 @@ const FILTERS: Record<"limits" | "fields" | "where" | "include", object> = {
...
+          properties: {
+            relation: {
+              type: "string",
+              example: "attachments",
+            },
+            scope: {
+              type: "object",
+              example: {
+                fields: ["filename", "mimetype"],
+                limits: { limit: 5, skip: 0, sort: { filename: "asc" } },
```
HayenNico, line 33: `addLookupFields` will not process `limits.sort` in the current version

test/DatasetV4.js
```diff
@@ -511,6 +512,84 @@ describe("2500: Datasets v4 tests", () => {
         });
     });
 
+    it("0204.0: should fetch dataset relation fields if provided in the filter as obj", async () => {
```
HayenNico: I think we should keep the test numbering to four-digit codes

HayenNico (changes requested):
Looks good, just a couple minor nitpicks from my side

src/datasets/interfaces/dataset-filters.interface.ts
minottic: thanks. The type fix was a bit challenging, I have to admit...

HayenNico (approved):
Looks great - and the generic `IFiltersV4` interface will be very useful when adding V4 API to the remaining subsystems

PR #2236, opened 2025-09-25
chore: refactor findOneComplete to avoid dup

minottic:
## Description
Avoids logic duplication and reuses existing more generic function

## Changes:

* refactor findOneComplete

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.service.ts
+++ b/src/datasets/datasets.service.ts
@@ -332,49 +332,19 @@ export class DatasetsService {
   }
 
   async findOneComplete(
-    filter: FilterQuery<DatasetDocument>,
+    filter: IDatasetFiltersV4<DatasetDocument, IDatasetFields>,
   ): Promise<OutputDatasetDto | null> {
-    const whereFilter: FilterQuery<DatasetDocument> = filter.where ?? {};
-    let fieldsProjection: string[] = filter.fields ?? {};
-
-    const limits: QueryOptions<DatasetDocument> = filter.limits ?? {
+    filter.limits = filter.limits ?? {
       skip: 0,
-      sort: { createdAt: "desc" },
+      sort: { createdAt: "desc" } as Record<
+        keyof DatasetDocument,
+        "asc" | "desc"
+      >,
     };
 
-    const pipeline: PipelineStage[] = [{ $match: whereFilter }];
-    const addedRelations = this.addLookupFields(pipeline, filter.include);
+    const [data] = await this.findAllComplete(filter);
 
-    if (Array.isArray(fieldsProjection)) {
-      fieldsProjection = Array.from(
-        new Set([...fieldsProjection, ...addedRelations]),
-      );
-    }
-
-    if (!isEmpty(fieldsProjection)) {
-      const projection = parsePipelineProjection(fieldsProjection);
-      pipeline.push({ $project: projection });
-    }
-
-    if (!isEmpty(limits.sort)) {
-      const sort = parsePipelineSort(limits.sort);
-      pipeline.push({ $sort: sort });
-    }
-
-    pipeline.push({ $skip: limits.skip || 0 });
-
-    try {
-      const [data] = await this.datasetModel
-        .aggregate<OutputDatasetDto | undefined>(pipeline)
-        .exec();
-
-      return data || null;
-    } catch (error) {
-      if (error instanceof Error) {
-        throw new BadRequestException(error.message);
-      }
-      throw new BadRequestException("An unknown error occurred");
-    }
+    return (data as OutputDatasetDto) || null;
   }
 
   async count(
--- a/test/DatasetV4.js
+++ b/test/DatasetV4.js
@@ -853,6 +853,7 @@ describe("2500: Datasets v4 tests", () => {
           skip: 0,
           sort: {
             datasetName: "asc",
+            createdAt: "asc",
           },
         },
       };
--- a/src/common/interfaces/common.interface.ts
+++ b/src/common/interfaces/common.interface.ts
@@ -35,9 +35,9 @@ export interface IProposalAcceptedMessage {
 }
 
 export interface ILimitsFilter {
-  limit: number;
-  skip: number;
-  order: string;
+  limit?: number;
+  skip?: number;
+  order?: string;
 }
 
 export interface IFilters<T, Y = null> {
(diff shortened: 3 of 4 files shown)


emigun:
Looks good to me! Great to avoid duplicated functionality.

(However, I don't fully understand the consequences of your type changes, so I'll let someone else approve)

test/DatasetV4.js
```diff
@@ -875,7 +875,6 @@ describe("2500: Datasets v4 tests", () => {
           },
         },
         include: ["all"],
-        fields: ["datasetName", "pid"],
```
HayenNico, line 1011: Why remove the fields projection as a valid queryFilter?

src/datasets/datasets.service.ts
```diff
@@ -194,7 +194,7 @@ export class DatasetsService {
 
     pipeline.push({ $skip: limits.skip || 0 });
 
-    pipeline.push({ $limit: limits.limit || 10 });
+    if (limits?.limit) pipeline.push({ $limit: limits.limit });
```
HayenNico: Removing the default changes the behavior in an edge case for the `/api/v4/datasets` (former fullquery) endpoint, when `queryFilter.limits` exists but does not include `queryFilter.limits.limit`. Before the response would always be limited to 10 datasets, with this the response could have arbitrarily many datasets.

I'm not opposed to the change, but then we should be consistent about not assuming default filters

HayenNico (changes requested):
All type changes look good, but I had a couple questions (see below).

This also conflicts with fixes to `findOneComplete` in #2212 so we need to pick an approach

test/DatasetV4.js
minottic: I think this test (as the other) was wrong. It does not make sense to ask for these fields and then expect different fiedls to be present in the response IMO

minottic:
> All type changes look good, but I had a couple questions (see below).
> 
> This also conflicts with fixes to findOneComplete in https://github.com/SciCatProject/scicat-backend-next/pull/2212 so we need to pick an approach

In principle my PR (a part from the defaults use case that I'll address) only reuses findAllComplete in findOneComplete. This avoids duplication and is conceptually more consistent I think as findOneComplete is just a special case of the findAllComplete. I also think is more intuitive for the user that the 2 endpoints work the same. The difference was only  the order of the project stage, which in findOne was applied before the relation, thus I had to change the test. I think for a user this would not make sense. If I ask for the "instrument" relation and then don't include "instrument" in the fields it is expected I believe "instruments" will not be there.

So I would lean towards my PR (a bit biased :) ), as a start and then apply the changes from #2212 as I added in the comment there

Junjiequan:
If the findOneComplete query can be done with the findAllComplete query, I don’t see why we need it at all. 
From the code, they look exactly the same. For findOne, I’d expect to see { where: { _id: pid } }, and for findAll, I’d expect { where: { some set of filters } }.

minottic:
it's just a utility that sets the filters and this allows me not to change where findOneComplete was historically used. I am not sure why this was ever implemented, but if you follow the old logic (or check the changes of this PR), it does exactly the same. BTW the add `addLookupFields` is also duplicated in a lot of services and can/should be reused, in a later PR.

Since it's a findOne and not findById, it does not set the ID. I agree the findById could reuse this. When reading the code, I think it would be way easier if for every controller, for the GETs we had essentially 1 function in service reused in all controllers, i.e. this findAllComplete with `mongo aggregate`. Then we can use it everywwhere, have 1 single entrypoint and even generalize it to work in v3 and v4 (see [here](https://github.com/SciCatProject/scicat-backend-next/pull/2237)). I think this would massively improve maintainability

test/DatasetV4.js
sbliven: @HayenNico The 'fields' filter is still present, as tested in other tests (eg 0207). It's used by `findAllComplete`, so it doesn't need special treatment in `findOneComplete`.

sbliven: Maybe we should keep this test though to document how `fields` interacts with `includes`. I guess the previous test was trying to show that the `fields` property only applied to dataset fields, and is overridden by `includes`

HayenNico: I also think this test case should stay in. It is true that the test doesn't work as expected for `fields` based on the current master branch, since the lookup was done differently in `findOneComplete` and `findAllComplete` (different ordering in aggregation pipeline, definitely a bug). This part was adressed in #2212, so at least the fixes for tests from that PR should be carried over to here.

HayenNico:
> > All type changes look good, but I had a couple questions (see below).
> > This also conflicts with fixes to findOneComplete in #2212 so we need to pick an approach
> 
> In principle my PR (a part from the defaults use case that I'll address) only reuses findAllComplete in findOneComplete. This avoids duplication and is conceptually more consistent I think as findOneComplete is just a special case of the findAllComplete. I also think is more intuitive for the user that the 2 endpoints work the same. The difference was only the order of the project stage, which in findOne was applied before the relation, thus I had to change the test. I think for a user this would not make sense. If I ask for the "instrument" relation and then don't include "instrument" in the fields it is expected I believe "instruments" will not be there.
> 
> So I would lean towards my PR (a bit biased :) ), as a start and then apply the changes from #2212 as I added in the comment there

I'm on board with this refactor, but I think at least part of #2212 (the bugfix for `findOneComplete` which led to inconsistent behaviour) should be merged first. A refactor should ideally not touch the tests at all to ensure full compatibility.

@nitrosx @Junjiequan @abdimo101 Do you think it would be better to split #2212 into two PRs, one for the `findOneComplete` bugfix + testing (which can be merged quickly), and a second for the implicit include logic added? The discussion between using scopes or dot-notation seems like it may take some extra time, which shouldn't block this PR

Junjiequan:
@HayenNico @nitrosx @abdimo101  I think it's better to merge the PR with bug fix (use correct lookup sequence for all controllers), add corresponding tests and then open an issue to discuss enhanced include logic later on

minottic:
@Junjiequan , @HayenNico @abdimo101 , sorry for the wait, I was carried away... I should have addressed the comments and this is only a rebase now that @abdimo101 has fixed the includes issue

HayenNico (approved):
Looks good 👍

PR #2237, opened 2025-09-26
feat: add scopes to v3 datasets

minottic:
## Description

Reuse findOneComplete and findOneAll for v3 endpoints to enable scopes. 

## Motivation
Reintroduce loopback3 features to v3 endpoints (removing the support is non backward compatible change), by few findAllComplet branching

## Changes:

* findAll, findById and findOne in v3 all use the same function now. Avoids dup, shared with v4, same expected behaviour for the user
* richer swagger for v3 endpoints
* v3 findById supports filtering now
* types extended to union v3 and v4

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -117,9 +117,9 @@ import { TechniqueClass } from "./schemas/technique.schema";
 import { DatasetType } from "./types/dataset-type.enum";
 import { HistoryService } from "src/history/history.service";
 import { convertGenericHistoriesToObsoleteHistories } from "src/datasets/utils/history.util";
-import { getSwaggerDatasetFilterContent } from "./types/dataset-filter-content";
 import { IncludeValidationPipe } from "src/common/pipes/include-validation.pipe";
 import { DATASET_LOOKUP_FIELDS } from "./types/dataset-lookup";
+import { getSwaggerDatasetFilterContentV3 } from "./types/dataset-filter-content.v3";
 
 @ApiBearerAuth()
 @ApiExtraModels(
@@ -902,7 +902,7 @@ export class DatasetsController {
       filterDescription,
     required: false,
     type: String,
-    content: getSwaggerDatasetFilterContent(undefined, "v3"),
+    content: getSwaggerDatasetFilterContentV3(),
   })
   @ApiResponse({
     status: HttpStatus.OK,
@@ -931,7 +931,9 @@ export class DatasetsController {
     );
     return Promise.all(
       (datasets as DatasetDocument[]).map((dataset) =>
-        this.convertCurrentToObsoleteSchema(dataset),
+        this.convertCurrentToObsoleteSchema(
+          this.datasetsService.hydrate(dataset),
+        ),
       ),
     );
   }
@@ -1195,7 +1197,7 @@ export class DatasetsController {
       filterDescription,
     required: false,
     type: String,
-    content: getSwaggerDatasetFilterContent(undefined, "v3"),
+    content: getSwaggerDatasetFilterContentV3(),
   })
   @ApiResponse({
     status: HttpStatus.OK,
@@ -1276,7 +1278,7 @@ export class DatasetsController {
       filterDescription,
     required: false,
     type: String,
-    content: getSwaggerDatasetFilterContent(undefined, "v3"),
+    content: getSwaggerDatasetFilterContentV3(),
   })
   @ApiResponse({
     status: HttpStatus.OK,
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -635,7 +635,7 @@ export class DatasetsController {
     }
 
     const outputDataset: OutputDatasetObsoleteDto = {
-      ...(inputDataset as DatasetDocument).toObject(),
+      ...((inputDataset as DatasetDocument).toObject?.() ?? inputDataset),
       ...propertiesModifier,
     };
 
@@ -934,9 +934,7 @@ export class DatasetsController {
     );
     return Promise.all(
       (datasets as DatasetDocument[]).map((dataset) =>
-        this.convertCurrentToObsoleteSchema(
-          this.datasetsService.hydrate(dataset),
-        ),
+        this.convertCurrentToObsoleteSchema(dataset),
       ),
     );
   }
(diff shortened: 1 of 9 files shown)


minottic:
depends on #2236 and it's opened against that branch. Will rebase after merge

nitrosx (changes requested):
IMHO, using hardcoded "v3" and "v4" values in the functions calls will make the code less transparent, hard to manage and harder to separate by versions later on.
I would like to propose the creation of separated `datasets.service.ts` for v3 and v4.

minottic:
fine for me, so these services will then do the branching in your proposal?

nitrosx:
@minottic if I understand your question correctly, I would create a service v3 which will be used by the all the endpoints of dataset api v3 and a service v4 which will be used by all the dataset endpoints v4.
It will introduce some code duplication for now, but it will make the code easier to read and it will be easier to remove obsolete V3 in the future

minottic:
ok, I will make a new commit for that

minottic:
@nitrosx please see with the latest changes, sorry for the wait!

src/datasets/interfaces/dataset-filters.interface.ts
```diff
@@ -42,21 +43,43 @@ export interface IDatasetFields {
...
 
 export type IDatasetFiltersV4<T, Y = null> = IFiltersV4<
   T,
   Y,
-  DatasetLookupKeysEnum[] | IDatasetRelation[]
+  DatasetLookupKeysEnum[] | IDatasetRelationV4[]
 >;
+
+export type IDatasetFiltersV3<T, Y = null> = Omit<IFilters<T, Y>, "include"> & {
+  include: DatasetLookupKeysEnum[] | IDatasetRelationV4<IDatasetScopesV3>[];
```
omkar-ethz: I believe this type should be `(DatasetLookupKeysEnum | IDatasetRelationV4<IDatasetScopesV3>)[]`, as the elements of the include array can be of both types, rather than all of include array being of one of the types.

src/datasets/datasets.controller.ts
```diff
@@ -1357,12 +1290,20 @@ export class DatasetsController {
...
+    @Query(new FilterPipe()) queryFilter: { filter?: string },
+  ) {
+    const filterObj = JSON.parse(queryFilter.filter ?? "{}");
+    filterObj.where = filterObj.where ?? {};
+    filterObj.where.pid = id;
+    const dataset = await this.findAll(request, headers, {
+      filter: JSON.stringify(filterObj),
+    });
+    if (dataset.length == 0) throw new ForbiddenException();
+    return dataset[0] as OutputDatasetObsoleteDto;
```
omkar-ethz: I think this could also be a `NotFoundException`. Before, `checkPermissionsForDatasetObsolete` would throw the appropriate exception. But not sure how to distinguish it from `ForbiddenException` in findAll.. Maybe it's still worth checking like in checkPermissionsForDatasetObsolete whether the dataset exists first?

src/datasets/datasets.controller.ts
```diff
@@ -1242,51 +1209,8 @@ export class DatasetsController {
...
-              });
-              break;
-            }
-          }
-        }),
-      );
-    }
-    return outputDataset;
+    const dataset = await this.findAll(request, headers, queryFilter);
+    return dataset[0] as OutputDatasetObsoleteDto;
```
omkar-ethz, line 1216: 1. Would it be possible to add `{limit: 1}` to the queryFilter? Otherwise we may retrieve multiple datasets unecessarily
2. If dataset is [], right now the user gets null JSON, i think it's ok, just making sure that's the expected behavior and not a 404.

src/datasets/datasets.controller.ts
```diff
@@ -915,61 +920,23 @@ export class DatasetsController {
...
-    }
-    return outputDatasets as OutputDatasetObsoleteDto[];
+    const datasets = await this.datasetsService.findAllComplete(
+      mergedFilters,
+      false,
+    );
+    return Promise.all(
+      (datasets as DatasetDocument[]).map((dataset) =>
+        this.convertCurrentToObsoleteSchema(
+          this.datasetsService.hydrate(dataset),
```
omkar-ethz: I think this is making `fields` filter ineffective, as it adds back in all the projected out fields.

omkar-ethz (approved):
lgtm!

src/datasets/datasets.controller.ts
```diff
@@ -1357,12 +1293,22 @@ export class DatasetsController {
...
+    await this.findOrThrow(id);
+    const filterObj = JSON.parse(queryFilter.filter ?? headers.filter ?? "{}");
+    filterObj.where = filterObj.where ?? {};
+    filterObj.where.pid = id;
+    const dataset = await this.findAll(request, headers, {
+      filter: JSON.stringify(filterObj),
+    });
+    if (dataset.length == 0)
+      throw new ForbiddenException("Unauthorized access");
+    return dataset[0] as OutputDatasetObsoleteDto;
```
Junjiequan, line 1311: I don't understand the change here.. Why do we need headers and queryFilter for findById. Isn't it suppose to accept param only?

Junjiequan (dismissed):

src/datasets/datasets.controller.ts
minottic: people use this (at least the query) to specify `fields` in findById here

minottic: and it's compatible with the old spec: https://dacat.psi.ch/explorer/#!/Dataset/Dataset_findById 

(I can remove the header indeed)

Junjiequan: I see. It just feels wrong to let findById to accep query filter.. It's like saying,  "give me that apple but that apple should be an orange"
I will not block this PR, but it would be nice to avoid this logic with v4

minottic: I agree. For us the struggle really is that the v3 endpoints are not backwards compatible, thing that we claim to some extent, and quite some people are complaining here during migration.

> It would be nice to avoid this logic with v4

I agree. We might still want to allow "fields" to limit the json properties

nitrosx (approved):
As long as the _**include relationship**_ does not get in dataset v4 and this is absolutely necessary for PSI to go in production, I'm ok to merge

minottic:
> As long as the include relationship does not get in dataset v4 and this is absolutely necessary for PSI to go in production, I'm ok to merge

to reiterate the comment on the thread, this is not PSI specific, this is to make v3 dataset endpoints backward compatible

PR #2238, opened 2025-09-26
fix: allow own to see emails from profile when masking

minottic:
## Description
When masking persona info, the email of the user in the UserModel can be different from emails in its identities, UserIdentity model. This PR merges all emails from user and identities and checks if any is equal to the one of the intercepted response

## Fixes

* get emails from identities
* mask only if not in identities

## Tests included

- [x] Passing? 

## Documentation


sbliven:
I'm still seeing my username masked in `api/v3/auth/whoami`. Does that relate to this PR?

fpotier (approved):

PR #2240, opened 2025-09-29
chore: update frontend config

Junjiequan:
## Description
Update configuration for new proposal table https://github.com/SciCatProject/frontend/pull/2014

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/config/frontend.config.json
+++ b/src/config/frontend.config.json
@@ -267,7 +267,7 @@
       {
         "name": "startTime",
         "type": "date",
-        "format": "yyyy/MM/dd",
+        "format": "yyyy-MM-dd",
         "width": 200,
         "enabled": true
       },
--- a/src/config/frontend.config.json
+++ b/src/config/frontend.config.json
@@ -3,6 +3,7 @@
     "nonAuthenticatedUser": "DATASETS",
     "authenticatedUser": "PROPOSALS"
   },
+  "checkBoxFilterClickTrigger": false,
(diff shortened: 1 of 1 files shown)


HayenNico (approved):
Looks good

One comment I had: I think single-selection and multi-selection `checkbox` elements should become different filter types once you start defining more (so for example `checkbox` and `checkboxMulti`) instead of being controlled by a central `checkBoxFilterClickTrigger` flag, so the behaviour can be defined case-by-case for each filter.

Junjiequan:
@HayenNico Initially we planned to have only multi-select checkboxes, but switched midway to support immediate filter trigger — which is why `checkBoxFilterClickTrigger` exists... 

Personally, I’d prefer to keep just one option (checkbox or checkboxMulti) to reduce complexity, since each filter type requires its own dedicated FE component to handle different behaviors.

HayenNico:
> @HayenNico Initially we planned to have only multi-select checkboxes, but switched midway to support immediate filter trigger — which is why `checkBoxFilterClickTrigger` exists...
> 
> Personally, I’d prefer to keep just one option (checkbox or checkboxMulti) to reduce complexity, since each filter type requires its own dedicated FE component to handle different behaviors.

I'm okay with this solution, I just made the comment since you mentioned in yesterday's meeting that you'd try to make more reusable/configurable filter components with corresponding type options available in the FE at a later point. So my point was just for future reference

PR #2249, opened 2025-10-02
fix: respect the `PublishedData.status` when creating with /api/v3

fpotier:
## Description

Improve the mapping between v3 and v4 `status` and no longer default to `PublishedDataStatus.PRIVATE`

## Motivation

Importing PublishedData that are already registered from other repositories 

## Fixes

Creating a PublishedData with the v3 api was not respecting the `status` in the DTO and always creating the instance with
`PublishedDataStatus.PRIVATE`

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/published-data/published-data.controller.spec.ts
+++ b/src/published-data/published-data.controller.spec.ts
@@ -5,6 +5,7 @@ import { AttachmentsService } from "src/attachments/attachments.service";
 import { CaslAbilityFactory } from "src/casl/casl-ability.factory";
 import { DatasetsService } from "src/datasets/datasets.service";
 import { ProposalsService } from "src/proposals/proposals.service";
+import { PublishedDataStatus } from "./interfaces/published-data.interface";
 import { PublishedDataController } from "./published-data.controller";
 import { PublishedDataService } from "./published-data.service";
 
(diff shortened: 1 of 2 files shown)


minottic (approved):
LGTM thanks! I think we should try to leverage more on class transformers rather than having these conversion functions. Basically, the DTOs would tranform the schema or the payload according to V3 or V3 definitions

fpotier:
> LGTM thanks! I think we should try to leverage more on class transformers rather than having these conversion functions. Basically, the DTOs would tranform the schema or the payload according to V3 or V3 definitions

Maybe I'm missing something but I wouldn't know how to use class transformer for this, my understanding is that class-transformer is only useful to convert plain JS objects to class instances but I might be wrong.
Otherwise I found this library that has integration with nest and could make the conversion between v3 and v4 dtos easier? 
https://automapperts.netlify.app/docs/getting-started/overview

minottic:
ah ok, maybe that library is indeed better. I think with class transformers one can achieve the same by specifying the `@transform` and then do instanceToPlain(plainToInstance(X)) which would result in a plain object but with the transformations. I think it's what nestjs does with it when annotating endpoints that in the end return JSONs

PR #2269, opened 2025-10-09
chore: refactor pipelines for syntax compatible with mongo <5

minottic:
## Description
The current aggregation pipelines used simplified mongo>5 syntax which allowed both "joining" on foreignKeys and have pipelines. This is not allowed in mongo <5, thus this change avoids using foreignKeys and does everything in match stage.

## Changes:

* all foreignKeys removed
* add match pipeline to replace join logic
* avoid overriding the pipeline when adding permissions

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets-access.service.ts
+++ b/src/datasets/datasets-access.service.ts
@@ -149,8 +149,9 @@ export class DatasetsAccessService {
     if (access) {
       const { canViewAny, canViewAccess, canViewOwner } = access;
       if (!canViewAny) {
+        let pipeline: PipelineStage.Lookup["$lookup"]["pipeline"];
         if (canViewAccess) {
-          fieldValue.$lookup.pipeline = [
+          pipeline = [
             {
               $match: {
                 $or: [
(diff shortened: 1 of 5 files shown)


omkar-ethz (approved):
LGTM! As discussed, perhaps create an issue to switch back to new syntax in the future, and/or run tests against mongo < 5 in backend-next / scicatlive CI.

minottic:
thanks! @Junjiequan, @nitrosx  this PR changes the mongo pipelines syntax in order to be mongo <5 and >=5 compatible. At PSI we need that since we run a quite old mongo, and, even if we plan an upgrade, did not want to fork the backend in the meantime. we could claim compatibility with mongo < 5 (in scicat backend) if we added mongo4 in the CI matrix during tests. If this is interesting to you, I can make the CI PR

minottic:
> perhaps create an issue

done [here](https://github.com/SciCatProject/scicat-backend-next/issues/2294)

PR #2270, opened 2025-10-10
fix: dataset sourceFolderHost should accept URLs

### LINKED ISSUE #2235: `sourceFolderHost` validation is incorrect
## Summary
The validation of the property `Dataset.sourceFolderHost` is incorrectly using `IsFQDN` (fails when a protocol is included) while the description states

> DNS host name of file server hosting sourceFolder, optionally including a protocol e.g. [protocol://][fileserver1.example.com](http://fileserver1.example.com/)

## Steps to Reproduce

Try to create a dataset with 
```json
...
  "sourceFolderHost": "https://www.desy.de/"
...
```
## Current Behaviour
Validation fails
```json
{
  "message": ["sourceFolderHost must be a valid domain name"], 
  "error": "Bad Request", 
  "statusCode": 400
}
```

## Expected Behaviour

Dataset created

cfelder:
IsFQDN does not allow the optional [protocol://]fqdn definition

Fixes: #2235

Change-Id: I3e02524051f13f67b4b065f662a77c03dcb6acb0


fpotier (approved):
Maybe you could add the property to existing test data with and without the scheme to make sure both work

PR #2271, opened 2025-10-10
feat: allow url actions to ignore errors from remote calls

sbliven:
## Description

This improves the 'url' job action. Previously, errors from the remote url would always be passed through to the client calling scicat. This adds an `ignoreErrors` option for cases where a failure of the url isn't critical for handling the job.

All errors are logged. This PR also improves logging by logging the response body, which can help debugging.

## Changes:

* allow url actions to ignore errors from remote calls

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/config/job-config/actions/urlaction/urlaction.spec.ts
+++ b/src/config/job-config/actions/urlaction/urlaction.spec.ts
@@ -39,7 +39,7 @@ describe("URLJobAction", () => {
       text: jest.fn().mockResolvedValue("OK"),
     });
 
-    const context = { request: job, job, env: process.env };
+    const context = { request: job, job, env: process.env, datasets: [] };
     await action.perform(context);
 
     expect(global.fetch).toHaveBeenCalledWith(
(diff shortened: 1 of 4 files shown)


src/config/job-config/actions/urlaction/urlaction.spec.ts
```diff
@@ -80,4 +80,33 @@ describe("URLJobAction", () => {
...
+      text: jest.fn().mockResolvedValue("Internal Server Error"),
+    });
+
+    const actionIgnore = new URLJobAction<CreateJobDto>({
+      ...config,
+      ignoreErrors: true,
+    });
+
+    const context = { request: job, job, env: process.env, datasets: [] };
+    await expect(actionIgnore.perform(context)).resolves.toBeUndefined();
```
minottic, line 98: what status code does this return? Should the status code check be included in the test?

minottic (commented):
minor comment, happy to approve after that

src/config/job-config/actions/urlaction/urlaction.spec.ts
sbliven: Actions don't set the status (except by throwing an HTTPException). The fact that it resolves successfully means it will get the success response specified in the controller (eg 201 for POST jobs).

It does appear this is not tested (`jobs.controller.spec.ts` is basically a stub), but I think testing the controller should be a different PR.

minottic (approved):

src/config/job-config/actions/urlaction/urlaction.spec.ts
minottic: ah, thansk for the explanation, my bad!

sbliven: More controller tests requested in #2280

PR #2272, opened 2025-10-13
fix: change hardcoded admin to ADMIN_GROUPS

minottic:
## Description
The permissions check checked against hardcoded "admin" is now replaced with ADMIN_GROUPS. This made all non "admin" users but are in the ADMIN_GROUPS env var receive 403 on job creation

## Fixes

* replace "admin" with ADMIN_GROUPS

## Tests included

- [ ] Passing? 

## Documentation


sofyalaski (approved):
Good catch!

PR #2281, opened 2025-10-13
feat: email actions improvements

sbliven:
## Description

Following #2271, this adds a `ignoreErrors` configuration option to the email action for cases where you don't want an error sending the email to block the job POST or PATCH request.

It also makes other small improvements and bug fixes regarding emails.

## Motivation

Generating the email can cause errors, either from templating or from sending the URL itself. This allows jobs to continue despite the error. Errors are logged in all cases.

## Fixes

* improve error message in the switch regex case
* fix a bug where job properties were not available to templates in some cases

## Changes:

* allow email actions to ignore errors from remote calls
* add an example job config for sending emails only for finished jobs
* add 'matches' and 'default' handlebars helpers

## Tests included

- [x] Passing? 

## Documentation


jobConfig.example.yaml
```diff
@@ -82,21 +82,37 @@ jobs:
...
-          subject: "[SciCat] {{ job.jobParams.subject }}"
-          bodyTemplateFile: demo_email.html
+          subject: |-
+            [SciCatLive] {{ default job.jobParams.subject "Job has been updated"}}
+          bodyTemplateFile: src/common/email-templates/job-template-simplified.html
     update:
       auth: "#jobOwnerUser"
-      actions: []
+      actions:
+        - # Only send emails for updates to status 'finished*'
```
sbliven, line 96: Should this be added to jobConfig.recommended.yaml as well, since sending email is such a common task?

src/config/job-config/handlebar-utils.ts
```diff
@@ -1,21 +1,20 @@
...
-  allowedProtoProperties: {
-    id: true,
-    type: true,
-    statusCode: true,
-    statusMessage: true,
-    createdBy: true,
-    jobParams: true,
-    contactEmail: true,
-  },
-  allowProtoPropertiesByDefault: false, // limit accessible fields for security
```
sbliven, line 14: There were two possible solutions to the problem of template variable coming from mongo (JobClass, DatasetClass, etc). The previous solution was whitelisting properties which we wanted to support. This wasn't maintained (now they are nested under `job`) so this broke. Instead we call `toObject` on mongo classes before adding them to the template context.

Another solution would have been to just allow all ProtoProperties here, and then we wouldn't need the overhead of toObject (at least for handlebars). I didn't feel confident about the security ramifications of that, and we might still need toObject for JSONPath calls (which also fail on ProtoProperties).

sbliven:
This is ready to review. I am still working on a new job-template.html file, but it will come in a fresh PR.

src/config/job-config/actions/emailaction/emailaction.spec.ts
minottic, line 1: would it make sense to test some of the handlers here?

src/app.module.ts
```diff
@@ -131,15 +124,7 @@ import { MaskSensitiveDataInterceptorModule } from "./common/interceptors/mask-s
...
-            adapter: new HandlebarsAdapter({
-              unwrapJSON: unwrapJSON,
-              keyToWord: formatCamelCase,
-              eq: (a, b) => a === b,
-              jsonify: jsonify,
-              job_v3: job_v3,
-              urlencode: urlencode,
-              base64enc: base64enc,
-            }),
+            adapter: new HandlebarsAdapter(handlebarsHelpers),
```
minottic, line 127: for a later PR maybe, should we move the content of `useFactory` to a dedicated mailer factory class/file?

src/config/job-config/handlebar-utils.ts
```diff
@@ -31,7 +30,6 @@ export function compileJobTemplate(
   const template: TemplateJob = hb.compile<JobTemplateContext>(input, options);
   return (context: JobTemplateContext, options?: RuntimeOptions) => {
     return template(context, {
-      ...jobTemplateOptions,
```
minottic, line 34: was this removed on purpose?

minottic (approved):
LGTM, thanks! some minor comments for eventual later PRs

src/config/job-config/handlebar-utils.ts
sbliven: Yes. It's not needed since we're calling toObject everywhere now

src/config/job-config/actions/emailaction/emailaction.spec.ts
sbliven: Which handlers do you mean?

minottic: I meant to have a template that has some handlers inside, e.g. "eq" ro "regex" or similar and test that the email is rendered correctly. But this could indeed be part of a later PR, if useful

src/app.module.ts
sbliven: Agreed. It is getting rather long. It might make sense to wrap handlebars too. Currently you need to remember to register helpers if you call hb directly from outside of a module (eg during testing).

Issue #2248, opened 2025-10-01
Include history in /api/v3/datasets


sbliven:
## Summary

The old dataset schema included a `history` array. Calls to V3 dataset endpoints should populate this from `HistoryClass` instances when present.

In v4 calls the history should not be present by default, but only with an `"include": ["history"]` filter. 

## Details

Relates to #584

sbliven:
This is closely related to #584 (might even be a duplicate)

PR #2282, opened 2025-10-14
feat: add backwards compatible history to v3/datasets response

omkar-ethz:
## Description

Add history array to v3/datasets response, with the same format as backend-v3. The main difference between GenericHistory and HistoryClass format is that in the former, changed fields are grouped into `before` and `after`. Whereas in the latter, each field is present at the top level with `field: {previousValue: ..., currentValue: ...}`.

Another difference is with `Delta` strategy, the current history implementation only records fields that are acutally changed. Whereas in old backend, for `Dataset.datasetlifecycle`, all fields from previous value of datasetlifecyle were recorded in `previousValue`.

## Motivation
Archive/retrieve workflows at PSI depend on backend-v3 representation of History

## Fixes

* https://github.com/SciCatProject/scicat-backend-next/issues/2248
* Update return type in `dataset.service` from `DatasetClass` to `DatasetDocument` 

## Changes:

* Created adapter from  `GenericHistory` to `HistoryClass`
* Created adapter from `(GenericHistory[], Dataset)` -> `HistoryClass[]` that preserves older datasetlifecycle in `previousValue`

## Tests included

- [x] Passing? 

## Documentation
- [x] swagger documentation updated (required for API changes) - auto updated `OutputDatasetObsoleteDto` to include the `history[]` array.

### Sample output

<details>
<summary>GET /datasets/{pid}</summary>

```json
{
...
  "principalInvestigator": "scicatingestor@your.site",
  "history": [
    {
      "updatedAt": "2025-10-14T08:05:10.202Z",
      "updatedBy": "ingestor",
      "_id": "",
      "keywords": {
        "previousValue": [
          "new",
          "keyword"
        ],
        "currentValue": [
          "new",
          "keyword",
          "trois"
        ]
      },
      "description": {
        "previousValue": "This is a test description just for change",
        "currentValue": "one last test description"
      }
    },
    {
      "updatedAt": "2025-10-14T07:37:04.757Z",
      "updatedBy": "ingestor",
      "_id": "",
      "isPublished": {
        "previousValue": true,
        "currentValue": false
      }
    },
    {
      "updatedAt": "2025-10-14T07:23:28.580Z",
      "updatedBy": "ingestor",
      "_id": "",
      "keywords": {
        "previousValue": [],
        "currentValue": [
          "new",
          "keyword"
        ]
      },
      "description": {
        "previousValue": "This is a test description what noooo",
        "currentValue": "This is a test description just for change"
      }
    },
    {
      "updatedAt": "2025-10-14T07:15:34.684Z",
      "updatedBy": "ingestor",
      "_id": "",
      "description": {
        "previousValue": "This is a test description back on v3-history",
        "currentValue": "This is a test description what noooo"
      }
    },
    {
      "updatedAt": "2025-10-14T07:15:24.809Z",
      "updatedBy": "ingestor",
      "_id": ""
    },
    {
      "updatedAt": "2025-10-14T07:00:24.262Z",
      "updatedBy": "ingestor",
      "_id": "",
      "keywords": {
        "previousValue": [
          "newtag"
        ],
        "currentValue": []
      },
      "description": {
        "previousValue": "This is a test description hello HEAD^2",
        "currentValue": "This is a test description back on v3-history"
      }
    },
    {
      "updatedAt": "2025-10-14T06:57:00.237Z",
      "updatedBy": "ingestor",
      "_id": "",
      "keywords": {
        "previousValue": [],
        "currentValue": [
          "newtag"
        ]
      }
    },
    {
      "updatedAt": "2025-10-14T06:56:44.914Z",
      "updatedBy": "ingestor",
      "_id": "",
      "description": {
        "previousValue": "This is a test description hello",
        "currentValue": "This is a test description hello HEAD^2"
      }
    },
    {
      "updatedAt": "2025-10-13T13:59:47.634Z",
      "updatedBy": "ingestor",
      "_id": ""
    },
    {
      "updatedAt": "2025-10-13T13:59:17.508Z",
      "updatedBy": "ingestor",
      "_id": "",
      "description": {
        "previousValue": "This is a test description hola",
        "currentValue": "This is a test description hello"
      }
    },
    {
      "updatedAt": "2025-10-13T13:57:50.173Z",
      "updatedBy": "ingestor",
      "_id": "",
      "description": {
        "previousValue": "This is a test description hello",
        "currentValue": "This is a test description hola"
      }
    }
  ]
}
```

</details>

### DIFF
--- a/src/common/utils/history.util.ts
+++ b/src/common/utils/history.util.ts
@@ -1,6 +1,9 @@
 import { HistoryClass } from "src/datasets/schemas/history.schema";
 import { GenericHistory } from "../schemas/generic-history.schema";
-import { DatasetClass } from "src/datasets/schemas/dataset.schema";
+import {
+  DatasetClass,
+  DatasetDocument,
+} from "src/datasets/schemas/dataset.schema";
 
 const IGNORE_FIELDS = ["updatedAt", "updatedBy", "_id"];
 
@@ -90,8 +93,9 @@ export function convertGenericHistoryToObsoleteHistory(
 // starting from the latest dataset, replay the history entries in reverse order to reconstruct the obsolete history entries
 export function convertGenericHistoriesToObsoleteHistories(
   histories: GenericHistory[],
-  currentDataset: Partial<DatasetClass>,
+  currentDataset: DatasetDocument,
 ): HistoryClass[] {
+  currentDataset = currentDataset.$clone();
   const result: HistoryClass[] = [];
   for (const history of histories) {
     const obsoleteHistory = convertGenericHistoryToObsoleteHistory(
@@ -99,7 +103,8 @@ export function convertGenericHistoriesToObsoleteHistories(
       currentDataset,
     );
     for (const key of Object.keys(history.before || {})) {
-      (currentDataset as Record<string, unknown>)[key] = history.before?.[key];
+      (currentDataset as unknown as Record<string, unknown>)[key] =
+        history.before?.[key];
     }
     result.push(obsoleteHistory);
   }
--- a/src/common/utils/history.util.ts
+++ b/src/common/utils/history.util.ts
@@ -73,7 +73,7 @@ export function convertGenericHistoryToObsoleteHistory(
     if (IGNORE_FIELDS.includes(field)) {
       continue;
     }
-    if (field === "datasetlifecycle") {
+    if (field === "datasetlifecycle" && datasetSnapshot.datasetlifecycle) {
       history.before[field] = {
         ...JSON.parse(JSON.stringify(datasetSnapshot.datasetlifecycle)),
         ...(history.before[field] as Record<string, unknown>),
--- a/src/common/utils/history.util.ts
+++ b/src/common/utils/history.util.ts
@@ -1,0 +1,107 @@
+import { HistoryClass } from "src/datasets/schemas/history.schema";
+import { GenericHistory } from "../schemas/generic-history.schema";
+import { DatasetClass } from "src/datasets/schemas/dataset.schema";
+
+const IGNORE_FIELDS = ["updatedAt", "updatedBy", "_id"];
+
+export function convertObsoleteHistoryToGenericHistory(
+  history: HistoryClass,
+  documentId: string,
+): GenericHistory {
+  const result: GenericHistory = {
+    subsystem: "Dataset",
+    documentId: documentId,
+    user: history.updatedBy,
+    operation: "update",
+    timestamp: history.updatedAt,
+    before: {},
+    after: {},
+  };
+  const changeList = Object.keys(history).filter(
+    (key) => !IGNORE_FIELDS.includes(key),
+  );
+  for (const key of changeList) {
+    if (
+      !history[key] ||
+      !history[key].hasOwnProperty("previousValue") ||
+      !history[key].hasOwnProperty("currentValue")
+    ) {
+      continue;
+    }
+    const fieldChange = history[key] as {
+      previousValue: unknown;
+      currentValue: unknown;
+    };
+    if (key === "datasetlifecycle") {
+      const currentValue = fieldChange.currentValue as Record<string, unknown>;
+      const previousValue = fieldChange.previousValue as Record<
+        string,
+        unknown
+      >;
+      // only retain the intersection of keys in currentValue and previousValue and whose value has changed. drop all others
+      const prunedPreviousValue: Record<string, unknown> = {};
+      const prunedCurrentValue: Record<string, unknown> = {};
+      for (const subKey of Object.keys(currentValue)) {
+        if (currentValue[subKey] !== previousValue[subKey]) {
+          prunedPreviousValue[subKey] = previousValue[subKey];
+          prunedCurrentValue[subKey] = currentValue[subKey];
+        }
+      }
+      fieldChange.previousValue = prunedPreviousValue;
+      fieldChange.currentValue = prunedCurrentValue;
+    }
+    result.before![key] = fieldChange.previousValue;
+    result.after![key] = fieldChange.currentValue;
+  }
+  return result;
+}
+
+// Given a dataset snapshot and a history entry, reconstruct the obsolete history entry
+export function convertGenericHistoryToObsoleteHistory(
+  history: GenericHistory,
+  datasetSnapshot: Partial<DatasetClass>,
+): HistoryClass {
+  const result: HistoryClass = {
+    updatedAt: history.timestamp,
+    updatedBy: history.user ?? "",
+    _id: "",
+  };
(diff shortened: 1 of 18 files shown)


src/datasets/schemas/history.schema.ts
```diff
@@ -39,6 +39,7 @@ export class HistoryClass {
    * After the migration, we should look in to change the structure
    * to a new one that can be better validated
    */
+  [key: string]: unknown;
```
minottic, line 42: what if the history class was a parametric type and could accept the class it's updating? Then this could be 

```TS
[key: keyof T]: {before: Partial<T>, after: Partial<T>}
```
or similar (and something else for v3)

src/datasets/utils/history.util.ts
```diff
@@ -0,0 +1,112 @@
+import { HistoryClass } from "src/datasets/schemas/history.schema";
+import { GenericHistory } from "../schemas/generic-history.schema";
+import {
+  DatasetClass,
+  DatasetDocument,
+} from "src/datasets/schemas/dataset.schema";
+
+const IGNORE_FIELDS = ["updatedAt", "updatedBy", "_id"];
+
+export function convertObsoleteHistoryToGenericHistory(
```
minottic: is this used anywhere?

src/datasets/utils/history.util.ts
```diff
@@ -0,0 +1,112 @@
...
+  currentDataset: DatasetDocument,
+): HistoryClass[] {
+  currentDataset = currentDataset.$clone();
+  const result: HistoryClass[] = [];
+  for (const history of histories) {
+    const obsoleteHistory = convertGenericHistoryToObsoleteHistory(
+      history,
+      currentDataset,
+    );
+    for (const key of Object.keys(history.before || {})) {
```
minottic, line 53: I am not sure I understand what this second for does. Isn't it already covered by line 85?

src/common/utils/history.util.ts
```diff
@@ -0,0 +1,112 @@
...
+  const result: HistoryClass = {
+    updatedAt: history.timestamp,
+    updatedBy: history.user ?? "",
+    _id: "",
+  };
+  for (const field in history.before) {
+    if (IGNORE_FIELDS.includes(field)) {
+      continue;
+    }
+    if (field === "datasetlifecycle") {
```
minottic: I think if this function is dataset specific, then maybe it should be in the dataset Utils, otherwise maybe one can make it generic (though I am unsure it's needed)

minottic (commented):
thanks! minor comments, LGTM otherwise. 

I think a class transformer would have been better suited by I see that we are already using these convertors throughout the code

src/common/utils/history.util.ts
minottic: and also, should it apply to other fields than datasetlifecycle?

src/datasets/schemas/history.schema.ts
omkar-ethz: This (`HistoryClass`) is the v3 class. It had an [embedsMany](https://github.com/SciCatProject/backend-v3/blob/master/common/models/dataset.json#L135) relation to Dataset in the old backend (was called `Message` then). It is specific to Dataset [i.e. only Dataset (specifically OutputDatasetObsoleteDto) uses this class in the current backend], so probably no benefit in making it generic.

I tried your suggestion for `T = DatasetClass`: 
```ts
type HistoryEntries = {
  [Property in keyof DatasetClass]: {
    previousValue: DatasetClass[Property];
    currentValue: DatasetClass[Property];
  };
};
export type HistoryExtended = HistoryClass & HistoryEntries;

// in client code
const { updatedAt, updatedBy, _id, ...changeList } = historyExtended;
// now we can expect changeList to be of type HistoryEntries
```
and it seems like a good idea, but there's no guarantee that the keys in the current `DatasetClass` have the same name and type as in the old backend's history. (e.g. we already know that in old history, dates were typecast to strings e.g. `datasetlifecycle.archiveRetentionTime`). Moreover, the destructring above doesn't really work, as fields like updatedBy and _id are shared between HistoryClass and DatasetClass.

As for the [GenericHistory](https://github.com/SciCatProject/scicat-backend-next/blob/6d279030872c2923925f02ec353f134fe6d3f2a4/src/common/schemas/generic-history.schema.ts#L14) (the current class which is supposed to be used for history for Datsets, Proposals, etc), it indeed might make sense  to have it as a parametric type. But that is more a design decision of the GenericHistory plugin and out of scope of this PR.

src/common/utils/history.util.ts
omkar-ethz: 1. In the previous backend, history was only tracked for Datasets. So there is no need for a generic function.
2. Good point, it's dataset specific, moved it from common/ to datasets/
3. As discussed, it might be cumbersome for scientific metadata,etc. So keeping it only for datasetlifecycle. Created an issue with "Won't fix": https://github.com/SciCatProject/scicat-backend-next/issues/2285

src/datasets/utils/history.util.ts
omkar-ethz: So this function "rewinds" the current dataset back in time by applying history.before.
$$currentDataset_T, currentDataset_{T-1},...$$

The function above this one, takes a dataset snapshot $currentDataset_t$ and `genericHistory` and calculates the old history's `field.previousValue` by merging  $currentDataset_t$.field and `genericHistory.previousValue`. It doesn't touch the dataset itself, this is why this loop is needed.

I guess there would be a more direct way to do this, but i found this two step process more intiutive.

src/datasets/utils/history.util.ts
omkar-ethz: This will be used in the `up` migration script: https://github.com/SciCatProject/scicat-backend-next/commit/afbb56d6219fc8d1b6a2e4164a3a16c9fabc2334, creating a PR soon

omkar-ethz: Migration script PR: https://github.com/SciCatProject/scicat-backend-next/pull/2286

src/datasets/utils/history.util.ts
minottic: ah that makes sense, thanks for the explanation!

src/datasets/utils/history.util.ts
minottic: could you please move it in the migration PR then? I am not sure I agree too much with the migration one, but we could merge this one in the meantime if this was part of the other PR

minottic (changes requested):

omkar-ethz:
@minottic sure, moved the obsolete-to-generic function in migration PR https://github.com/SciCatProject/scicat-backend-next/pull/2286

minottic (approved):
LGTM, thanks!

Issue #2247, opened 2025-10-01
Migration script for HistoryClass


sbliven:
## Summary

Datasets (and other schemas) may contain a `history` array if created with old versions of scicat. We should add a database migration script that migrates these history arrays to `HistoryClass` instances.

## Details

May relate to #584

omkar-ethz:
Slight correction, `HistoryClass` is the old history schema, the new schema is `GenericHistory` ([class](https://github.com/SciCatProject/scicat-backend-next/blob/a1d7def9668869319ebda23eff0009011a200f5d/src/common/schemas/generic-history.schema.ts#L14))

omkar-ethz:
https://github.com/SciCatProject/scicat-backend-next/pull/2286 and https://github.com/SciCatProject/scicat-backend-next/pull/2288 solve the issue

PR #2286, opened 2025-10-15
feat: add migration script to/from legacy history to generic history

omkar-ethz:
## Description

Depends on https://github.com/SciCatProject/scicat-backend-next/pull/2282

## Motivation

With the above PR, history for new records will be included in the datasets response. The migration script is needed to backfill the older history records as well.

## Fixes

* https://github.com/SciCatProject/scicat-backend-next/issues/2247

## Changes:

* Added up and down migration function for legacy history <-> generic history
* Added a util function to convert obsolete history to generic history, used in `up` migration
* Modified `migrate:db:[up|down]` to include `npm run build`, as the migration script depends on compiled dataset/utils module 
## Tests included

- [x] Passing? 

## Documentation

### Testing

Tested locally that up / down scripts preserve records across migration.


minottic:
is there a way to do this with an aggregation pipeline (or at least with a cursor)? I know it might duplicate logic, but I think to find and then loop is fairly inefficient

omkar-ethz:
> is there a way to do this with an aggregation pipeline (or at least with a cursor)? I know it might duplicate logic, but I think to find and then loop is fairly inefficient

`find` returns a [cursor](https://www.mongodb.com/docs/drivers/node/current/crud/query/cursor/#asynchronous-iteration). It would have been memory inefficient if I did a `toArray` on the cursor.
As for doing the entire migration with an aggregation pipeline, i think the logic is complex enough, (e.g. in the `up` case to compute the intersection of fields in datasetlifecycle.previous/current value that have changed; or in the `down` case to reconstruct the previousValue using datasetSnapshot), that it might be too complex to express this using mongodb operators.. do you think it's worth trying?
As a small optimization, i replaced the insertOne with insertMany in the `up` case.

minottic:
> find returns a [cursor](https://www.mongodb.com/docs/drivers/node/current/crud/query/cursor/#asynchronous-iteration)

ah, you are right, since this is not using mongoose but the mongo driver, my bad.

I gave a quick try to see how the aggregation pipeline would look, and it's indeed a bit complex and prone to typos. Happy to use this then, let's reconsider if necessary after we test the migration

migrations/20251014092142-legacy-history-to-generic-history.js
```diff
@@ -0,0 +1,63 @@
...
+    await db.collection("Dataset").updateMany({}, { $unset: { history: "" } });
+  },
+
+  /**
+   * @param db {import('mongodb').Db}
+   * @param client {import('mongodb').MongoClient}
+   * @returns {Promise<void>}
+   */
+  async down(db, client) {
+    for await (const dataset of db.collection("Dataset").find({})) {
```
minottic: should this be limited to objects with empty (or unset) history? It would speed up partial (unfinished) migrations which are restarted.

minottic (commented):
minor, comment, happy to approve otherwise

migrations/20251014092142-legacy-history-to-generic-history.js
omkar-ethz: Good idea! filtering on unset history. (as empty history i.e. `[]` still indicates a migrated document)

minottic (approved):

PR #2288, opened 2025-10-16
fix: convert all object fields from obsolete to generic history

omkar-ethz:
## Description

Fixes an inconsitency with https://github.com/SciCatProject/scicat-backend-next/pull/2286 - migrate all object-typed fields to the GenericHistory delta format, not just datasetlifecyle. Using `computeDeltaWithOriginals` of the history plugin to construct the deltas.

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/datasets/utils/history.util.ts
+++ b/src/datasets/utils/history.util.ts
@@ -4,6 +4,7 @@ import {
   DatasetClass,
   DatasetDocument,
 } from "src/datasets/schemas/dataset.schema";
+import { computeDeltaWithOriginals } from "src/common/utils/delta.util";
 
 const IGNORE_FIELDS = ["updatedAt", "updatedBy", "_id"];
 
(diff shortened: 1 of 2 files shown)


src/datasets/utils/history.util.ts
```diff
@@ -35,23 +36,19 @@ export function convertObsoleteHistoryToGenericHistory(
...
-          prunedPreviousValue[subKey] = previousValue[subKey];
-          prunedCurrentValue[subKey] = currentValue[subKey];
-        }
-      }
-      fieldChange.previousValue = prunedPreviousValue;
-      fieldChange.currentValue = prunedCurrentValue;
+      const { delta, originals } = computeDeltaWithOriginals(previousValue, {
+        ...previousValue,
+        ...currentValue,
+      });
```
omkar-ethz, line 49: The rationale behind this logic is:
in the old backend's history ([code](https://github.com/SciCatProject/backend-v3/blob/eb2bf8fe0d9d6ebb2f62d91391740658bb430048/common/models/utils.js#L203)), previousValue will contain datsetInstance[field], and currentValue will contain the fields to patch. computeDeltaWithOriginals expects both "complete" (oldObject, newObject). so we construct the newObject by overriding previousValue i.e. datsetInstance[field] with ...currentValue

minottic (approved):

PR #2304, opened 2025-10-21
feat(datasets v3): add validation of scientificMetadata with schema

sofyalaski:
## Description

Add the functionality implemented in Datasets v4 already. If the schema for scientific metadata is provided, it will validate the scientificMetadata object against it and put it under `scientificMetadataValid` property.

## Motivation

Since V3 version of datasets supports the obsolete classes and accepts instances of a union type, I couldn't use the pipe the way it was done in datasets V4 and had to put the function `validateScientificMetadataAgainstSchema` in the controller.

## Changes:

* adds optional properties `scientificMetadataSchema` and `scientificMetadataValid`  to the `UpdateDatasetObsoleteDto`. 

## Tests included

- [x] Passing? 

## Documentation


despadam (changes requested):
Looks way better, thanks @sofyalaski! 

My only comment is that there is a typo in this file's name: `src/datasets/utils/scintificMetadata.ts`

despadam (approved):

test/DatasetCustom.js
```diff
@@ -1,21 +1,23 @@
...
 
-let accessTokenAdminIngestor = null,
-  accessTokenArchiveManager = null,
-  accessTokenUser1 = null,
-  accessTokenUser2 = null,
-
-  pid = null,
-  minPid = null,
-  explicitPid = null;
+var accessTokenAdminIngestor = null;
```
minottic: I think the use of `var` is discouraged, in favour of `let` https://stackoverflow.com/questions/47465515/why-is-var-not-deprecated

src/datasets/datasets.controller.ts
```diff
@@ -1440,9 +1447,16 @@ export class DatasetsController {
         dtoType = PartialUpdateDatasetDto;
         break;
     }
+
+    const patchedDto = this.scientificMetadataValidator.patchedMetadata(
```
minottic: can this be merged with the method below? I think the whole logic of how the validationStatus flag is set should be part of the validator rather than the controller

src/datasets/pipes/scientific-metadata-validation.pipe.ts
minottic: this is a shallow copy, which I guess is fine in this case, but one should pay attention not to modify nested mutable fields. To stay on the safe side, one can use `structuredClone`

src/datasets/utils/scientificMetadata.ts
```diff
@@ -0,0 +1,121 @@
...
+      Logger.log(
+        error instanceof Error ? error.message : String(error),
+        "ScientificMetadataValidation",
+      );
+      return false;
+    }
+  }
+
+  async addValidationStatus(datasetDto: DatasetDto): Promise<ValidatedDto> {
+    const updatedDto = { ...datasetDto };
```
minottic: this is a shallow copy, which I guess is fine in this case, but one should pay attention not to modify nested mutable fields. To stay on the safe side, one can use `structuredClone`

minottic (commented):
I am not sure I understand why the Pipe could not be used with `UsePipes`. Even if that was not suitable, I wonder if something like this would have worked, e.g.: 

```ts
@Body(new ScientificMetadataValidationPipe())
```

or, if not:

```ts
new ScientificMetadataValidationPipe().transform(updateDatasetObsoleteDto)
```

if one needs to distinguish between patch and post with the same interceptor, would it work to remove this

```ts
 if (
      this.request.method === "PATCH" &&
      datasetDto instanceof PartialUpdateDatasetDto
    ) {
```

and try catch [this line](https://github.com/SciCatProject/scicat-backend-next/pull/2304/files#diff-70b0efb394999b362e9f8485047239413f5d85002adacf6a2cf1a01f73171fb0L39)?

sofyalaski:
> I am not sure I understand why the Pipe could not be used with `UsePipes`. 

It's because in datasets V3 the DTO is passed as a union type, so we would then need to infer the type in the pipe, which is already being done in the V3 controller. In the end, we would not duplicate the pipe, but the whole functionality to infer the type instead.

minottic:
> It's because in datasets V3 the DTO is passed as a union type, so we would then need to infer the type in the pipe, which is already being done in the V3 controller. In the end, we would not duplicate the pipe, but the whole functionality to infer the type instead.

what I don't understand is where the Pipe needs the dto type. Doesn't it only need to know if the payload or the exsiting ojbect has a schemaUrl field?

src/datasets/pipes/scientific-metadata-validation.pipe.ts
```diff
@@ -26,23 +45,26 @@ export class ScientificMetadataValidationPipe
   ) {}
 
   async transform(datasetDto: DatasetDto): Promise<ValidatedDto> {
-    const updatedDto = { ...datasetDto };
+    const updatedDto = JSON.parse(JSON.stringify(datasetDto));
```
minottic, line 48: I think using `structuredClone` is a bit more optimized to do copies

minottic (approved):
thanks! and sorry for the back and forth! very minor comment

despadam (approved):
Awesome, glad we were able to use the pipe now. Looks much cleaner.

src/datasets/pipes/scientific-metadata-validation.pipe.ts
sofyalaski: no worries, I just didn't quite get your first message, sorry for that.
As for the structured clone, it indeed does a better job at deep cloning the dto, such that the properties are immutable because they were defined as readonly in the dto, and then I can't change the validation status downstream

PR #2315, opened 2025-10-29
refactor: change the naming of the config flag to hideEmptyMetadataTable

abdimo101:
## Description
This PR updates the name of the config flag showEmptyMetadataTable to hideEmptyMetadataTable.

## Motivation
When set to `false`, metadata sections are displayed even when empty (default behavior). Set to `true` to hide empty sections.
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/config/frontend.config.json
+++ b/src/config/frontend.config.json
@@ -63,7 +63,7 @@
   "datasetDetailsShowMissingProposalId": false,
   "notificationInterceptorEnabled": true,
   "metadataEditingUnitListDisabled": true,
-  "showEmptyMetadataTable": false,
+  "hideEmptyMetadataTable": false,
   "datafilesActionsEnabled": true,
   "datafilesActions": [
     {


src/config/frontend.config.json
```diff
@@ -63,7 +63,7 @@
   "datasetDetailsShowMissingProposalId": false,
   "notificationInterceptorEnabled": true,
   "metadataEditingUnitListDisabled": true,
-  "showEmptyMetadataTable": false,
+  "hideEmptyMetadataTable": false,
```
minottic, line 66: shouldn't this set to true to be backwards compatible?

minottic (approved):
LGTM, thanks! Minor comment

src/config/frontend.config.json
Junjiequan: In this case `false` should be used for backward compatibility. It keeps the previous behavior, whether this flag is provided or not.

Junjiequan (approved):

src/config/frontend.config.json
minottic: oh ok. My comment was based on the previous value being `"showEmptyMetadataTable": false,`, which would equal `"hideEmptyMetadataTable": true,`...

PR #2318, opened 2025-10-30
fix: use custom email list validation rule for contactEmail

### LINKED ISSUE #2317: Incorrect validation rule for `contactEmail`
## Summary
Validation rules for `contactEmail` are incorrect: The spec is:
"Email of the contact person for this dataset. The string may contain a list of emails, which should then be separated by semicolons."
whereas the validation rule `@IsEmail` enforces a single email,

## Steps to Reproduce

POST /v3/datasets/isValid:
```
{
    "classification": "IN=medium,AV=low,CO=low",
    "contactEmail": "omkar.zade@psi.ch; scicatingestor@psi.ch",
    "creationTime": "2025-05-28T11:37:29.778706+02:00",
    "datasetName": "test-omkar-11",
    "description": "this dataset has a folder inside",
    "isPublished": false,
    "license": "CC BY-SA 4.0",
    "owner": "ingestor",
    "ownerEmail": "scicatingestor@psi.ch",
    "ownerGroup": "svc-afs",
    "sourceFolder": "/home/zade_o/Downloads/test-dataset-11",
    "type": "base"
}
```

## Current Behaviour

```
{
  "valid": false
}
```

## Expected Behaviour

```
{
  "valid": true
}
```

omkar-ethz:
## Description

Fixes https://github.com/SciCatProject/scicat-backend-next/issues/2317

## Motivation

Users at PSI used to be able to ingest datasets with multiple contactEmails as per the spec of the field, but weren't anymore due to this bug.

## Fixes

https://github.com/SciCatProject/scicat-backend-next/issues/2317

## Changes:

* Added a custom email list class-validator ([reference](https://github.com/typestack/class-validator?tab=readme-ov-file#custom-validation-classes))
* Using it on the `contactEmail` field in updateDataset[obsolete]dto's which are the base classes for dataset creation dto's. 
## Tests included

- [x] Passing? 

Tested locally using scicatlive that the payload mentioned in the issue now returns `{valid: true}`

## Documentation

### DIFF
--- a/src/common/interceptors/mask-sensitive-data.interceptor.ts
+++ b/src/common/interceptors/mask-sensitive-data.interceptor.ts
@@ -7,7 +7,7 @@ import {
 } from "@nestjs/common";
 import { ConfigService } from "@nestjs/config";
 import { APP_INTERCEPTOR } from "@nestjs/core";
-import { isEmail, ValidationArguments } from "class-validator";
+import { isEmail } from "class-validator";
 import { from, map, mergeMap, Observable } from "rxjs";
 import { JWTUser } from "src/auth/interfaces/jwt-user.interface";
 import { AccessGroupsType } from "src/config/configuration";
@@ -22,6 +22,7 @@ class MaskSensitiveDataInterceptor implements NestInterceptor {
   constructor(
     private readonly configService: ConfigService,
     private userIdentitiesService: UserIdentitiesService,
+    private customEmailListValidator: CustomEmailList,
   ) {
     this.adminGroups =
       this.configService.get<AccessGroupsType>("accessGroups")?.admin;
@@ -72,12 +73,11 @@ class MaskSensitiveDataInterceptor implements NestInterceptor {
   }
 
   private isToMaskEmailList(value: string | unknown, ownEmails: Set<string>) {
-    const customEmailListValidator = new CustomEmailList();
     return (
       typeof value === "string" &&
-      customEmailListValidator.validate(value, {} as ValidationArguments) &&
-      !customEmailListValidator
-        .extractEmails(value)
+      this.customEmailListValidator.validate(value) &&
+      !this.customEmailListValidator
+        .splitEmails(value)
         .every((e) => ownEmails.has(e))
     );
   }
@@ -87,11 +87,11 @@ class MaskSensitiveDataInterceptor implements NestInterceptor {
   }
 
   private maskListValue(value: string, ownEmails: Set<string>): string {
-    const customEmailListValidator = new CustomEmailList();
-    return customEmailListValidator
-      .extractEmails(value)
-      .map((e) => (ownEmails.has(e) ? e : this.maskValue()))
-      .join("; ");
+    return this.customEmailListValidator.joinEmails(
+      this.customEmailListValidator
+        .splitEmails(value)
+        .map((e) => (ownEmails.has(e) ? e : this.maskValue())),
+    );
   }
 
   private async getIdentityEmails(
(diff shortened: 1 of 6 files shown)


minottic (commented):
thanks! LGTM. Could you please check the mask-sensitive-data interceptor as well? I think that one also relies on single emails

omkar-ethz:
> thanks! LGTM. Could you please check the mask-sensitive-data interceptor as well? I think that one also relies on single emails

thank you! updated the logic in the interceptor - so if the contactEmails are `accessible@domain.com; notaccessible@domain.com` the result will be `accessible@domain.com; *****`

src/common/interceptors/mask-sensitive-data.interceptor.ts
```diff
@@ -64,10 +71,29 @@ class MaskSensitiveDataInterceptor implements NestInterceptor {
     return typeof value === "string" && isEmail(value) && !ownEmails.has(value);
   }
 
+  private isToMaskEmailList(value: string | unknown, ownEmails: Set<string>) {
+    const customEmailListValidator = new CustomEmailList();
```
minottic: I think one can instantiate the class as a class property, so we avoid multiple instantantiation (even if cheap)

src/datasets/utils/email-list-validator.util.ts
```diff
@@ -0,0 +1,24 @@
...
+  isEmail,
+} from "class-validator";
+
+@ValidatorConstraint({ name: "customEmailList", async: false })
+export class CustomEmailList implements ValidatorConstraintInterface {
+  extractEmails(text: string): string[] {
+    return text.split(";").map((e) => e.trim());
+  }
+
+  validate(text: string, _: ValidationArguments) {
```
minottic: can the ValidationArguments be optional?

minottic: or maybe the validationArguments could be a set of allowed emails? then checking it in the mask becomes trivial

src/common/interceptors/mask-sensitive-data.interceptor.ts
```diff
@@ -64,10 +71,29 @@ class MaskSensitiveDataInterceptor implements NestInterceptor {
...
   private maskValue(): string {
     return "*****";
   }
 
+  private maskListValue(value: string, ownEmails: Set<string>): string {
+    const customEmailListValidator = new CustomEmailList();
+    return customEmailListValidator
+      .extractEmails(value)
+      .map((e) => (ownEmails.has(e) ? e : this.maskValue()))
+      .join("; ");
```
minottic: I feel the logic of the join should be consistent with the extract. What if we had an extractEmail (or splitEmail) in the validator as is and a joinEmails that takes a list of emails and does the join? In this case if we change the separator in the validator we recombine consistently here

src/common/interceptors/mask-sensitive-data.interceptor.ts
minottic: (or maybe the new is redundant anyway)

minottic (approved):
some minor comments

src/common/interceptors/mask-sensitive-data.interceptor.ts
omkar-ethz: great point! made this change

src/datasets/utils/email-list-validator.util.ts
omkar-ethz: had a look at the ValidatorConstraintInterface that this class implements, and indeed validationArgs was optional - removed it from the implementation as we're not using it, thanks!

or maybe the validationArguments could be a set of allowed emails? then checking it in the mask becomes trivial
--> this would be nice in the mask, but when used as a decorator, the `validationArgs` are passed by the framework i.e. class-validator.. so i think there might be some conflict if we overload its definition.

src/common/interceptors/mask-sensitive-data.interceptor.ts
omkar-ethz: Providing it via DI, in case CustomEmailList needs to participate in DI in future (e.g. making the SEPARATOR part of app-config), thanks!

PR #2321, opened 2025-10-30
fix: accept v3 fields in filters when using v3 jobs

minottic:
## Description
This bug can be tried in the JOB view in the frontend, where the sorting has no effect. The cause is that the FE sends the fields of V3 to sort, but these have been renamed in the DB so the sorting has no effect. This PR introduces some pipelines that have the effect of mapping the v3 fields input by the client to the DB ones, at all levels, in fields, whereConditions and limits. IMO this logic should take place in all v3 endpoints. If so, it can be implemented in later PRs

It's opened agains #2320 as it reuses the jobv3v4 mapping. I will open against master once merged

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/jobs/pipes/v3-filter.pipe.ts
+++ b/src/jobs/pipes/v3-filter.pipe.ts
@@ -1,0 +1,182 @@
+import { PipeTransform, Injectable, ArgumentMetadata } from "@nestjs/common";
+import { jobV3toV4FieldMap } from "../types/jobs-filter-content";
+import _ from "lodash";
+
+type KeyMap = Record<string, string>;
+
+type Func = (value: unknown) => unknown;
+
+type FuncMap = Record<string, Func>;
+
+interface TransformDeepOptions {
+  keyMap?: KeyMap;
+  funcMap?: FuncMap;
+  arrayFn?: Func;
+  valueFn?: Func;
+}
+
+const transformDeep = (
+  obj: unknown,
+  opts: TransformDeepOptions = {},
+): unknown => {
+  const { keyMap = {}, funcMap = {}, arrayFn, valueFn } = opts;
+
+  if (Array.isArray(obj)) {
+    return obj.map((item) =>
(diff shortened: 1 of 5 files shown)


src/jobs/pipes/v3-filter.pipe.ts
```diff
@@ -0,0 +1,182 @@
...
+        const isArray = _.isArray(value);
+        const order = (isArray ? value : [value]).reduce((acc, orderValue) => {
+          const [field, direction] = (orderValue as string).split(":");
+          return acc.concat(`${keyMappings[field]}:${direction}`);
+        }, [] as string[]);
+        return isArray ? order : order[0];
+      },
+    });
+    super(
+      [sortToOrderPipe, new V3ConditionToV4Pipe(keyMappings, false)],
```
omkar-ethz: I think the `V3ConditionToV4Pipe` isn't needed here, right?
But in case it is, it will be included twice in `V3FilterToV4Pipe`, which might not be intended

src/jobs/pipes/v3-filter.pipe.ts
```diff
@@ -0,0 +1,182 @@
...
+      newObj[mappedKey] = valueFn ? valueFn(transformed) : transformed;
+    }
+    return newObj;
+  }
+
+  return obj;
+};
+
+class ParseJsonPipe implements PipeTransform<string, string> {
+  transform(value: string): string {
```
omkar-ethz, line 49: should be `PipeTransform<string, string | object>`, same for `transform`

omkar-ethz (approved):
LGTM! minor comment/question

src/jobs/pipes/v3-filter.pipe.ts
minottic: it's needed there since the filter keys need to be replaced with the correct v3 fields after mapping. I have removed it from V3FilterToV4Pipe. I will add some comments. Thanks!

minottic: you are right, thanks! Fixed

PR #2331, opened 2025-11-07
fix: implementing optimistic concurrency control

### LINKED ISSUE #1969: Support for conditional requests/Strategy for not missing updates?

FloKo2669:
## Description
This is a direct followup PR for #2138, addressing the change requests. Mainly introducing a helper function to reduce the code duplication in each controller function. The following Endpoints are modified:

- Sample
- Attachments
- Datasets
- datasets.v4
- instruments
- origdatablocks
- origdatablocks.v4
- proposals

The tests were left as they were in PR #2138, except for minor adjustments. 

## Changes compared to original PR:

- introduced helper function to reduce code duplication
- removed the check in pusblished dataset
- uniform header access using request.headers
- removed findByIdAndUpdateInternal in publish datasets

### DIFF
--- a/migrations/20251120153452-add-updatedby-to-instrument.js
+++ b/migrations/20251120153452-add-updatedby-to-instrument.js
@@ -1,0 +1,14 @@
+module.exports = {
+  async up(db, client) {
+    await db
+      .collection("Instrument")
+      .updateMany(
+        { updatedBy: { $exists: false } },
+        [{ $set: { updatedBy: "$createdBy" } }]
+      );
+  },
+
+  async down(db, client) {
+    // No path backward as it's not easy to find on down where updatedBy was added
+  },
+};
--- a/src/attachments/attachments.v4.controller.spec.ts
+++ b/src/attachments/attachments.v4.controller.spec.ts
@@ -72,7 +72,6 @@ describe("AttachmentsController - findOneAndUpdate", () => {
     const result = await controller.findOneAndUpdate(
       { headers },
       "123",
-      headers,
       dto,
     );
 
@@ -84,7 +83,7 @@ describe("AttachmentsController - findOneAndUpdate", () => {
     const dto = { description: null };
     const headers = { "content-type": "application/merge-patch+json" };
 
-    await controller.findOneAndUpdate({ headers }, "123", headers, dto);
+    await controller.findOneAndUpdate({ headers }, "123", dto);
 
     const expectedPatched = jmp.apply(mockAttachment, dto);
     expect(service.findOneAndUpdate).toHaveBeenCalledWith(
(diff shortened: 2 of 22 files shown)


nitrosx (commented):
Do we need a migration script for the db given that instruments are now inheriting from ownable?
Other than that, it is good to go for me.

FloKo2669:
@nitrosx Valid point. However, I am not entirely sure that a migration script is needed, because the two newly added required fields are createdBy and updatedBy. createdBy is already set for Instrument (at least in the test data of SciCatLive). Both fields are set automatically during creation/update in the service, and to my understanding, they are only validated during these operations. However, I can see how it could lead to problems or unexpected behavior when updatedBy is not set.

I also found the migration script that sets updatedBy for other types (Attachment, Dataset, Proposal, etc.). So I added a migration script with the same logic for Instrument to be safe.
However, I am not an expert in MongoDB migrations, so I am happy for anyone to check if it works.

nitrosx (approved):

bpedersen2 (approved):

PR #2341, opened 2025-11-11
fix: readd fields as obj for v3 for backward compatibility

minottic:
## Description
Before this #2237, the v3 controller accepted fields in the format {field1: 1, field2: 0} and not only [field1, field2]. This PR makes it support both formats. 
 
## Fixes
Add support for both formats. It should have been a Pipe, but given historic code that merges headers and qwargs, the pipe implementation would have required quite some refactoring

## Changes:

* changes made

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/test/RawDataset.js
+++ b/test/RawDataset.js
@@ -393,6 +393,37 @@ describe("1900: RawDataset: Raw Datasets", () => {
       });
   });
 
+  it("0133 should fetch datasets with include and scope and fields as obj", async () => {
+    const filter = {
+      where: { pid: decodeURIComponent(pid) },
+      fields: { pid: 1, datasetName: true, description: 0 },
+      include: [
+        { relation: "instruments" },
+        { relation: "proposals", scope: { fields: ["abstract"] } }
+      ],
+    };
+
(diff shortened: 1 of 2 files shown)


fpotier (approved):

test/RawDataset.js
```diff
@@ -393,6 +393,37 @@ describe("1900: RawDataset: Raw Datasets", () => {
       });
   });
 
+  it("0133 should fetch datasets with include and scope and fields as obj", async () => {
+    const filter = {
+      where: { pid: decodeURIComponent(pid) },
+      fields: { pid: 1, datasetName: true, description: 0 },
+      include: [
+        { relation: "instruments" },
+        { relation: "proposals", scope: { fields: ["abstract"] } }
```
omkar-ethz, line 402: This nested `fields` will not be transformed if given in object syntax, is this intended?

minottic: yes, that was not transformed even before the refactor. I might add a new PR later that introduces that, but I would need some consistency between filters and pipes

PR #2343, opened 2025-11-11
refactor: use custom filter decorator to simplify controller

minottic:
## Description
Refactor: use nestjs custom decorator to reduce duplication and keep controller simpler. See [here](https://docs.nestjs.com/custom-decorators)

## Fixes
This PR includes a small bug fix in findOne. Happy to split into 2 PRs if better

* findOne bug fix

### DIFF
--- a/src/datasets/decorators/filter.decorator.ts
+++ b/src/datasets/decorators/filter.decorator.ts
@@ -1,0 +1,31 @@
+import {
+  BadRequestException,
+  createParamDecorator,
+  ExecutionContext,
+  HttpException,
+  HttpStatus,
+} from "@nestjs/common";
+
+export const Filter = createParamDecorator(
+  (data: unknown, ctx: ExecutionContext) => {
+    const request = ctx.switchToHttp().getRequest();
+    if (request?.headers?.filter && request?.query?.filter)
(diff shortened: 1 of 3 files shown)


src/datasets/decorators/filter.decorator.ts
```diff
@@ -0,0 +1,31 @@
...
+      );
+    let filter = {};
+    try {
+      if (request?.query?.filter) filter = JSON.parse(request.query.filter);
+      if (request?.headers?.filter) filter = JSON.parse(request.headers.filter);
+    } catch (err) {
+      const error = err as Error;
+      throw new BadRequestException(`Invalid JSON in filter: ${error.message}`);
+    }
+    return { filter: JSON.stringify(filter) };
```
fpotier, line 29: why not return the parsed version here as every call site calls `JSON.parse()`?

fpotier (approved):

src/datasets/decorators/filter.decorator.ts
minottic: thanks! Yes, I also thought of that, but then I had to change all the types and how it was consumed later. I will do in another PR

PR #2344, opened 2025-11-12
refactor: unify replaceLike and fitler pipe

minottic:
## Description
Refactors FilterPipe to avoid regex replace and unifies with replaceLikeOperator. It applies replacements recursively

## Motivation
Single function that does all the mongo syntax replacement

## Changes:

* all mongo syntax replacements in one Pipe
* avoids regexes
* applies replacements recursively, as long as they are relatives of `where`
* can be applied on both strings and {filter: string}

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/common/pipes/filter.pipe.ts
+++ b/src/common/pipes/filter.pipe.ts
@@ -1,5 +1,4 @@
 import { Injectable, PipeTransform } from "@nestjs/common";
-import { isObject } from "lodash";
 import {
   transformDeep,
   TransformObjValuesPipe,
@@ -18,10 +17,7 @@ import {
 @Injectable()
 export class FilterPipe
   implements
-    PipeTransform<
-      { filter?: string; fields?: string } | string,
-      { filter?: string; fields?: string } | string
-    >
+    PipeTransform<{ filter?: string } | string, { filter?: string } | string>
 {
   private static readonly replaceOperatorsMap = {
     inq: "$in",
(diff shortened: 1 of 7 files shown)


src/common/pipes/filter.pipe.ts
```diff
@@ -1,38 +1,75 @@
...
+ * - Parses JSON strings if needed and returns the result in the same format (string or object).
+ */
 @Injectable()
 export class FilterPipe
   implements
     PipeTransform<
-      { filter?: string; fields?: string },
-      { filter?: string; fields?: string }
+      { filter?: string; fields?: string } | string,
+      { filter?: string; fields?: string } | string
```
omkar-ethz: As the {fields?: string} portion of the interface is never used in this class, it might be simplified everywhere to `{ filter?: string } | string`

src/common/pipes/filter.pipe.ts
```diff
@@ -1,38 +1,75 @@
...
 
-      filter = filter.replace(/{"inq":/g, '{"$in":');
-      // nin => $nin
+  constructor() {
+    this.replaceOperatorsPipe = new TransformObjValuesPipe({
+      where: (value: unknown) => {
+        return transformDeep(value, {
+          funcMap: {
+            ilike: (val: unknown, par: unknown) => {
+              if (!isObject(par)) return par;
```
omkar-ethz: not sure when this branch will be reached (i think par is guaranteed to be an object as it is initalized as `newObj = {}` in transformDeep. but if it is reached, should this be `return val?`

omkar-ethz (approved):
LGTM!

src/common/pipes/filter.pipe.ts
minottic: thanks, changed!

src/common/pipes/filter.pipe.ts
minottic: thanks!

PR #2345, opened 2025-11-13
refactor: return json from FilterPipe

minottic:
## Description
Return JSON from filterPipe to avoid running JSON.parse from all controllers. This is also needed by #2346 to coope with date valued filters


fpotier (approved):

Junjiequan (approved):
Looks good refactoring to me

PR #2347, opened 2025-11-13
fix: exclude history from v3 when not in fields

minottic:
## Description
History is included in all requests, even when excluded from the fields

## Fixes

* avoid including history when not in fields

## Tests included

- [x] Passing? 

## Documentation


omkar-ethz (approved):
Thanks, LGTM!

PR #2355, opened 2025-11-17
feat(POC): add run time config management service

Junjiequan:
## Description
This PR introduces a runtime configuration system.

Introduced runtime-config module:
- Controller
- Service
- DTO
- MongoDB schema

Added new API endpoints:
- GET /api/v3/runtime-config/:cid -> Public
- PUT /api/v3/runtime-config/:cid -> Only ADMIN users can access

Supported configuration IDs:
- Defaults:
  - frontendConfig
  - frontendTheme
- Can be extended using:
  - `CONFIG_SYNC_TO_DB_LIST ENV` variable

On application startup:
- backend reads CONFIG_SYNC_TO_DB_LIST
- For each cid in the list:
  - If record does not exist -> create from listed config file or from defaults
  - If record exists -> overwrite using config file 
  - iI cid corresponding config file do not exists -> throw log error
image

📌📌 **Config files are always the source of truth** 📌📌

Minor admin/config endpoint workflow changes:
- `admin/config` now internally call `runtime-config/frontendConfig`
- `admin/theme` now internally call `runtime-config/frontendTheme`

Both endpoints still exist.
Only the internal flow changed, with no breaking API changes.

## Motivation
This feature adds a DB-backed config store so the frontend config can be edited at runtime without rebuilding. The backend still treats the JSON files as the source of truth and auto-loads them into the Storage collection on startup (POC scope is only frontend config, even though the backend can load any config file). This makes it easier to test UI changes, share configs across environments, and edit them easily through JSONForms. 

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation


sbliven:
@Junjiequan Can you add an example for how to use this?

bpedersen2 (commented):
Sound good in general. 

I guess a filter for certain fields will be needed in the end if we allow be-config changes, e.g. to avoid loss of all admin groups or changes to mongodb settings( I would probably keep all security-related fields as  file-only).

Frontend-config should be less critical.

src/config/runtime-config/runtime-config.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+import { AppAbility } from "src/casl/casl-ability.factory";
+import { CheckPolicies } from "src/casl/decorators/check-policies.decorator";
+import { RuntimeConfig } from "./schemas/runtime-config.schema";
+@ApiBearerAuth()
+@ApiTags("runtime-config")
+@Controller("runtime-config")
+export class RuntimeConfigController {
+  constructor(private readonly runtimeConfigService: RuntimeConfigService) {}
+
+  @AllowAny()
```
sbliven, line 38: Needs the swagger descriptors (ApiOperation, ApiParam, etc)

src/config/runtime-config/dto/runtime-config.dto.ts
```diff
@@ -0,0 +1,36 @@
...
+import { IsString, IsOptional, IsObject } from "class-validator";
+
+export class OutputRuntimeConfigDto {
+  @ApiProperty({
+    type: String,
+    description: "Unique config identifier (e.g. 'frontend', 'backend', etc.)",
+    example: "frontend",
+  })
+  @IsString()
+  _id: string;
```
sbliven: I suggest `key` for the unique identifier. `_id` should only be used for the internal mongo document id (which should not be accessible in the API).

sbliven: Does it make sense to have more than "frontend" and "backend"? I would make this an enum. Or are you thinking there might be use cases where other microservices store most of their configs on the scicat backend?

src/config/runtime-config/dto/runtime-config.dto.ts
```diff
@@ -0,0 +1,36 @@
...
+  data: Record<string, unknown>;
+
+  @ApiProperty({
+    type: String,
+    required: false,
+    description: "Optional description of this configuration entry",
+  })
+  @IsOptional()
+  @IsString()
+  description?: string;
```
sbliven: What's the use of this? Just internal documentation?

src/config/runtime-config/utils.ts
```diff
@@ -0,0 +1,42 @@
+export const isPlainObject = (
```
sbliven: Use `common.utils.IsRecord()`

src/config/runtime-config/runtime-config.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+  @Get("data/:id")
```
sbliven: `data` strikes me as a surprising choice for the endpoint path. I see a close link between this functionality and the /admin/config endpoint. Maybe it could just be `GET/PUT/PATCH /admin/config/:id`? And there should be a way to get either the merged configuration or just the overridden values.

sbliven (dismissed):
I think we should discuss the core premise of this POC, which is that the configuration should be stored in the database instead of in files. While it would be convenient to be able to change the configuration on the fly, I'm worried that it will become even harder to track configuration changes. If we do it, I think there should be a clear process to "burn in" the configuration: export non-default settings to an env file, empty the database, and restart the backend with a clean configuration. Otherwise it's hard to keep track of changes.

Can you add a description of the usage to the README or docs? Even if it changes later, it would be helpful to read how the merging works. I think you said that the database is considered canon and overrides anything in the config files or the defaults.

How does this fit with `/admin/config`? Would this provide the merged "frontend" data? 

Likewise, will calls to the configService be merged to the "backend" data? If so, we should double-check that there aren't any remaining direct calls to `configuration()`, as those would bypass the configService.

I guess you would add more endpoints
- get the full, merged configuration
- get/set individual properties (or would this just be done with PATCH application/json+merge?)
- way to delete configuration changes (resetting from the files or defaults)

src/config/runtime-config/dto/runtime-config.dto.ts
Junjiequan: > _id should only be used for the internal mongo document id (which should not be accessible in the API).

I'm not sure about this statement. 
`await this.datablocksService.findOne({
      where: { _id: id },
    });` we have been implementing query like this in many places, I don't see what's the cons here..

> Does it make sense to have more than "frontend" and "backend"? I would make this an enum. Or are you thinking there might be use cases where other microservices store most of their configs on the scicat backend?

We don’t want to lock it to just “frontend” or “backend”.
The idea is to let operators store any JSON config they want, and the backend shouldnt restrict it. They decide what config files they want to use, not us. It could be `frontend-config`, `published-data-schema.json`, `proposal-ui-config` etc

Junjiequan:
> I think we should discuss the core premise of this POC, which is that the configuration should be stored in the database instead of in files. While it would be convenient to be able to change the configuration on the fly, I'm worried that it will become even harder to track configuration changes. If we do it, I think there should be a clear process to "burn in" the configuration: export non-default settings to an env file, empty the database, and restart the backend with a clean configuration. Otherwise it's hard to keep track of changes.

We defintelly needs to discuss more about this, I just wanted to pull the idea into code, so we can have something to disccuss upon. 

If i understand your proposal correctly, you are suggesting that:

1. the backend always starts with a empty config collection
2. it then loads the config files from disk (mounted or default)
3. creates a fresh config record based on provided files

So, the config file is the source of truth, config record is temporary copy for real time editing. once the backend restarts the temporary state resets and rebuilt from the file again.

Is it correct intepretation?



> Can you add a description of the usage to the README or docs? Even if it changes later, it would be helpful to read how the merging works. I think you said that the database is considered canon and overrides anything in the config files or the defaults.

Sure, I will provide it as soon as possible.

> How does this fit with /admin/config? Would this provide the merged "frontend" data?

It's up for discussion that wheter if we still want to keep /admin/config after we have the runtimeConfig endpoint.

> Likewise, will calls to the configService be merged to the "backend" data? If so, we should double-check that there aren't any remaining direct calls to configuration(), as those would bypass the configService.

The backend data ( environment variables I assume ) is completly seperated logic and should not be relevant to the run time config service at all, they are out of scope this time at least. The run time service serves for frontend related configs only.

> I guess you would add more endpoints
> 
> get the full, merged configuration
> get/set individual properties (or would this just be done with PATCH application/json+merge?)
> way to delete configuration changes (resetting from the files or defaults)

I imagine the entry sturcture will look like this:
```json
[
  {
    "_id": "frontendConfig",
    "description": "Some text to be rendered on the UI",
    "data": {
      "...": "raw JSON blob"
    }
  },
  {
    "_id": "frontendTheme",
    "data": {
      "...": "raw JSON blob"
    }
  },
  {
    "_id": "someOtherConfig",
    "data": {
      "...": "raw JSON blob"
    }
  }
]
```
On the frontend we use `jsonforms`, which always submits the full JSON object. So, I’m not sure individual property endpoints add much value right now. JSONForms doesn’t send partial updates, so we’d essentially be duplicating logic.
For now, setting/getting the whole config blob seems enough, unless we decide to go with other json editors.

src/config/runtime-config/dto/runtime-config.dto.ts
nitrosx: That's the idea.
If a site admin uses this subsystem to store additional configurations, we should also have a name and a description for management purposes.

src/config/runtime-config/dto/runtime-config.dto.ts
nitrosx: I would suggest just `id` as we ill store the whole configuration in one document... or at least, that was the idea last time we brainstormed.

src/config/runtime-config/dto/runtime-config.dto.ts
nitrosx, line 1: I would also the fields `createdBy`, `updatedAt`, `createdAt`.

src/config/runtime-config/runtime-config.controller.ts
nitrosx: I do not remember why we decided to move it away from `admin/config`.
If I remember correctly, the reason for it is that this subsystem can become a store for additional blobs besides configurations. In our mind `admin/config` relates only to admin users, which is true for updating but not for retrieving.
Hopefully it makes sense, but I will be happy to discuss further

src/config/runtime-config/runtime-config.module.ts
nitrosx, line 1: The naming is perfect if we are going to use this subsystem to store only configurations.

src/config/runtime-config/runtime-config.service.ts
nitrosx, line 1: Same as my previous comment

src/config/runtime-config/utils.ts
nitrosx: Agreed

nitrosx (changes requested):

nitrosx:
@sbliven valid feedback!!!
Let me and Jay brainstorm once or twice more and get back to you.
We can discuss afterwards.

Junjiequan:
@sbliven @nitrosx 
Most of the comments are addressed, documentation included  along with updated PR description.
The description field is removed from the runtimeConfig record as it is not going to be used now, but we can add it back when we need it.

src/config/runtime-config/runtime-config.service.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  async updateConfig(
+    cid: string,
+    config: Record<string, unknown>,
+    updatedBy: string,
+  ): Promise<OutputRuntimeConfigDto | null> {
+    const updatedDoc = await this.runtimeConfigModel.findOneAndUpdate(
+      { cid: cid },
+      { $set: { data: config, updatedBy } },
```
nitrosx: Shouldn't it be:
```
{ $set: { data: config, updatedBy: updatedBy } },
```

src/config/runtime-config/schemas/runtime-config.schema.ts
```diff
@@ -0,0 +1,43 @@
...
+  updatedBy: string;
+
+  @ApiProperty({
+    type: String,
+    required: false,
+    description: "User that created the configuration",
+  })
+  @Prop({ type: String, required: true, default: "system", immutable: true })
+  createdBy: string;
+}
```
nitrosx: Can we leverage QueryableClass or OwnableClass?

src/config/runtime-config/schemas/runtime-config.schema.ts
```diff
@@ -0,0 +1,43 @@
...
+import { ApiProperty } from "@nestjs/swagger";
+import { Document } from "mongoose";
+
+export type RuntimeConfigDocument = RuntimeConfig & Document;
+
+@Schema({
+  collection: "RuntimeConfig",
+  timestamps: true,
+})
+export class RuntimeConfig {
```
nitrosx: Can we extend QueryableClass or OwnableClass?

src/config/runtime-config/runtime-config.service.ts
nitrosx: Can we use the same functionalities that we have in Datasets, where we create and update fields are populated and updated automatically?

nitrosx (changes requested):

src/config/runtime-config/runtime-config.service.ts
Junjiequan: There's no difference between {data: config, updatedBy: updatedBy} and  {data: config, updatedBy}

src/config/runtime-config/schemas/runtime-config.schema.ts
Junjiequan: extended QueryableClass. OwnableClass is not needed yet for runtimeConfig service

src/config/runtime-config/runtime-config.service.ts
Junjiequan: `addUpdatedByField` and `addCreatedByField` is included

src/config/runtime-config/schemas/runtime-config.schema.ts
Junjiequan: extended QueryableClass. OwnableClass is not needed yet for runtimeConfig service

src/config/runtime-config/runtime-config.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+import { OutputRuntimeConfigDto } from "./dto/output-runtime-config.dto";
+import { RuntimeConfigService } from "./runtime-config.service";
+import { PoliciesGuard } from "src/casl/guards/policies.guard";
+import { Action } from "src/casl/action.enum";
+import { AppAbility } from "src/casl/casl-ability.factory";
+import { CheckPolicies } from "src/casl/decorators/check-policies.decorator";
+import { RuntimeConfig } from "./schemas/runtime-config.schema";
+import { UpdateRuntimeConfigDto } from "./dto/update-runtime-config.dto";
+@ApiBearerAuth()
+@ApiTags("runtime-config")
```
nitrosx: I would change it to `runtime configurations` to maintain consistency with the other subsystems.
Check dataset: https://github.com/SciCatProject/scicat-backend-next/blob/dff0e4dcc58e3e6a67e4295b16f69e7fa39c7d65/src/datasets/datasets-public.v4.controller.ts#L39-L45

nitrosx (changes requested):
Minor changes to the swagger controller tags.

Food for thoughts: we should add authorization to the get configuration endpoint, so admins can restrict access to some entries. The authorization is not for the endpoint is for each individual entry.
...but this is for the future

src/config/runtime-config/runtime-config.controller.ts
Junjiequan: fixed in https://github.com/SciCatProject/scicat-backend-next/pull/2355/changes/d46b3259e65c56cfdab165b4f73a21d8e083131a

src/config/runtime-config/utils.ts
Junjiequan: this utility is removed as we sync only once on startup

src/config/runtime-config/runtime-config.controller.ts
Junjiequan: fixed

src/config/runtime-config/dto/runtime-config.dto.ts
Junjiequan: description is removed for now, we can add it when there's really need for it.

src/config/runtime-config/dto/runtime-config.dto.ts
Junjiequan: created cid fields for unique identifier

src/config/runtime-config/runtime-config.controller.ts
Junjiequan: Added all missing descriptors

nitrosx (approved):

Junjiequan:
@sbliven All the previous comments have been addressed. if there's no more comments, I'm planning to merge the PR in this week as this PR has been stalled for quite long time.

PR #2379, opened 2025-11-26
fix: convert publicationYear field from number to string

HayenNico:
## Description
Tiny fix for publication wirkflow to convert publicationYear from number to string as is expected by the DataCite API

## Motivation
Registration atempts fail with a 422 error due to wron publicationYear type

## Fixes

* Change type of publicationYear (mangaed by system, configuration ignored by FE) to string
* Add migration script to fix the type for existing publishedData instances

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/migrations/20251126180700-change-publicationYear-type.js
+++ b/migrations/20251126180700-change-publicationYear-type.js
@@ -1,0 +1,29 @@
+module.exports = {
+  async up(db, client) {
+    await db
+      .collection("PublishedData")
+      .find({})
+      .forEach(async (publishedData) => {
+        await db
+          .collection("PublishedData")
+          .updateOne(
+            { _id: publishedData._id },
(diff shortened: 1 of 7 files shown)


fpotier (commented):
I'm not sure this is correct, the [API specs](https://support.datacite.org/reference/post_dois) use integer for `publicationYear`. The examples might be wrong, have you tested with a string?
According to the thread on slack, the 422 might be caused because your instance is configured for the MDS (XML) API that was used in v3 but v4 uses the JSON one.

HayenNico:
@fpotier I believe the DataCite API also accepts strings for publicationYear (I've tested, this works when running this branch), but you are correct that integers should be accepted. I had quickly thrown this PR together in response to an isue raised by @rkweehinzmann, but I believe the problem lies elsewhere for that.

I'll close this PR, no point in making an unnecessary change that requires a DB mirgation

PR #2383, opened 2025-11-27
refactor: simplify origs crud

minottic:
## Description
Simplify origdatablock cruds code by exiting when empty and avoiding reduce logic. It's open against #2382 

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -2112,28 +2112,24 @@ export class DatasetsController {
       Action.DatasetOrigdatablockCreate,
     );
 
-    if (dataset) {
-      const createOrigDatablock: CreateOrigDatablockDto = {
-        ...createDatasetOrigDatablockDto,
-        datasetId: pid,
-        ownerGroup: dataset.ownerGroup,
-        accessGroups: dataset.accessGroups,
-        instrumentGroup: dataset.instrumentGroup,
-      };
-      const datablock =
-        await this.origDatablocksService.create(createOrigDatablock);
+    if (!dataset) throw new NotFoundException(`dataset: ${pid} not found`);
 
-      const updateDatasetDto: PartialUpdateDatasetObsoleteDto = {
-        size: dataset.size + datablock.size,
-        numberOfFiles: dataset.numberOfFiles + datablock.dataFileList.length,
-      };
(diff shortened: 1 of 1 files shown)


src/datasets/datasets.controller.ts
```diff
@@ -2297,28 +2293,27 @@ export class DatasetsController {
       pid,
       Action.DatasetOrigdatablockUpdate,
     );
+    if (!dataset) throw new NotFoundException(`dataset: ${pid} not found`);
```
omkar-ethz: `checkPermissionsForDatasetExtended` already throws NotFoundException, so I believe this is not needed.

src/datasets/datasets.controller.ts
```diff
@@ -2403,29 +2398,20 @@ export class DatasetsController {
       pid,
       Action.DatasetOrigdatablockDelete,
     );
+    if (!dataset) throw new NotFoundException(`dataset: ${pid} not found`);
```
omkar-ethz: ditto

sbliven:
This is just a refactor, right? No behavioral changes? Given @omkar-ethz's comment, I guess it doesn't even change a null return to an exception?

src/datasets/datasets.controller.ts
```diff
@@ -2297,28 +2293,27 @@ export class DatasetsController {
...
-            origDatablockBeforeUpdate.dataFileList.length +
-            origDatablock.dataFileList.length,
-        });
-        return origDatablock;
-      }
-    }
-    return null;
+
+    if (!origDatablockBeforeUpdate)
+      throw new NotFoundException(`origDatablock: ${oid} not found`);
```
omkar-ethz, line 2299: @sbliven i think this bit is still change is behavior: before null was returned now exception is thrown

minottic: yes, it's a small change outside of the refactor, though it's inline with other deletes with non existing IDs, for example the DELETE datasets/{id}. I'll be happy to remove it, if you are afraid it breaks clients, though I think no one uses this

omkar-ethz (approved):

PR #2384, opened 2025-11-27
feat: metadata keys collection

nitrosx:
## Description
This PR introduces a standalone MetadataKeys Synchronization Service and replaces the legacy GET /datasets/metadataKeys endpoint. Metadata key search and synchronization are now handled in a dedicated module with explicit permission handling.

## Motivation
The previous implementation lived in the Datasets service and had no proper permission-based filtering. This resulted in performance issues, instability when limits set high, and potential exposure of metadata keys to users without access.

## Changes:
- Added MetadataKeysController with GET /metadatakeys as the new search endpoint
- Implemented permission-aware search logic using CASL
- Introduced MetadataKeysService:
   - `findAll` as the only public-facing, permission-protected method
   - Internal sync methods: `insertManyFromSource`, `deleteMany`, `replaceManyFromSource`
- Centralized extraction of keys from scientificMetadata, metadata, and customMetadata
- Improved performance, using aggregation pipeline compared to the legacy endpoint 

## Tests included
- [ ] Passing? 

## Documentation


nitrosx (commented):
It looks good.

Junjiequan (approved):

Junjiequan:
I will update description and add some test

src/config/frontend.config.json
minottic: not in the scope of the PR?

src/attachments/attachments.v4.controller.ts
minottic: not in the scope of the PR?

PR #2391, opened 2025-12-02
feat: include history in gets only when explicit

minottic:
## Description
Replays v3 history only when explicitly requested

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -594,21 +594,10 @@ export class DatasetsController {
         }
       }
 
-      const includeHistory =
-        Array.isArray(fields) && fields.includes("history");
-
-      if (includeHistory) {
-        const currentDataset = (await this.datasetsService.findOne({
-          where: { pid: inputDataset._id },
-        })) as DatasetDocument;
-        propertiesModifier.history = convertGenericHistoriesToObsoleteHistories(
-          await this.historyService.find({
-            documentId: inputDataset._id,
-            subsystem: "Dataset",
-          }),
-          currentDataset,
+      if (Array.isArray(fields) && fields.includes("history"))
+        propertiesModifier.history = await this.convertToObsoleteHistory(
+          inputDataset._id,
         );
-      }
     }
 
     const outputDataset = {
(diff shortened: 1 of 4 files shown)


test/jest-e2e-tests/datasetHistory.e2e-spec.ts
minottic, line 70: retrievable is missing

omkar-ethz (approved):
eslint fails for some reason, other than that LGTM - thanks for the fix!

src/datasets/datasets.controller.ts
```diff
@@ -1281,7 +1289,10 @@ export class DatasetsController {
     });
     if (dataset.length == 0)
       throw new ForbiddenException("Unauthorized access");
-    return dataset[0] as OutputDatasetObsoleteDto;
+    return {
+      ...dataset[0],
+      history: await this.convertToObsoleteHistory(dataset[0].pid),
```
omkar-ethz, line 1294: (minor, but maybe worth fixing)
i think history should not be included unconditionally - rather only if present in `fields` filter, or if no fields filter is provided.

also, if the `fields` filter contained history, then `findAll` will already have included it, so this additional db call and computation would be redundant.

i think simply checking if `fields` is undefined/empty here would fix the issue, (pseudocode):
```ts
if (_.isUndefined(filterObj.fields) || _.isEmpty(filterObj.fields)) {
  return {
    ...dataset[0],
    history: await this.convertToObsoleteHistory(dataset[0].pid),
  };
} else return dataset[0];
```

test/jest-e2e-tests/datasetHistory.e2e-spec.ts
```diff
@@ -74,26 +76,55 @@ describe("Test v3 history in datasetLifecycle", () => {
...
+        .expect(TestData.SuccessfulGetStatusCode)
+        .then((res) => {
+          expect(res.body[0].datasetName).toBeDefined();
+          expect(res.body[0].history).toBeUndefined();
+        });
+    });
+  });
+
+  it("Should check v3 built history with field including history and datasetName", async () => {
     const filter = { fields: ["datasetName", "history"] };
```
omkar-ethz, line 103: (minor, related to comment in dataset.controller) i think even with `fields: ["datasetName"]` history will be returned. perhaps this is not the intended behavior.

src/datasets/datasets.controller.ts
minottic: ah you are right, thanks, let me fix it!

minottic: should be fixed by #2407

PR #2392, opened 2025-12-02
fix(publisheddata): return the correct format for `publisher` in v3

fpotier:
## Description

V3 api should return `"publisher": "string"` not `"publisher: { "name": "string" }`
## Changes:

- fix `publisher` property in reponse
- use class-transformer for serialization of the reponse

## Tests included

- [ ] Passing? 

## Documentation


src/published-data/dto/published-data.obsolete.dto.ts
```diff
@@ -1,8 +1,47 @@
...
+import _ from "lodash";
+import { PublishedDataStatus } from "../interfaces/published-data.interface";
+import { PublishedData } from "../schemas/published-data.schema";
+
+function mapPublishedDataV3toV4Field(
+  publishedData: PublishedData,
+  key: keyof PublishedData | keyof PublishedDataObsoleteDto | string,
+):
+  | PublishedDataObsoleteDto[keyof PublishedDataObsoleteDto]
+  | PublishedData[keyof PublishedData]
```
minottic: if this is v3 to v4, why does it need `PublishedData`? Isn't this the v4 DTO?

src/published-data/dto/published-data.obsolete.dto.ts
```diff
@@ -1,8 +1,47 @@
...
+  | unknown
```
minottic: why `unknown`?

src/published-data/dto/published-data.obsolete.dto.ts
```diff
@@ -23,6 +63,8 @@ export class PublishedDataObsoleteDto {
   })
   @IsString()
   @IsOptional()
+  @Expose()
+  @Transform(({ obj, key }) => mapPublishedDataV3toV4Field(obj, key))
```
minottic: maybe it's worth adding `toClassOnly` evereywhere, since that's the only place where the transformation is required

src/published-data/dto/published-data.obsolete.dto.ts
```diff
@@ -188,6 +301,22 @@ export class PublishedDataObsoleteDto {
   })
   @IsString({ each: true })
   @IsOptional()
+  @Expose()
+  @Transform(
+    ({ obj, key }) => {
```
minottic: all these functions are the same I believe, e.g. for `relatedIdentifiers` and `name` etc. Maybe it's woth adding them to the `mapPublishedDataV3toV4Field` or having a dedicated function to reduce duplication

minottic: (also, maybe it's a bit cleaner to invert the if:

```TS
      if (!Array.isArray(relatedIdentifiers)) return null;
      return relatedIdentifiers
        .filter(
          (i) =>
            typeof i === "object" && i !== null && "relatedIdentifier" in i,
        )
        .map((i) => i.relatedIdentifier);
```

src/published-data/published-data.controller.ts
```diff
@@ -131,8 +121,9 @@ export class PublishedDataController {
     }
 
     if ("publisher" in inputObsoletePublishedData) {
-      propertiesModifier.metadata.publisher =
-        inputObsoletePublishedData.publisher;
+      propertiesModifier.metadata.publisher = {
+        name: inputObsoletePublishedData.publisher,
```
minottic, line 125: this is a bug fix?

src/published-data/published-data.controller.ts
```diff
@@ -484,21 +438,26 @@ export class PublishedDataController {
...
+  ): Promise<PublishedDataObsoleteDto> {
     const publishedData = await this.publishedDataService.findOne(idFilter);
     if (!publishedData) {
       throw new NotFoundException(
         `No PublishedData with the id '${idFilter["doi"]}' exists`,
       );
     }
 
-    return this.convertCurrentToObsoleteSchema(publishedData);
+    // return this.convertCurrentToObsoleteSchema(publishedData);
```
minottic: remove comment?

src/published-data/published-data.controller.ts
```diff
@@ -296,7 +246,7 @@ export class PublishedDataController {
     const createdPublishedData =
       await this.publishedDataService.create(publishedDataDto);
 
-    return this.convertCurrentToObsoleteSchema(createdPublishedData);
+    return createdPublishedData as unknown as PublishedDataObsoleteDto;
```
minottic, line 259: I believe this casting to `unknown` is necessary, due to the missing overlap? That's a pity...

src/published-data/published-data.controller.ts
```diff
@@ -230,51 +221,6 @@ export class PublishedDataController {
     return outputPublishedData;
   }
 
-  convertCurrentToObsoleteSchema(
-    inputPublishedData: PublishedData | null,
-  ): PublishedDataObsoleteDto {
-    if (!inputPublishedData) {
-      throw new BadRequestException(
```
minottic, line 237: I think the throw will never be thrown with this refactor. Maybe it's ok, just to be aware...

minottic (commented):
are there tests anywhere (either unit or mocha) that check the transformations?

test/PublishedData.js
```diff
@@ -58,7 +58,7 @@ describe("1600: PublishedData: Test of access to published data", () => {
       .expect(TestData.EntryCreatedStatusCode)
       .expect("Content-Type", /json/)
       .then((res) => {
-        res.body.should.have.property("publisher").and.be.string;
+        res.body.should.have.property('publisher').that.is.a('string');
```
minottic: minor... everywhere else is `"`

src/published-data/dto/published-data.obsolete.dto.ts
fpotier: because metadata is defined as `metadata?: Record<string, unknown>;`

src/published-data/published-data.controller.ts
fpotier: not 100% sure but I think this was never executed, class-validator already returns 400 on bad input (missing properties, invalid json)

src/published-data/dto/published-data.obsolete.dto.ts
```diff
@@ -23,6 +79,10 @@ export class PublishedDataObsoleteDto {
   })
   @IsString()
   @IsOptional()
+  @Expose()
```
Junjiequan, line 80: Do we really need to use `@Expose` in this file?

If Im not mistaken `@Expose` should only be used when `excludeExtraneousValues` is enabled or if the class is decorated with `@Exclude` to specify which field to expose within the output

minottic: I believe it's because when the serialization from the DB data to the DTO takes place, the DB fields that are not in the DTO are not removed if `excludeExtraneousValues` is disabled. Thus the endpoints are decorated with the serializer with `excludeExtraneousValues` enabled and then one needs to be explicit on what to expose

src/published-data/dto/published-data.obsolete.dto.ts
minottic: what I meant is, why does this return `PublishedData[keyof PublishedData]`? Isn't this supposed to return `PublishedData[keyof PublishedData]` only? same for `unkown`

src/published-data/dto/published-data.obsolete.dto.ts
Junjiequan: Ah sorry, I missed the part where `excludeExtraneousValues: true` is defined 😓.  It makes sense

minottic (commented):
I have a bit the impression tests are missing, at least for the bug fix, I expected to see a test now. 

For what concerns the refactor, that looks very nice IMO! Was this mapping ever tested, even before your refactor?

fpotier:
> I have a bit the impression tests are missing, at least for the bug fix, I expected to see a test now.
> 
> For what concerns the refactor, that looks very nice IMO! Was this mapping ever tested, even before your refactor?

tests for the mapping were not really existing, my last commit adds some

fpotier:
this also fixes a bug where `publicationYear` would always be overwritten with the current year

minottic (commented):
I think the tests from [8358def](https://github.com/SciCatProject/scicat-backend-next/pull/2392/commits/8358def280348fc0af7d50419b57b1ef85365423) are great, but since they also required a bunch of bug fixes, would it make sense to move that commit to a dedicated PR (if cherry-picking the commit into master is not too conflict prone)? then we limit this to a refactor and the git history will look cleaner.

minottic (approved):
> I think the tests from [8358def](https://github.com/SciCatProject/scicat-backend-next/pull/2392/commits/8358def280348fc0af7d50419b57b1ef85365423) are great, but since they also required a bunch of bug fixes, would it make sense to move that commit to a dedicated PR (if cherry-picking the commit into master is not too conflict prone)? then we limit this to a refactor and the git history will look cleaner.

actually forget this, I missed before that this PR already contains bug fixes, and it's not only a refactor. 

Maybe it's worth though to open an issue to refactor the convertObsoleteToCurrent in the same way leveraging DTOs

PR #2407, opened 2025-12-09
fix: exclude history if fields

minottic:
## Description
Fixes the bug introduced by #2391 and raised by [this comment](https://github.com/SciCatProject/scicat-backend-next/pull/2391#discussion_r2602403238)

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -1289,6 +1289,11 @@ export class DatasetsController {
     });
     if (dataset.length == 0)
       throw new ForbiddenException("Unauthorized access");
+    if (
+      Array.isArray(filterObj?.fields) &&
+      !filterObj.fields.includes("history")
+    )
+      return dataset[0];
     return {
       ...dataset[0],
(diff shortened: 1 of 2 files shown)


src/datasets/datasets.controller.ts
```diff
@@ -1289,6 +1289,11 @@ export class DatasetsController {
...
     if (dataset.length == 0)
       throw new ForbiddenException("Unauthorized access");
+    if (
+      Array.isArray(filterObj?.fields) &&
+      !filterObj.fields.includes("history")
+    )
+      return dataset[0];
     return {
       ...dataset[0],
       history: await this.convertToObsoleteHistory(dataset[0].pid),
```
omkar-ethz: thx! this fixes the behavior, but history is computed twice if `fields` contains history: once here, and once in findAll. The new commit fixes this as well. i.e. only compute history in the controller if fields is undefined / empty. Other case is handled in findAll.

omkar-ethz (approved):
LGTM after the new commit!

PR #2408, opened 2025-12-10
feat: include username in groups when roles

minottic:
## Description
This PR ports [the behaviour of adding the username](https://github.com/SciCatProject/backend-v3/blob/master/server/boot/0-script.js#L215) in the list of currentGroups for users with "roles", as it happened in the old backend

Adds tests where missing

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/auth/guards/jwt-auth.guard.spec.ts
+++ b/src/auth/guards/jwt-auth.guard.spec.ts
@@ -1,0 +1,98 @@
+import { Test, TestingModule } from "@nestjs/testing";
+import { RolesService } from "src/users/roles.service";
+import { UsersService } from "src/users/users.service";
+import { ConfigService } from "@nestjs/config";
+import { JWTUser } from "../interfaces/jwt-user.interface";
+import { JwtStrategy } from "../strategies/jwt.strategy";
+import { User } from "src/users/schemas/user.schema";
+
+describe("JwtStrategy", () => {
+  let strategy: JwtStrategy;
+  let rolesService: Partial<RolesService>;
+  let usersService: Partial<UsersService>;
+  let configService: Partial<ConfigService>;
+  const userPayload = {
+    _id: "user-id-123",
+    username: "john_doe",
+    email: "john@example.com",
+    authStrategy: "jwt",
+  } as User;
+
+  beforeEach(async () => {
+    rolesService = {
+      find: jest.fn(),
+    };
+
+    usersService = {
+      findByIdUserIdentity: jest.fn(),
+    };
+
+    configService = {
+      get: jest.fn().mockReturnValue("testSecret"),
+    };
+
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [
+        JwtStrategy,
+        { provide: RolesService, useValue: rolesService },
+        { provide: UsersService, useValue: usersService },
+        { provide: ConfigService, useValue: configService },
+      ],
+    }).compile();
+
+    strategy = module.get<JwtStrategy>(JwtStrategy);
+  });
+
+  it("should be defined", () => {
+    expect(strategy).toBeDefined();
+  });
+
+  describe("validate", () => {
+    it("should return JWTUser with roles and accessGroups", async () => {
+      (rolesService.find as jest.Mock).mockResolvedValue([
+        { name: "admin" },
+        { name: "editor" },
+      ]);
+
+      (usersService.findByIdUserIdentity as jest.Mock).mockResolvedValue({
+        profile: {
+          username: "john_doe",
+          accessGroups: ["group1", "group2"],
+        },
+      });
+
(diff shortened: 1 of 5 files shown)


fpotier (approved):

nitrosx (commented):
I have hard time to justify saving usernames in the accessGroups. Besides making the name accessGroups misleading, it will mix groups and usernames in the same list with no easy method to distinguish what is a group and what is a username.

Let\s say that my user name is `abc` and a dataset is accessible from the group `abc` which is legit. That will cause confusion and grant access to dataset that user should not have access to

nitrosx:
We should  find a different solution to fits better the current BE but allows you to reduce to the minimum changes on you side.

nitrosx:
After a conversation with @minottic, here is the use case that this PR is trying to address:
- each beamline has a functional account to ingest datasets and review them
- the beamline functional account is part of the ingestor group
- the ingestor group is listed in the `create_dataset_priviliged` which allows to create datasets for anybody
- the datasets created at the beamline are owned by the user/proposal group
- the beam line functional user has to have access to all the datasets produced on that beamline

Proposed solutions:
_Solution #1_
- change the beamline functional account `role` field in the functionalAccounts.json file to something specific to the account.
- replace in SciCat database the accessGroup `ingestor` with the account group.
- list in `create_dataset_privileged` all the groups for all the beamline functional accounts
- when creating datasets, make sure to add the group for the beamline functional account in the AccessGroups
- apply such change to all existing groups

Example:
- beamline functional account: BL1
- account role in functionalAccount.json: ingestorBL1 (it is added in accessGroups in UserIdentity collection)
- create_dataset_privileged = "..., ingestorBL1, ingestorBL2,..."
- Datasets from BL should have `ingestorBL1` listed in their `accessGroups` field

_Solution #2_
- add to the functional account definition in the functionalAccounts.json on optional field `roles` that is an array of string
- set such field for the beamline functional account to the an array containing the `ingestor` group and the beamline specific group
- change the code to accepts the `roles` field.
  https://github.com/SciCatProject/scicat-backend-next/blob/1116c9db6d918fbc1296460e719bdf035959163d/src/users/users.service.ts#L63
  From:
  ```
  const { role, global, ...createAccount } = account;
  ```
  To
  ```
  const { role, global, roles, ...createAccount } = account;
  ```

  https://github.com/SciCatProject/scicat-backend-next/blob/1116c9db6d918fbc1296460e719bdf035959163d/src/users/users.service.ts#L78
  From:
  ```
          if (role) {
            // add role as access group
            accessGroups.push(role);
            if (!(role in roles)) {
              roles[role] = [];
            }
            roles[role].push(user._id.toString());
          }
  ```
  To:
  ```
          if (role) {
            // add role as access group
            accessGroups.push(role);
            if (!(role in roles)) {
              roles[role] = [];
            }
            roles[role].push(user._id.toString());
         } 
         if (roles) {
            // add each role as access group
            for (const role of roles) {
              accessGroups.push(role);
              if (!(role in roles)) {
                roles[role] = [];
              }
              roles[role].push(user._id.toString());
          }
  ```
- list in `create_dataset_privileged` only the `ingestor` group
- when ingesting dataset, add in `accessGroups` the group related to the beamline 
- update older datasets adding in their `accessGroups` the group related to their beamline

Example:
- beamline functional account: BL1
- account roles in functionalAccount.json: ingestor, ingestorBL1  (they are added in accessGroups in UserIdentity collection)
- create_dataset_privileged = "..., ingestor,..."
- Datasets from beamline should have `ingestorBL1` listed in their `accessGroups` field

With this solution accessGroups will contains only groups, which makes it easy to maintain.

minottic:
thanks! these work very well, but they require some work and we have the planned upgrade date the first week of Jan. 

for example:
1. would require me to mantain a very long list of roles in the create_priviliged...
2. I can give 2 a try, but also requires some work, as I need to recreate all accounts and might require a migration + the code change

if 2 takes too much time, I could suggest adding a new env var, called READ_PRIVILIGED_BY_USERNAME="ingestor" and all accounts with "ingestor" role will be able to read based on their username

nitrosx:
Given the time constraints, you can approach #2 as follow:
- update functional accounts directly in db by adding the beamline specific account. For example:
  ```
  {
    ...
    authStrategy: 'local',
    profile: {
      displayName: 'Ingestor Beamline 1',
      email: 'BL1@your.site',
      username: 'ingestorBL1',
      accessGroups: [ 'ingestor', 'ingestorBL1`],
      ...
    },
    provider: 'local',
  }
  ```
- ask the beamline to add the group at ingestion time
- update all datasets adding the beamline group, such as `ingestorBL1`, to all the relevant datasets
At this point you are done with your migration.

Now we change the code and you can update your functional account json file

nitrosx:
We implemented solution number 2 with PR #2411 
This PR is not needed anymore.

Issue #2357, opened 2025-11-19
Add 'reply-to' to graph email method


sbliven:
## Summary

Allow the 'reply-to' field of emails to be set via an env variable. This should work similarly to SMTP.

PR #2410, opened 2025-12-11
feat: allow setting 'replyTo' in emails

sbliven:
## Description

Adds `EMAIL_REPLYTO` env variable. If set, emails will be configured to reply to this address.

## Motivation

This was supported in v3 and requested in #2357.

## Changes:

* Allow reply-to field to be set in emails #2357 

## Tests included

- [x] Passing? 

## Documentation


src/common/graph-mail.ts
```diff
@@ -121,7 +140,7 @@ export class MSGraphMailTransport implements Transport {
     accessToken: string,
     mail: MailMessage,
   ): Promise<SentMessageInfo> {
-    const { to, subject, text, html, from } = mail.data;
+    const { to, subject, text, html, from, replyTo, cc, bcc } = mail.data;
```
minottic, line 142: where do cc and bcc come from?

src/common/graph-mail.ts
```diff
@@ -10,6 +10,7 @@ import { Address } from "nodemailer/lib/mailer";
 import { firstValueFrom } from "rxjs";
 import { Injectable, Logger } from "@nestjs/common";
 import { isAxiosError } from "@nestjs/terminus/dist/utils";
+import { get } from "lodash";
```
minottic: I believe this is not used (at least eslint says so)

minottic (approved):
minor comments

src/common/graph-mail.ts
sbliven: Currently they don't have env variables assigned to them and are undefined in my tests. However, `mail.data` gets set by the Node MailerService, so I just implemented it in graph-mail for consistency & future-proofing (it is supported by SMTP by default).

src/common/graph-mail.ts
sbliven: Fixed. Weird it wasn't flagged by the CI lint action

minottic: I believe it's beacuase it's a warning

PR #2411, opened 2025-12-12
feat: extend fa role to be a list

minottic:
## Description
Following from #2408 this extends role in fa to be a list. This covers #2408 by adding the username in the role adds it to the allowed accessGroups. 

The field is kept `role`, even thought it should be `roles` to keep backward compatibility.

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/functionalAccounts.json.minimal.example
+++ b/functionalAccounts.json.minimal.example
@@ -24,7 +24,7 @@
     "username": "proposalIngestor",
     "email": "scicatproposalingestor@your.site",
     "password": "aman",
-    "role": "proposalingestor",
+    "role": ["proposalingestor"],
     "global": true
   }
 ]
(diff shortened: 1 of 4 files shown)


functionalAccounts.json.minimal.example
```diff
@@ -24,7 +24,7 @@
     "username": "proposalIngestor",
     "email": "scicatproposalingestor@your.site",
     "password": "aman",
-    "role": "proposalingestor",
+    "role": ["proposalingestor"],
```
nitrosx, line 27: Can we change it to `roles` so it is clear that we are expecting a list?

minottic: it's in the descriptioon of the pr, the reason is this:

```
The field is kept role, even thought it should be roles to keep backward compatibility.
```

if I changed it to `roles` anyone pulling after this PR would need to update their functional accounts file

nitrosx (approved):

PR #2415, opened 2025-12-15
fix(publisheddata): use default public URL prefix when registering a DOI

fpotier:
## Description

The `url` used when registering a DOI defaults to the configured LandingPage url unless a `metadata.landingPage` is defined.
## Motivation

We would prefer to omit the landing page url when registering a `PublishedData` and default to the configured LandingPage url.
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [x] Passing? 

## Documentation


src/published-data/published-data.v4.controller.ts
```diff
@@ -717,65 +718,69 @@ export class PublishedDataV4Controller {
...
+      relatedItems,
+      relatedIdentifiers,
+      language,
+      dates,
+      sizes,
+      formats,
+      geoLocations,
+      fundingReferences,
+    } = metadata || {};
+    let { url } = metadata || {};
```
fpotier: should this remain `landingPage`? not sure if anyone is already using it

src/published-data/published-data.v4.controller.ts
```diff
@@ -717,65 +718,69 @@ export class PublishedDataV4Controller {
...
+      sizes,
+      formats,
+      geoLocations,
+      fundingReferences,
+      landingPage,
+    } = metadata || {};
+
+    const url = _.isNil(landingPage)
+      ? `${this.configService.get<string>("publicURLprefix")}${encodeURIComponent(doi)}`
+      : `https://${landingPage}${encodeURIComponent(doi)}`;
```
fpotier: Is anyone using this?
The current format is weird (`$hostname/`) to make a valid url. 
I think it should either be the full prefix (scheme + hostname + /) or only the hostname

src/published-data/published-data.v4.controller.ts
```diff
@@ -717,65 +718,69 @@ export class PublishedDataV4Controller {
...
-        relatedItems: relatedItems,
-        relatedIdentifiers: relatedIdentifiers,
-        language: language,
-        dates: dates,
-        sizes: sizes,
-        formats: formats,
-        geoLocations: geoLocations,
-        fundingReferences: fundingReferences,
-        url: `https://${landingPage}${encodeURIComponent(doi)}`,
+  doiRegistrationJSON(publishedData: PublishedData): object {
```
minottic, line 721: should this be private?

src/published-data/published-data.v4.controller.ts
```diff
@@ -717,65 +718,69 @@ export class PublishedDataV4Controller {
...
+      relatedItems,
+      relatedIdentifiers,
+      language,
+      dates,
+      sizes,
+      formats,
+      geoLocations,
+      fundingReferences,
+      landingPage,
+    } = metadata || {};
```
minottic: prefer `??` ?

src/published-data/published-data.v4.controller.ts
```diff
@@ -717,65 +718,69 @@ export class PublishedDataV4Controller {
...
+
+    const url = _.isNil(landingPage)
```
minottic: why not the standard 

```
landingPage ? `${this.configService.get<string>("publicURLprefix")}${encodeURIComponent(doi)}`
      : `https://${landingPage}${encodeURIComponent(doi)}`;Collapse comment
```
?

minottic (approved):
minor comments, thanks for the changes!

src/published-data/published-data.v4.controller.ts
fpotier: that's what I initially did but needed to call it from outside for testing, is there a better way to do?

src/published-data/published-data.v4.controller.ts
fpotier: that was the original code but happy to replace it

src/published-data/published-data.v4.controller.ts
minottic: I think in tests you would then need to access it with [''] and any before. But I am always confused what's better

src/published-data/published-data.v4.controller.ts
fpotier: looks like it is set by the frontend

PR #2425, opened 2025-12-17
fix: return last identity

minottic:
## Description
As multiple identities can in principle exist for one username, return the last


fpotier (approved):
is there any test?

minottic:
> is there any test?

fair point. I wasn't sure how to test this, I'll think of something

PR #2430, opened 2025-12-22
Send email on doi registration desy changes

rkweehinzmann:
DRAFT PR for
The idea was that we can have emails as a separate action, not only happening after a job, but refactor the email action logic so that it can be used as a standalone module whenever needed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,41 @@
+stages:
+  - build
+  - publish
+
+build_image:
+  stage: build
+  image:
+    name: gcr.io/kaniko-project/executor:debug
+    entrypoint: [""]
+  script:
+    - |-
+      tag=$CI_COMMIT_SHORT_SHA
+      if [[ -n "$CI_COMMIT_TAG" ]]; then
+        tag=$CI_COMMIT_TAG
+      fi
+    - mkdir -p /kaniko/.docker
+    - echo "{\"auths\":{\"$CI_REGISTRY\":{\"username\":\"$CI_REGISTRY_USER\",\"password\":\"$CI_REGISTRY_PASSWORD\"}}}" > /kaniko/.docker/config.json
+    - >
+      /kaniko/executor 
+      --context $CI_PROJECT_DIR
+      --dockerfile $CI_PROJECT_DIR/Dockerfile
+      --destination $CI_REGISTRY_IMAGE:${tag}
+      --cache=true
+
+push_image_harbor:
+  stage: publish
+  image:
+    name: gcr.io/kaniko-project/executor:debug
+    entrypoint: ['']
+  script:
+    - mkdir -p /kaniko/.docker
+    - echo "{\"auths\":{\"${HARBOR_HOST}\":{\"auth\":\"$(echo -n ${HARBOR_USERNAME}:${HARBOR_PASSWORD} | base64 -w 0)\"}}}" > /kaniko/.docker/config.json
+    - >-
+      /kaniko/executor
+      --context "${CI_PROJECT_DIR}"
+      --dockerfile ${CI_PROJECT_DIR}/Dockerfile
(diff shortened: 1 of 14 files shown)


src/common/services/mail.service.ts
```diff
@@ -0,0 +1,80 @@
...
+          "MailService.sendMail",
+        );
+        Logger.error(error, "MailService.sendMail");
+      }
+    }
+  }
+
+ // Fetch the email content for a template from the configuration
+  getEmailContent(templateName: string, doi: string, userEmail: string): { subject: string; body: string } | null {  
+    const emailTemplates = this.configService.get("emailTemplate")
```
sbliven: emailTemplates is the filename. You need to parse it at some point; either here (needs to be parsed for each email) or once on startup (see how JOB_CONFIG_FILE gets parsed). I suggest using the yaml parser (which also supports json).

src/common/services/mail.service.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (emailTemplates) {      
+      const subject = emailTemplates.subject.replace("{doi}", doi);
```
sbliven: This should use handlebars for templating. Rather than passing the doi, I would pass a context object with a well-defined interface. I recommend keeping it similar to the JobTemplateContext for consistency, eg

```
export interface PublicDatasetContext<DtoType extends JobDto> {
  publishedData: PublishedData; // res
  env: Record<string, string | undefined>; //process.env
}
```

The method should be similar to `EmailAction.perform(contxt)`

.gitlab-ci.yml
```diff
@@ -0,0 +1,39 @@
+stages:
```
sbliven, line 1: I guess .gitlab-ci is desy-specific and should not be included in the final PR

src/common/common.module.ts
```diff
@@ -1,5 +1,5 @@
 import { Module } from "@nestjs/common";
-import { MailService } from "./mail.service";
```
sbliven, line 2: We don't have 'services' directories elsewhere in the project. Is this file rename necessary?

sbliven (commented):
Generalizing EmailAction.perform so that it can be used in other places is a good idea. I have a few suggestions how to implement that.

The JSON file keys are fundamentally a way to configure an event that is run at the end of the publishedDataset/register call. I think a more idiomatic way to implement this would be to use the [NestJS Event Emitter](https://docs.nestjs.com/techniques/events) package. That way we can easily emit events for other endpoints if needed. By using the event emitter there is a clear place to parse the file (the module implementing the handler functions), something that is currently lacking.

src/common/services/mail.service.ts
sbliven: Ah, I guess you meant to inject EmailConfigService

src/common/services/email-config.service.ts
```diff
@@ -0,0 +1,33 @@
...
+    this.loadConfig();
+  }
+
+  // Loads the configuration file if it exists
+  private loadConfig(): void {
+    const configFilePath = path.join(__dirname, "email-templates.json");
+
+    if (fs.existsSync(configFilePath)) {
+      const configFile = fs.readFileSync(configFilePath, "utf-8");
+      this.config = JSON.parse(configFile);
```
sbliven: This can be very easily swapped out with `js-yaml.load` while keeping full support for json. See jobconfig.service.ts:60.

src/common/services/email-config.service.ts
```diff
@@ -0,0 +1,33 @@
...
+export class EmailConfigService {
+  private config: Record<string, any> | null = null;
+
+  constructor() {
+    this.loadConfig();
+  }
+
+  // Loads the configuration file if it exists
+  private loadConfig(): void {
+    const configFilePath = path.join(__dirname, "email-templates.json");
```
sbliven:
```suggestion
const configFilePath = this.configService.get("emailTemplate")
```

rkweehinzmann:
outdated will close it!

PR #2456, opened 2026-01-07
chore: apply casl on mongo and not in memory

minottic:
## Description
Running the casl and skip/limits in memory makes the application go down for large DBs

### DIFF
--- a/src/casl/casl-ability.factory.ts
+++ b/src/casl/casl-ability.factory.ts
@@ -17,7 +17,7 @@ import { Datablock } from "src/datablocks/schemas/datablock.schema";
 import { DatasetClass } from "src/datasets/schemas/dataset.schema";
 import { ElasticSearchActions } from "src/elastic-search/dto";
 import { Instrument } from "src/instruments/schemas/instrument.schema";
-import { JobClass } from "src/jobs/schemas/job.schema";
+import { JobClass, JobDocument } from "src/jobs/schemas/job.schema";
 import { CreateJobAuth, UpdateJobAuth } from "src/jobs/types/jobs-auth.enum";
 import { Logbook } from "src/logbooks/schemas/logbook.schema";
 import { OrigDatablock } from "src/origdatablocks/schemas/origdatablock.schema";
@@ -29,6 +29,8 @@ import { UserIdentity } from "src/users/schemas/user-identity.schema";
 import { UserSettings } from "src/users/schemas/user-settings.schema";
 import { User } from "src/users/schemas/user.schema";
 import { Action } from "./action.enum";
+import { accessibleBy } from "@casl/mongoose";
+import { FilterQuery } from "mongoose";
 
 type Subjects =
   | string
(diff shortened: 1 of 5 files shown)


fpotier (approved):

src/casl/casl-ability.factory.ts
```diff
@@ -1833,27 +1840,60 @@ export class CaslAbilityFactory {
...
+  jobsMongoQueryReadAccess(user: JWTUser) {
+    const abilities = this.jobsAccess(user);
+    return {
+      $or: [
+        accessibleBy(abilities, Action.JobReadAny).ofType(JobClass),
+        accessibleBy(abilities, Action.JobReadAccess).ofType(JobClass),
+      ],
+    };
+  }
+
```
Junjiequan, line 1896: I think we are mixing the concepts of query building and authorization in the CASL factory file, which will make it harder to maintain.

The official documentation provides a cleaner solution for future reference:
https://casl.js.org/v6/en/package/casl-mongoose#mongo-db-adapter

minottic: [that's](https://casl.js.org/v6/en/package/casl-mongoose#mongo-db-adapter) exactly what I am using, and I wanted to avoid composing the filter 3 times for the functions I use. 

I wasn't sure indeed in which file to put this function, I thought in the CASL one was the right place as it covers endpoints and instances access so I thought it would cover mongo CASLs as well, but I am happy to move it somewhere else in another PR.

BTW, this approach could simplify a lot all the query building in the datasets or in the other ones, and could also be extended to other CASLs, e.g. write or update in the same way.

If we plan a medium/big refactor with this, it might also make sense to apply these at the "Ownable" mongo hooks

Junjiequan: What I was thinking is to chain the accessibleBy directly to the DB query like this:
```JS

app.get('/api/posts/:id', async (req, res) => {
  const post = await Post.accessibleBy(ability).findByPk(req.params.id);
  res.send(pick(post, post.accessibleFieldsBy(ability))
});
```
But either way, I think it's a good initiatives to start using accessibleBy.

I've created a plan to refactor using `accessibleBy` quite sometimes ago, but I didn't have time to work on it yet https://github.com/SciCatProject/scicat-backend-next/issues/2122, please feel free to comment or edit it if something is unclear

minottic: [that's](https://casl.js.org/v6/en/package/casl-mongoose#accessible-records-plugin) actually deprecated, I started with that and then changed to the utility function.

Junjiequan: Ah, I see. Maybe a wrapper like this maybe works? `findByIdWithAccess`, we can discuss more when we actually start working on it

sbliven:
What's the performance penalty from this? It's necessary for the PSI database, but I guess it must be quite a bit slower, right?

src/casl/casl-ability.factory.ts
minottic: but the disadvantage there would be that it's not applicable to aggregation pipelines. TBH, I also find being able to see the query which would be applied quite useful, and that's easy with the utility function. Happy to discuss further.

minottic:
> What's the performance penalty from this? It's necessary for the PSI database, but I guess it must be quite a bit slower, right?

It's only performance gains (quite huge). way more than before that loaded all documents in memory and made the app restart. "Queries like logics" outside of mongo are always discouraged

PR #2469, opened 2026-01-14
refactor: add encoding/decoding for samples metadata keys

abdimo101:
## Description
This PR implements encoding and decoding of metadata keys in `sampleCharacteristics` to handle special characters in MongoDB.

 * Added `@Transform` decorators in `OutputSampleDto` and `UpdateSampleDto` to encode/decode metadata keys
 * Updated `samples.controller.ts` and `samples.service.ts` to use DTOs.
 * Created migration script `20260114145500-encode-sample-metadatakeys.js` to encode existing sample metadata keys in the database
 * Added tests for metadata keys in `Sample.js`

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/migrations/20260114145500-encode-sample-metadatakeys.js
+++ b/migrations/20260114145500-encode-sample-metadatakeys.js
@@ -1,0 +1,68 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
+const {
+  encodeScientificMetadataKeys,
+  decodeScientificMetadataKeys,
+} = require("../dist/common/utils");
+
+module.exports = {
+  async up(db, client) {
+    for await (const sample of db
+      .collection("Sample")
+      .find({ sampleCharacteristics: { $exists: true } })) {
+      const metadata = sample.sampleCharacteristics;
+      if (!metadata || typeof metadata !== "object") continue;
+
+      let encodedMetadata;
+      try {
+        encodedMetadata = encodeScientificMetadataKeys(metadata);
+      } catch (err) {
+        console.error(
+          `Error encoding sampleCharacteristics for Sample (Id: ${sample._id}):`,
+          err,
+        );
+        continue;
+      }
+
+      console.log(
+        `Updating Sample (Id: ${sample._id}) with encoded sampleCharacteristics keys`,
+      );
+      await db
+        .collection("Sample")
+        .updateOne(
+          { _id: sample._id },
+          { $set: { sampleCharacteristics: encodedMetadata } },
+        );
+    };
+  },
+
+  async down(db, client) {
+    for await (const sample of db
+      .collection("Sample")
+      .find({ sampleCharacteristics: { $exists: true } })) {
+      const metadata = sample.sampleCharacteristics;
+      if (!metadata || typeof metadata !== "object") continue;
+
+      let decodedMetadata;
+
+      try {
+        decodedMetadata = decodeScientificMetadataKeys(metadata);
+      } catch (err) {
+        console.error(
+          `Error decoding sampleCharacteristics for Sample (Id: ${sample._id}):`,
+          err,
+        );
+        continue;
+      }
+
+      console.log(
+        `Reverting Sample (Id: ${sample._id}) to decoded sampleCharacteristics keys`,
+      );
+      await db
+        .collection("Sample")
+        .updateOne(
+          { _id: sample._id },
+          { $set: { sampleCharacteristics: decodedMetadata } },
+        );
+    };
+  },
+};
--- a/migrations/20260114145500-encode-sample-metadatakeys.js
+++ b/migrations/20260114145500-encode-sample-metadatakeys.js
@@ -6,6 +6,8 @@ const {
 
 module.exports = {
   async up(db, client) {
+    const bulkOps = [];
+
     for await (const sample of db
       .collection("Sample")
       .find({ sampleCharacteristics: { $exists: true } })) {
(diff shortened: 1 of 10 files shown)


migrations/20260114145500-encode-sample-metadatakeys.js
```diff
@@ -0,0 +1,68 @@
...
+        `Updating Sample (Id: ${sample._id}) with encoded sampleCharacteristics keys`,
+      );
+      await db
+        .collection("Sample")
+        .updateOne(
+          { _id: sample._id },
+          { $set: { sampleCharacteristics: encodedMetadata } },
+        );
+    };
+  },
```
Junjiequan, line 44: Can you check if [bulkWrite ](https://www.mongodb.com/docs/manual/core/bulk-write-operations/) can be used here for better performance?

abdimo101: I know use bulkWrite for both dataset and sample migrations scripts

minottic:
I have to admit I never fully understood why this was needed for datasets... since I see this here now, could you please explain me why we need this?

abdimo101:
> I have to admit I never fully understand why this was needed for datasets... since I see this here now, could you please explain me why we need this?

Hi @minottic, the reason for why encoding is needed, is to handle special characters in metadata key names that break MongoDB and Elasticsearch. MongoDB does not allow dots in field names, and Elasticsearch has issues with spaces. By encoding these characters, it ensures that metadata keys work correctly regardless of how users name them.

minottic:
> Hi @minottic, the reason for why encoding is needed, is to handle special characters in metadata key names that break MongoDB and Elasticsearch. MongoDB does not allow dots in field names, and Elasticsearch has issues with spaces. By encoding these characters, it ensures that metadata keys work correctly regardless of how users name them.

thanks for the explanation! but how did they end up in mongo then (you wouldn't need the migration scripts if mongo did not failed to store them) ?

abdimo101:
> > Hi @minottic, the reason for why encoding is needed, is to handle special characters in metadata key names that break MongoDB and Elasticsearch. MongoDB does not allow dots in field names, and Elasticsearch has issues with spaces. By encoding these characters, it ensures that metadata keys work correctly regardless of how users name them.
> 
> thanks for the explanation! but how did they end up in mongo then (you wouldn't need the migration scripts if mongo did not failed to store them) ?

MongoDB does store them, but it has limitations when field names contain [dots or even dollar signs](https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/). For example in SciCat frontend, if a metadata key contains a dot, users wont be able to perform condition filter searches using that specific metadata key.

minottic:
> MongoDB does store them, but it has limitations when field names contain [dots or even dollar signs](https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/). For example in SciCat frontend, if a metadata key contains a dot, users wont be able to perform condition filter searches using that specific metadata key.

but why do you need the migration then? Isn't it enough to translate at run time for search and returns?

Junjiequan:
@minottic the migration is for old scientific Metada keys that contains dot or dollar etc.
Mongo has poor support for keys yhat contain special characters.
Encode for write Decode for read is what we thought would be easier to maintain. runtime query transformation can do the same job, but then we may need to include the transfomer in many places, such as find, update, delete, aggregate etc..
That being said, the migration scirpt will be very slow for big number of datasets and it seems causing some troubles to your side..so if you have better idea we could also test that out

minottic:
> @minottic the migration is for old scientific Metada keys that contains dot or dollar etc.
> Mongo has poor support for keys yhat contain special characters.
> Encode for write Decode for read is what we thought would be easier to maintain. runtime query transformation can do the same job, but then we may need to include the transfomer in many places, such as find, update, delete, aggregate etc..
> That being said, the migration scirpt will be very slow for big number of datasets and it seems causing some troubles to your side..so if you have better idea we could also test that out

thanks for the explanation! It was indeed quite slow for us - thus the question -, but now it's done, so I am ultimately fine with it (and also thanks for the bulkWrite!). We have very few samples, so this new migration should not be a problem for us

src/samples/samples.controller.ts
```diff
@@ -317,17 +323,22 @@ export class SamplesController {
...
+  ): Promise<OutputSampleDto[]> {
     const sampleFilters: IFilters<SampleDocument, ISampleFields> =
       this.updateFiltersForList(request, JSON.parse(filters ?? "{}"));
-    return this.samplesService.findAll(sampleFilters);
+    const samples = await this.samplesService.findAll(sampleFilters);
+
+    const samplesObj = (samples as SampleDocument[]).map((sample) =>
+      sample.toObject(),
+    );
+    return plainToInstance(OutputSampleDto, samplesObj);
```
minottic: if you want to avoid the explicit call to plainToInstance and be more "nestJS native", you could use a serializer (thus simply returning `samples`). See [here](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/published-data/published-data.controller.ts#L244-L247) for example

minottic (approved):
a small cosmetic comment, up to you if to try it or not

src/samples/samples.controller.ts
abdimo101: Fixed in b32d01a86aeffa1869aaa645f4f6831e115f0a6f

PR #2472, opened 2026-01-15
fix(legacy-history): fix behavior of previousValue and include full datasetlifecycle in currentValue

omkar-ethz:
## Description

This PR fixes three issues with legacy history, specifically with datasetlifecycle:
1. Before, history records for `datasetlifecycle.previousValue` were reconstructed as {...current lifecycle, ...genericHistory.before}. However this had the flaw that if certain fields were added for the first time in this update, i.e. they were only in genericHistory.after but not in genericHistory.before, they would still be present in previousValue. This is now fixed.
2.   Although not a bug, PSI's Central IT relies on entire snapshot of datasetlifecycle to be present also in datasetlifecycle.currentValue, not just the changed fields (this was already the case with datasetlifecycle.previousValue) - this is also changed
3. backend-v3 returned `history[]` in /datasets response in sorted order of time. Hence reverse the response in the /datsets controller before returning. Note that the adapter `convertGenericHistoriesToObsoleteHistories` still expects records sorted by descending time - hence we don't change the DB query. Corresponding frontend change: https://github.com/SciCatProject/frontend/pull/2183

## Motivation

Some backward incompatiblities in legacy history broke tools that relied on it

## Fixes

* Bug fixed (#X)

## Tests included

- [x] Passing? 

## Documentation

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -618,7 +618,7 @@ export class DatasetsController {
         subsystem: "Dataset",
       }),
       currentDataset,
-    );
+    ).reverse();
   }
 
   // POST https://scicat.ess.eu/api/v3/datasets
--- a/src/datasets/utils/history.util.ts
+++ b/src/datasets/utils/history.util.ts
@@ -107,7 +107,8 @@ export function convertGenericHistoryToObsoleteHistory(
   return result;
 }
 
-// starting from the latest dataset, replay the history entries in reverse order to reconstruct the obsolete history entries
+// starting from the latest dataset, replay the history entries in reverse order to reconstruct the obsolete history entries.
+// assumes histories is sorted in descending order of updatedAt
 export function convertGenericHistoriesToObsoleteHistories(
   histories: GenericHistory[],
   currentDataset: DatasetDocument | DatasetClass,
--- a/src/datasets/utils/history.util.ts
+++ b/src/datasets/utils/history.util.ts
@@ -5,6 +5,7 @@ import {
   DatasetDocument,
 } from "src/datasets/schemas/dataset.schema";
 import { computeDeltaWithOriginals } from "src/common/utils/delta.util";
+import { omit } from "lodash";
 
 const IGNORE_FIELDS = ["updatedAt", "updatedBy", "_id"];
 
@@ -79,19 +80,15 @@ export function convertGenericHistoryToObsoleteHistory(
         string,
         unknown
       >;
-      const reconstructedBefore = JSON.parse(
-        JSON.stringify(datasetSnapshot.datasetlifecycle),
-      );
-
-      // Delete keys from before that are only present in after
-      // as it implies they were added in this update
-      for (const key of Object.keys(afterPartial)) {
-        if (!(key in beforePartial)) {
-          delete reconstructedBefore[key];
-        }
-      }
-      // apply before to construct previous snapshot
-      Object.assign(reconstructedBefore, beforePartial);
+      // omit keys that are only in afterPartial, as it means they were added in this update
+      // and apply beforePartial to the resulting datasetlifecycle snapshot
+      const reconstructedBefore = {
+        ...omit(
+          JSON.parse(JSON.stringify(datasetSnapshot.datasetlifecycle)),
+          Object.keys(afterPartial).filter((key) => !(key in beforePartial)),
+        ),
+        ...beforePartial,
+      };
 
       history.before[field] = reconstructedBefore;
 
--- a/src/datasets/utils/history.util.ts
+++ b/src/datasets/utils/history.util.ts
@@ -83,9 +83,15 @@ export function convertGenericHistoryToObsoleteHistory(
       // omit keys that are only in afterPartial, as it means they were added in this update
       // and apply beforePartial to the resulting datasetlifecycle snapshot
       const reconstructedBefore = {
-        ...omit(
-          JSON.parse(JSON.stringify(datasetSnapshot.datasetlifecycle)),
-          Object.keys(afterPartial).filter((key) => !(key in beforePartial)),
+        ...JSON.parse(
+          JSON.stringify(
+            omit(
+              datasetSnapshot.datasetlifecycle,
+              Object.keys(afterPartial).filter(
+                (key) => !(key in beforePartial),
+              ),
+            ),
+          ),
         ),
         ...beforePartial,
       };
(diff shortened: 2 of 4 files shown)


Junjiequan:
I'm trying to understand the second issue, is the document strategy not enough or?

```JS
      if (trackableStrategy === "document") {
        // Document strategy - store full document
        beforeData = originalDoc;
        afterData = updatedDoc ? updatedDoc.toObject() : null;
```

omkar-ethz:
> I'm trying to understand the second issue, is the document strategy not enough or?
> 
> ```js
>       if (trackableStrategy === "document") {
>         // Document strategy - store full document
>         beforeData = originalDoc;
>         afterData = updatedDoc ? updatedDoc.toObject() : null;
> ```

That's a good idea. If I understand correctly, the document strategy will also store all other top-level fields in both before and after snapshot? For v3 compatibility, we still want the "delta" strategy behavior for most simple top level fields e.g. `ownerGroup`. Only for object valued fields (of which PSI is only using `datasetlifecycle` in the history), we want the "document" strategy behavior.. (this is not even 1-1 compatibility with v3 behavior, but an over-approximation: in v3, the history.after just blindly recorded all fields present in the PATCH request).
Another side note, maybe you've already noticed: in DB we still store pure "deltas" also for `datasetlifecycle`, only in the API response we construct this backward compatible view of history. But you raise a good point: that these converters should not be called when the tracking strategy is "document", so I will implement a fix for that soon.

Junjiequan:
Aha, thanks for the good explanation. It seems `TRACKABLE_STRATEGY` is one for all, thus cannot be targeted to the specific schema or certain field. If there's a strong demand, maybe in the future (when we deprecates dataset v3 ) we can extend the logic to assign different strategy by schema or fields.

> that these converters should not be called when the tracking strategy is "document", so I will implement a fix for that soon.

At ESS we relies purely on the new history plugin system with new history endpoint, so I don't have a strong opnion on this

src/datasets/utils/history.util.ts
```diff
@@ -71,12 +71,32 @@ export function convertGenericHistoryToObsoleteHistory(
...
-      };
+      const beforePartial = (history.before[field] || {}) as Record<
+        string,
+        unknown
+      >;
+      const afterPartial = (history.after?.[field] || {}) as Record<
+        string,
+        unknown
+      >;
+      const reconstructedBefore = JSON.parse(
```
minottic: is this a clone? Maybe a cloneDeep from lodash?

minottic: would it be easier to build `reconstructedBefore` as a filtered version of datasetlifecycle?

```TS
import { omit } from 'lodash';

reconstructedBefore = omit(datasetSnapshot.datasetlifecycle, Object.keys(afterPartial).filter(key => !(key in beforePartial))

```

src/datasets/utils/history.util.ts
```diff
@@ -71,12 +71,32 @@ export function convertGenericHistoryToObsoleteHistory(
...
+        }
+      }
+      // apply before to construct previous snapshot
+      Object.assign(reconstructedBefore, beforePartial);
+
+      history.before[field] = reconstructedBefore;
+
       history.after![field] = JSON.parse(
-        JSON.stringify(history.after![field] as Record<string, unknown>),
+        JSON.stringify(datasetSnapshot.datasetlifecycle),
```
minottic, line 102: cloneDeep?

src/datasets/utils/history.util.ts
minottic: (or a one liner): 

```TS
reconstructedBefore = {
  ...omit(datasetSnapshot.datasetlifecycle, Object.keys(afterPartial).filter(key => !(key in beforePartial)),
  ...beforePartial
}
```

src/datasets/datasets.controller.ts
```diff
@@ -618,7 +618,7 @@ export class DatasetsController {
         subsystem: "Dataset",
       }),
       currentDataset,
-    );
+    ).reverse();
```
minottic, line 621: if you reverse here, why do we need the FE reverse as well?

minottic (approved):
since we clarified the double reverse offline, the other comments are cosmetic changes, so up to you

src/datasets/utils/history.util.ts
omkar-ethz: as discussed offline, `cloneDeep` caused the class-transformers `plainToInstance` in the controller ([here](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/datasets.controller.ts#L608)) to error. From their [documentation](https://github.com/typestack/class-transformer?tab=readme-ov-file#what-is-class-transformer), it seems that it only handles plain JS objects, whereas with cloneDeep we get a LifecycleClass class object.

src/datasets/utils/history.util.ts
omkar-ethz: thanks! this is more succint, changed.

omkar-ethz:
> since we clarified the double reverse offline, the other comments are cosmetic changes, so up to you

thanks for the review! could you also have a look at the FE [PR](https://github.com/SciCatProject/frontend/pull/2183) :)

src/datasets/utils/history.util.ts
```diff
@@ -71,12 +72,28 @@ export function convertGenericHistoryToObsoleteHistory(
...
+      >;
+      const afterPartial = (history.after?.[field] || {}) as Record<
+        string,
+        unknown
+      >;
+      // omit keys that are only in afterPartial, as it means they were added in this update
+      // and apply beforePartial to the resulting datasetlifecycle snapshot
+      const reconstructedBefore = {
+        ...omit(
+          JSON.parse(JSON.stringify(datasetSnapshot.datasetlifecycle)),
```
minottic: do you  need the JSON.parse(...) here now? the omit should return a copy

minottic: if you need it for the same reason as cloneDeep, it is slightly more efficient to do after the omit :)

omkar-ethz: yes it is still needed to convert it from LifecycleClass to plain / object literal. Otherwise we get the same error as cloneDeep in the controller. Without it, omit returns a copy but of the LifecycleClass object

omkar-ethz: > if you need it for the same reason as cloneDeep, it is slightly more efficient to do after the omit :)

good point, changed:)

PR #2484, opened 2026-01-21
feat(publisheddata): automatically add issued date to datacite record

fpotier:
## Description

Automatically adds a Date Issued when a DOI is registered to DataCite
## Motivation

We need to add this entry to our future records to comply with a swiss metadata standard
In this PR it's not configurable and only occurs if it doesn't break validation.
Maybe it should be made opt-in/out?
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/published-data/published-data.v4.controller.ts
+++ b/src/published-data/published-data.v4.controller.ts
@@ -26,7 +26,7 @@ import {
 } from "@nestjs/swagger";
 import { Request } from "express";
 import { Validator } from "jsonschema";
-import { isArray } from "lodash";
+import { cloneDeep, isArray } from "lodash";
 import { FilterQuery, QueryOptions } from "mongoose";
 import { firstValueFrom } from "rxjs";
 import { AttachmentsService } from "src/attachments/attachments.service";
@@ -576,24 +576,8 @@ export class PublishedDataV4Controller {
     publishedData.registeredTime = data.registeredTime;
     publishedData.status = data.status;
 
-    const issuedDate = {
-      date: data.registeredTime.toISOString(),
-      dateType: "Issued",
-    };
-
-    if (!publishedData.metadata) publishedData.metadata = {};
-    publishedData.metadata.dates = publishedData.metadata.dates ?? [];
-
-    if (
-      isArray(publishedData.metadata?.dates) &&
-      !publishedData.metadata.dates.some(
-        (d) => d.hasOwnProperty("dateType") && d.dateType === "Issued",
-      )
-    ) {
-      publishedData.metadata.dates.push(issuedDate);
-    }
-
     await this.validateMetadata(publishedData.metadata);
+    await this.tryAddDateIssued(publishedData, publishedData.registeredTime);
 
     await Promise.all(
       publishedData.datasetPids.map(async (pid) => {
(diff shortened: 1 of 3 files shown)


src/published-data/published-data.v4.controller.ts
```diff
@@ -783,4 +789,34 @@ export class PublishedDataV4Controller {
...
+    const originalMetadata = cloneDeep(publishedData.metadata);
+
+    try {
+      if (!publishedData.metadata) publishedData.metadata = {};
+      publishedData.metadata.dates = publishedData.metadata.dates ?? [];
+
+      if (
+        isArray(publishedData.metadata.dates) &&
+        !publishedData.metadata.dates.some(
+          (d) => d.hasOwnProperty("dateType") && d.dateType === "Issued",
```
minottic: is this required: `d.hasOwnProperty("dateType")` ?

src/published-data/published-data.v4.controller.ts
```diff
@@ -783,4 +789,31 @@ export class PublishedDataV4Controller {
...
+        )
+      ) {
+        publishedData.metadata.dates.push({
+          date: registeredTime.toISOString(),
+          dateType: "Issued",
+        });
+      }
+      await this.validateMetadata(publishedData.metadata);
+    } catch (_validationError) {
```
minottic: you probably don't need this var?

test/jest-e2e-tests/publishedData.e2e-spec.ts
```diff
@@ -86,6 +86,14 @@ describe.each([undefined, "", "https://api.test.datacite.org/dois"])(
         .then((res) => {
           expect(res.body.status).toEqual("registered");
           expect(res.body.registeredTime).toBeDefined();
+          expect(res.body.metadata.dates).toBeDefined();
+          expect(Array.isArray(res.body.metadata.dates)).toBe(true);
+          expect(res.body.metadata.dates.length).toBeGreaterThan(0);
```
minottic: can this test the exact count?

minottic (commented):
minor comments

src/published-data/published-data.v4.controller.ts
fpotier: correct, but TS doesn't allow an empty `catch()`, so I changed the eslint rules to allow ignoring the caught exception

minottic: don't you want to log the error?

fpotier: could be an option but I thought it would be annoying to have logs about this if you know from the get-go that your schema doesn't support it.
but maybe it's still better to log it

minottic (approved):
thanks! good point about the comment, up to you!

Junjiequan:
Please wait for merging until we review it

nitrosx (commented):
Good idea, but I am not sure that this is the right solution.
With this solution, we imply that the site admin or the facility data team wants to have this specific field "issued" in the metadata published on DataCite. We cannot make this assumption.
Also, it breaks away from the previous published data configuration which is managed in configuration, where we allow the site admin to configure the metadata submitted to DataCite.

I think that the Core team should do some brainstorming and discuss more the topic.

nitrosx:
I think we should transfer this feature in this configuration file: [publishedDataConfig.example.json](https://github.com/SciCatProject/scicat-backend-next/blob/master/publishedDataConfig.example.json)

fpotier:
> Good idea, but I am not sure that this is the right solution. With this solution, we imply that the site admin or the facility data team wants to have this specific field "issued" in the metadata published on DataCite. We cannot make this assumption. Also, it breaks away from the previous published data configuration which is managed in configuration, where we allow the site admin to configure the metadata submitted to DataCite.
> 
> I think that the Core team should do some brainstorming and discuss more the topic.

Thanks for the feedback, so you would suggest for now to add something like this to the config file?
```json
"options": {
  "addDateIssued": true
}
```

I am currently working on making PSI's Scicat conformant with the swiss DataCite metadata standard and facing some limitations with the way it's implemented currently so I'd be interested to participate in future discussions

nitrosx:
I see that as an acceptable solution, although not really generic!!!
I think the `Issues` field should be included in the published data configuration together with the fact that you want it populated with a specific date.
Hopefully it makes sense!!!

PR #2487, opened 2026-01-22
fix: return fields intersection in fullquery

minottic:
## Description
Full query used to do the union of permission fields. This PR fixes it to be an intersection, meaning that the where condition with multiple fields is changed from a list of ORs to a list of ANDs. This is a bug particularly in the datasets overview web page when using filters 

## Tests included

- [x] Passing? 

## Documentation


Junjiequan (approved):
nice catch, looks good to me

Junjiequan:
Maybe we should include a comment there to differentiate the userGroup from regular filters

PR #2491, opened 2026-01-26
feat: support v3 fields in publisheddata map filter

minottic:
## Description
Allows filtering with v3 returned fields. This converts fields in filter to DB fields which are then consumed by mongoose

## Tests included

- [ ] Included for each change/fix?
- [ ] Passing? 

## Documentation
- [ ] swagger documentation updated (required for API changes)
- [ ] official documentation updated

PR #2492, opened 2026-01-26
chore: move mapping common logic to shared function

minottic:
## Description
Reduce duplication by reusing the same mapping utility function


minottic:
opened against #2491

nitrosx (approved):

PR #2493, opened 2026-01-26
chore: remove mixup of filter and fields in v3 publish

minottic:
## Description
This PR simplifies the publishedData controllers by removing the overlap between fields and filters and delegating all to where, fields and order inside `filter`, as done by other endpoints.

## Tests included

- [ ] Included for each change/fix?
- [ ] Passing? 

## Documentation
- [ ] swagger documentation updated (required for API changes)
- [ ] official documentation updated

PR #2502, opened 2026-01-26
feat: add ability to exclude fields

minottic:
## Description
Queries with `fields: ["-field1"]` enable excluding specific fields on demand. This is useful especially with publishedData to exclude thumbnail which slow down GET all. It's open agains #2493 

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/common/pipes/filter.pipe.ts
+++ b/src/common/pipes/filter.pipe.ts
@@ -1,6 +1,7 @@
 import { Injectable, PipeTransform } from "@nestjs/common";
 import { IFilters } from "../interfaces/common.interface";
-import { get, isEmpty, isPlainObject, keys, pickBy } from "lodash";
+import { get, isEmpty, isPlainObject, mapKeys } from "lodash";
+import { normalizeProjection } from "../utils";
 
 type KeyMap = Record<string, string>;
 
@@ -170,10 +171,10 @@ export class FieldsPipe<T = unknown> extends FilterPipeAbstract<T> {
 
   applyTransform(value: unknown) {
     if (this.allowObjectFields && isPlainObject(value)) {
-      const activeKeys = keys(
-        pickBy(value as Record<string, boolean>, Boolean),
+      const fields = mapKeys(value as object, (_, key) =>
+        get(this.apiToDBMap, key, key),
       );
-      return activeKeys.map((key) => get(this.apiToDBMap, key, key));
+      return normalizeProjection(fields);
     }
     if (Array.isArray(value))
       return value.map((key) => get(this.apiToDBMap, key, key));
--- a/test/DatasetV4.js
+++ b/test/DatasetV4.js
@@ -826,6 +826,26 @@ describe("2500: Datasets v4 tests", () => {
           firstDataset.should.not.have.property("datablocks");
         });
     });
+
+    it("0212: should fetch specific dataset fields excluding fields provided in field", async () => {
+      const filter = {
+        fields: {datasetName: 0, contactEmail: 1},
+      };
+
+      return request(appUrl)
+        .get(`/api/v4/datasets`)
+        .query({ filter: JSON.stringify(filter) })
+        .auth(accessTokenAdminIngestor, { type: "bearer" })
+        .expect(TestData.SuccessfulGetStatusCode)
+        .expect("Content-Type", /json/)
+        .then((res) => {
+          res.body.should.be.a("array");
+          const [firstDataset] = res.body;
+
+          firstDataset.should.not.have.property("datasetName");
+          firstDataset.should.have.property("contactEmail");
+        });
+    });
   });
 
   describe("Datasets v4 findOne tests", () => {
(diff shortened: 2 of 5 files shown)


src/common/pipes/filter.pipe.ts
```diff
@@ -168,15 +168,24 @@ export class FieldsPipe<T = unknown> extends FilterPipeAbstract<T> {
...
-      return activeKeys.map((key) => get(this.apiToDBMap, key, key));
     }
     if (Array.isArray(value))
-      return value.map((key) => get(this.apiToDBMap, key, key));
+      return value.map((k) =>
+        this.resolveNegative(
+          k.startsWith("-") ? k.slice(1) : k,
+          k.startsWith("-"),
+        ),
+      );
```
Junjiequan: `-` symbol to exclude fields within a query is not so intuitive to me. I think this should be more explicit, an additional `exclude` param could be better? e.g, `{fields: xxx, exclude: xxx}`

test/DatasetV4.js
```diff
@@ -826,6 +826,27 @@ describe("2500: Datasets v4 tests", () => {
...
+        .then((res) => {
+          res.body.should.be.a("array");
+          const [firstDataset] = res.body;
+
+          firstDataset.should.not.have.property("datasetName");
+          firstDataset.should.have.property("pid");
+        });
+    });
+
   });
```
Junjiequan, line 849: Question here, 
dataset v4 findAll is using projection, do we really need to use `-` or other customized logic to exclude fields

src/common/pipes/filter.pipe.ts
nitrosx: I agree with @Junjiequan.
Using the character `-` in front of the field names is not intuitive and can lead to wrong interpretations.
I would lean more towards a solution where we have `fields` or `include` to actively select the fields that we want to return and `exclude` to list the fields that we do not want to be returned.
We could also adopt the mongodb projection syntax... `fields` argument can be a list of fields to be returned (as it is now) or it can be an object where the key is the field name and the value indicates if it needs to be returned (1) or not (0). fields that are not listed are automatically excluded.
If such functionality is not enough and you want to just list the fields to be excluded, I would strongly suggest solution number one: `fields` or `include` to select fields to be included, `exclude` to list fields that should not be returned

test/DatasetV4.js
nitrosx: same question from me.
Please see my previous comment.

nitrosx (changes requested):

nitrosx:
Here is the relative documentation of MongoSB aggregation project functionality: https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/

src/common/pipes/filter.pipe.ts
minottic: mongoose uses `-` for [fields exclusion](https://mongoosejs.com/docs/api/query.html#Query.prototype.select()) when using `.select`, that's why I leaned towards it. If fields was an object, I could easily see a way to be more explicit, i.e. `{field1: true, field2: false}` (and this was the behaviour in loopback3, and covered by the filter pipe transformer whenever fields is an object [here](https://github.com/SciCatProject/scicat-backend-next/blob/225630b8bbea039b21277a82ed4841a9cae9d247/src/common/pipes/filter.pipe.ts#L177-L179)  and [here](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/common/utils.ts#L416-L426)). I had to resort to prepending the minus because fields was chosen to be an array (not sure why we chose it to be an array - see [here](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/types/dataset-filter-content.ts#L42-L47) and [here](https://github.com/SciCatProject/scicat-backend-next/blob/225630b8bbea039b21277a82ed4841a9cae9d247/test/PublishedData.js#L233-L243)).

I would be happy to convert fields into an object (or extend it to support objects and lists - which will be easy, thanks to the filter pipe), but I think having both `fields` and `exclude` is rather counter intuitive for the users (since `fields` does not really semantically equal `include`)

minottic: > We could also adopt the mongodb projection syntax... fields argument can be a list of fields to be returned (as it is now) or it can be an object where the key is the field name and the value indicates if it needs to be returned (1) or not (0). fields that are not listed are automatically excluded.

this would work for me, and it's rather easy to implement, and it's actually already supported in v3 endpoints, [here](https://github.com/SciCatProject/scicat-backend-next/blob/225630b8bbea039b21277a82ed4841a9cae9d247/src/common/pipes/filter.pipe.ts#L177-L179) and [here](https://github.com/SciCatProject/scicat-backend-next/blob/225630b8bbea039b21277a82ed4841a9cae9d247/test/PublishedData.js#L233-L243)

Junjiequan: Object fields with 1/true or 0/false sounds good to me rater than just an array. If it is easier to implement I suggest we follow that pattern, and its more intuitive IMO

nitrosx (approved):

PR #2503, opened 2026-01-26
feat: support filter in header in publishData

minottic:
## Description
Accept filter in header in publishedData GET all

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/published-data/published-data.controller.ts
+++ b/src/published-data/published-data.controller.ts
@@ -67,6 +67,7 @@ import {
 import { PublishedDataService } from "./published-data.service";
 import { PublishedData } from "./schemas/published-data.schema";
 import { V3_FILTER_PIPE } from "./pipes/filter.pipe";
+import { Filter } from "src/datasets/decorators/filter.decorator";
 
 @ApiBearerAuth()
 @ApiTags("published data")
@@ -280,7 +281,7 @@ export class PublishedDataController {
     excludeExtraneousValues: true,
   })
   async findAll(
-    @Query(...V3_FILTER_PIPE, RegisteredFilterPipe)
+    @Filter(...V3_FILTER_PIPE, RegisteredFilterPipe)
     filter?: {
       filter: IPublishedDataFilters;
     },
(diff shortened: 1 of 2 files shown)


src/published-data/published-data.controller.ts
fpotier: shouldn't this be also removed from the api specs?

minottic: it's removed [here](https://github.com/SciCatProject/scicat-backend-next/pull/2493)

fpotier: ah mb

sbliven:
Putting it in the query seems more natural to me. Can we accept both query and header styles?

minottic:
> Putting it in the query seems more natural to me. Can we accept both query and header styles?

indeed. this PR extends filter to be passed either as query or as header (see test)

nitrosx:
@minottic can you provide more background on your use case , so I can understand the reason for accepting filters in header?
Thanks

minottic:
it's because it was supported by loopback 3 and thus it's a backward compatible feature. Especially it's used by the landingpage

nitrosx:
Thanks. Got it.
Would it be more productive to update the landing page?

minottic:
> Thanks. Got it.
> Would it be more productive to update the landing page?

dunno, this code change is rather small and well maintainable, since it's simply changing the decorator to a more generic one (that btw is already used in datasetv3). I have to admit I am always a bit lost where to do the change, as I never really understood to what level we want to be backward compatible. If we strive for backward compatibility, then  this change clearly goes here, if not we can choose

Junjiequan:
> dunno, this code change is rather small and well maintainable, since it's simply changing the decorator to a more generic one (that btw is already used in datasetv3). I have to admit I am always a bit lost where to do the change, as I never really understood to what level we want to be backward compatible. If we strive for backward compatibility, then this change clearly goes here, if not we can choose

We may need to set clear boundries for supporting backward compatibility. if in the feature we need to support backward compatibility for v3 and v4... altogether, it will be impossible or very expensive to maintain

minottic:
imo backward compatibility it's only supposed to be in v3 endpoints (as a client will call the same endpoint, so it should not expect a different response/payload). Once they call v4 there should be no expectation of backward compatibility

This PR only adds the filter to the v3 endpoint, as it was supported before

nitrosx (approved):

PR #2507, opened 2026-01-29
fix: exclude arrays from returning filter object

minottic:
## Description
The IF condition wrongly returned an object with the filter key when the input was an array as the array has the "filter" method. The filter object should only be returned when the Pipe is applied on an object with the filter key. The IF is tightened in this PR 

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/common/pipes/filter.pipe.ts
+++ b/src/common/pipes/filter.pipe.ts
@@ -80,6 +80,12 @@ export abstract class FilterPipeAbstract<T = unknown> implements PipeTransform<
     return obj;
   }
 
+  private hasFilterDataProperty(
+    obj: Record<string, unknown>,
+  ): obj is { filter: unknown } {
+    return Object.hasOwn(obj, "filter") && typeof obj.filter !== "function";
+  }
+
   transform(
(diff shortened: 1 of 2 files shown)


fpotier (approved):

src/common/pipes/filter.pipe.ts
```diff
@@ -92,9 +98,9 @@ export abstract class FilterPipeAbstract<T = unknown> implements PipeTransform<
     const parsedFilter = FilterPipeAbstract.transformDeep(
       FilterPipeAbstract.parseJson(inValue as string),
       { valueFn: (val) => FilterPipeAbstract.parseJson(val as string) },
-    ) as object;
+    ) as Record<string, unknown>;
```
sbliven, line 101: Are you sure this is typesafe? If you were getting cases where parsedFilter.filter is a function, that implies that this sometimes returns an Array rather than a record.

minottic: you are right, that can indeed be an array as well, my bad

PR #2558, opened 2026-02-24
feat: rest api provider for access groups

abdellah257:
## Description
a REST API Provider to fetch access groups

## Motivation
additional way to get access groups (used at ILL) using a rest api

## Changes:
- New provider at auth/access-group-provider
- Changes to configuration.ts to take into account REST env variables
- `env.example` updated for a env variables usage example
```
ACCESS_GROUPS_REST_ENABLED=true
ACCESS_GROUPS_SERVICE_REST_AUTH_KEY="Authorization"
ACCESS_GROUPS_SERVICE_REST_AUTH_VALUE=""
ACCESS_GROUPS_SERVICE_REST_API_URL=""
ACCESS_GROUPS_SERVICE_REST_USER_ID_FIELD=""  # the field to be used from UserPayload
```

## Tests included

- `access-group-from-rest-api-call.service.spec.ts`


src/auth/access-group-provider/access-group-from-rest-api-call.service.ts
```diff
@@ -0,0 +1,75 @@
...
+import { HttpService } from "@nestjs/axios";
+import { firstValueFrom } from "rxjs";
+import { get } from "lodash";
+import { AxiosError } from "axios";
+/**
+ * This service is used to fetch access groups from a REST API.
+ */
+@Injectable()
+export class AccessGroupFromRestApiService extends AccessGroupService {
+  private readonly logger = new Logger(AccessGroupFromRestApiService.name);
```
Junjiequan: Can we use Logger imported from `import { Logger } from "@nestjs/common";` for the code consistency.

Which can be then used like this
```js
     Logger.log(
        accessGroups,
        "AccessGroupFromGraphQLApiService:",
      );
```

Junjiequan (dismissed):

Junjiequan (approved):
lgtm, but please double check package-lock.json before you merge this

PR #2562, opened 2026-02-24
BREAKING CHANGE: migrate elasticsearch to opensearch

Junjiequan:
## Description
THis PR migrates the search integration from Elasticsearch to OpenSearch.

### Dependencies

- Removed @elastic/elasticsearch and @nestjs/elasticsearch
- Added @opensearch-project/opensearch ^3.5.1 JavaScript client compatible with opensearchproject/opensearch:3.5.0

### Code Changes
- Renamed all references from Elasticsearch to OpenSearch

### Configuration
- Removed hardcoded index settings and mappings. Configuration is now loaded from opensearchConfig.json if present; otherwise defaults from opensearchConfig.example.json are used

### Search Integration
- Added opensearchQuery and opensearchFacet, used when OpenSearch is enabled and contains data
- New OpenSearch integration supports only $text queries, performing partial search on description and datasetName
- Restricted OpenSearch endpoints to admin users only. Search functionality is handled internally through the OpenSearch service

### Data Synchronization
- Dataset synchronization from MongoDB to OpenSearch excludes scientificMetadata, history, and datasetlifecycle

## Notes:

- opensearchQuery extracts the $text query, searches OpenSearch, retrieves dataset IDs up to the provided limit (default: 1000), and then applies additional filtering using MongoDB queries

- opensearchFacet extracts the $text query, searches OpenSearch, and retrieves results up to the index setting max_result_window. The example configuration sets this to 2000000, which may impact performance depending on server capacity and dataset volume

If performance issues arise with large result sets, alternative approaches such as pagination, scroll queries, or other optimizations can be considered as future improvements

## Tests included

- [ ] Passing? 

## Documentation


src/opensearch/interfaces/mappingInterface.type.ts
```diff
@@ -0,0 +1,5 @@
+export type MappingProperty = Record<string, unknown>;
+
+export interface MappingObject {
+  [key: string]: MappingProperty;
+}
```
Junjiequan: Remove

src/opensearch/dto/create-index.dto.ts
```diff
@@ -0,0 +1,20 @@
...
+    example: {
+      properties: {
+        datasetName: { type: "text" },
+        ownerGroup: { type: "keyword" },
+      },
+    },
+  })
+  @IsObject()
+  @IsOptional()
+  mappings: Partial<TypeMapping>;
```
Junjiequan: Partial is not needed

src/opensearch/providers/fields.enum.ts
```diff
@@ -0,0 +1,8 @@
+export enum MustFields {
+  DatasetName = "datasetName",
+  Description = "description",
+}
+
+export enum ShouldFields {
```
Junjiequan: Change to FilterFields

src/opensearch/providers/query-builder.service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  public buildSearchQuery(filter: ISearchFilter) {
+    try {
+      const accessFilter = this.buildFilterFields(filter);
+      const textQuery = this.buildTextQuery(filter);
+
+      // NOTE: The final query flow is as follows:
+      // step 1. Build filter fields conditions must match all filter fields
+      // step 2. Build should fields conditions must match at least one should field
+      // step 3. Build text query conditions must match all text query fields
```
Junjiequan: Update or remove comments

.env.example
```diff
@@ -84,16 +84,17 @@ MS365_CLIENT_ID=<clientId>
...
-ES_FIELDS_LIMIT=400000
-ES_INDEX="dataset"
-ES_REFRESH=<"wait_for"|"false">
+OPENSEARCH_ENABLED=<"yes"|"no">
+OPENSEARCH_ENABLED="https://localhost:9200"
+OPENSEARCH_ENABLED="admin"
+OPENSEARCH_ENABLED="Scicat-password2026"
+OPENSEARCH_DEFAULT_INDEX="dataset"
+# use wait_for for testing and development to ensure that the index is updated before the next operation,
+# use false for production to improve performance by not waiting for the index to refresh before returning a
```
Junjiequan: provide better explanation

test/OpenSearch.js
```diff
@@ -0,0 +1,61 @@
...
+let accessTokenAdminIngestor = null,
+  accessTokenArchiveManager = null,
+  pid = null;
+
+const isESenabled = process.env.ELASTICSEARCH_ENABLED == "yes";
+
+(isESenabled ? describe : describe.skip)(
+  "ElastiSearch: CRUD, filtering and search test case",
+  () => {
+    before(async () => {
```
Junjiequan: change  ES to OS

package.json
```diff
@@ -30,18 +30,16 @@
     "test:api": "npm run test:api:jest --maxWorkers=50% && concurrently -k -s first \"wait-on http://localhost:3000/explorer/ && npm run test:api:mocha\" \"npm run start:test\"",
     "test:api:jest": "dotenv -o -e test/config/.env -e test/config/.env.override -- jest --config ./test/config/jest-e2e.json --maxWorkers=50%",
     "test:api:mocha": "dotenv -o -e test/config/.env -e test/config/.env.override -- mocha --config ./test/config/.mocharc.js -r chai/register-should.js",
-    "prepare:local": "docker-compose -f CI/E2E/docker-compose-local.yaml --env-file CI/E2E/.env.elastic-search up  -d && cp test/config/functionalAccounts.json functionalAccounts.json && cp proposalTypes.example.json proposalTypes.json && cp publishedDataConfig.example.json publishedDataConfig.json"
+    "prepare:local": "docker-compose -f CI/E2E/docker-compose-local.yaml --env-file CI/E2E/.env.open-search up  -d && cp test/config/functionalAccounts.json functionalAccounts.json && cp proposalTypes.example.json proposalTypes.json && cp publishedDataConfig.example.json publishedDataConfig.json"
```
Junjiequan: Just hard code values and remove it from ENV for local development

src/elastic-search/configuration/indexSetting.ts
```diff
@@ -1,27 +1,20 @@
-import {
-  AnalysisEdgeNGramTokenizer,
-  AnalysisPatternReplaceCharFilter,
-  IndicesIndexSettingsAnalysis,
-  MappingDynamicTemplate,
-} from "@elastic/elasticsearch/lib/api/types";
-
 //Tokenizers
-export const autocomplete_tokenizer: AnalysisEdgeNGramTokenizer = {
+export const autocomplete_tokenizer: any = {
```
emigun: I'm confused by the diff. It says a bunch of files are changed in folder `elastic-search` but the folder doesn't exist anymore when checking out this branch. When I search for lines like these, they don't exist anywhere either 🤔

Junjiequan: The folder is now renamed to `opensearch`

emigun: Yes I know. But the diff was incorrect. Now it says outdated and the problem is gone...

.github/workflows/test.yml
```diff
@@ -130,14 +130,10 @@ jobs:
 
       - name: API tests
         env:
-          ELASTICSEARCH_ENABLED: ${{ matrix.elasticsearch_enabled }}
-          COMPOSE_PROFILES: ${{ matrix.elasticsearch_enabled == 'yes' && 'elasticsearch' || ''}}
-          CLUSTER_NAME: es-cluster
-          ES_PASSWORD: duo-password
-          MEM_LIMIT: 4G
-          STACK_VERSION: 8.8.2
+          COMPOSE_PROFILES: ${{ matrix.opensearch_enabled == 'yes' && 'opensearch' || ''}}
```
minottic, line 133: will the additonal open-search settings from the env file be used here?

src/config/configuration.ts
```diff
@@ -420,6 +424,7 @@ const configuration = () => {
     frontendConfig: jsonConfigMap.frontendConfig,
     frontendTheme: jsonConfigMap.frontendTheme,
     publishedDataConfig: jsonConfigMap.publishedDataConfig,
+    opensearchConfig: jsonConfigMap.opensearchConfig,
```
minottic, line 437: what's this? and why are EVNs not enough?

src/datasets/datasets.controller.ts
```diff
@@ -985,12 +990,22 @@ export class DatasetsController {
...
 
-    const parsedFilters: IFilters<DatasetDocument, IDatasetFields> = {
-      fields: fields,
-      limits: JSON.parse(filters.limits ?? "{}"),
-    };
+    let datasets: DatasetDocument[] | null;
 
-    const datasets = await this.datasetsService.fullquery(parsedFilters);
+    const osEnabled =
+      this.configService.get<string>("opensearch.enabled") || "no";
```
minottic: better in the constructor to avoid multiple loading?

src/datasets/datasets.service.ts
```diff
@@ -186,8 +189,10 @@ export class DatasetsService {
 
     const savedDataset = await createdDataset.save();
 
-    if (this.ESClient && createdDataset) {
-      await this.ESClient.updateInsertDocument(savedDataset.toObject());
+    if (this.opensearchService && createdDataset) {
+      await this.opensearchService.updateInsertDocument(
+        sanitizeDatasetForOpensearch<DatasetDocument>(savedDataset.toObject()),
```
minottic: I think openSearch should have a DTO and we should use plainToInstance as usual

src/datasets/datasets.service.ts
```diff
@@ -292,63 +295,112 @@ export class DatasetsService {
...
+    delete mongoQuery.$text;
+
+    const osResult = await this.opensearchService.search(
+      { text, userGroups, isAdmin: isAdmin },
+      defaultOsIndex,
+      modifiers.limit,
+      modifiers.skip,
+    );
+    const datasets = await this.datasetModel
+      .find({ pid: { $in: osResult.data }, ...mongoQuery })
```
minottic, line 351: I think this will fail if the pid list from osResult is too long

src/datasets/datasets.service.ts
```diff
@@ -292,63 +295,112 @@ export class DatasetsService {
...
       );
 
-      datasets = await this.datasetModel
-        .find({ pid: { $in: esResult.data } })
-        .sort(modifiers.sort)
-        .exec();
-    }
+    const modifiers: QueryOptions = parseLimitFilters(filter.limits);
+
+    delete mongoQuery.$text;
```
minottic, line 342: I think there's a bit of duplication here with the IF clauses in the controller. Isn't it enough to just surround lines 329 to 337 and build the $in conditionally when opensearch is active?

src/datasets/datasets.service.ts
```diff
@@ -292,63 +295,112 @@ export class DatasetsService {
...
+    );
+
+    fields.openSearchIdList = osResult.data;
+    delete fields.text;
+    const pipeline = createFullfacetPipeline<
+      DatasetDocument,
+      IDatasetOpenSearchPipeline
+    >(this.datasetModel, "pid", fields, facets, "");
+
+    return await this.datasetModel.aggregate(pipeline).exec();
```
minottic, line 415: same comment about duplication

src/opensearch/opensearch.service.ts
```diff
@@ -0,0 +1,384 @@
...
+    }
+    if (!this.osConfigs) {
+      Logger.warn(
+        `Missing Opensearch index configuration, using default settings and mappings`,
+        "Opensearch",
+      );
+    }
+  }
+
+  async onModuleInit() {
```
minottic: what happens if the connection to os is lost after the init? shouldn't this be recovered?

src/datasets/datasets.service.ts
```diff
@@ -599,9 +640,20 @@ export class DatasetsService {
...
 
-  async isElasticSearchDBEmpty() {
-    if (!this.ESClient) return;
-    const count = await this.ESClient.getCount();
-    return count.count > 0;
+  async syncDatasetsToOpensearch(index: string) {
+    try {
+      await this.opensearchService.checkIndexExists(index);
+
+      const datasets = await this.datasetModel
```
minottic: this will retrive all datasets, I don't think that's feasible in memory. It should be batched

src/opensearch/opensearch.service.ts
```diff
@@ -0,0 +1,384 @@
...
+    const { body: indexExists } = await this.osClient.indices.exists({ index });
+
+    if (!indexExists) {
+      throw new Error(`Index ${index} not found`);
+    }
+  }
+
+  // *** NOTE: below are helper methods ***
+
+  async performBulkOperation<T extends { _id: unknown }>(
```
minottic, line 360: shouldn't this be batched?

minottic (commented):
thanks a lot! I am not sure I see where this is used in the v4 datasets, isn't it only applied to v3 datasets? 

I also don't fully understand the pletora of docker composes, why so many?

src/config/configuration.ts
Junjiequan: This is used to to control index settings and mappings, which I explained in the documentation  that is just uploaded https://github.com/SciCatProject/backend/pull/2562/changes#diff-ad12ea0a128599f100b5d5a48511e8aab308c0b7ea85da20c7e5744d47de48f4

src/datasets/datasets.service.ts
Junjiequan: I was thinking of using a const to have a single place to manage excluded fields for both update operations and MongoDB queries during data sync:
```
const DATASET_OPENSEARCH_EXCLUDED_FIELDS = [
  "scientificMetadata",
  "history",
  "datasetlifecycle",
] as const;
```

But maybe seperate them into DTO and mongo query variable would be easier to read and maintain?

src/datasets/datasets.service.ts
Junjiequan: I don't quite get it, could you give a snippet?

src/datasets/datasets.service.ts
Junjiequan: In practice this shouldn’t crash, frontend pagination is capped at 100 per page, and after access filtering and text search the result set is usually small. There's two edge cases, one is admin users, since they bypass accessGroup based filtering and may hit larger result sets. This is already documented as a known limitation (if it really becomes an issue we can discuss about performance optimization for this). You can also limit the maximum number of returned results via index settings. The other edge case is user set very large limit and call fullquery endpoint directly, to avoid such case, `max_result_window`  needs to be set when creating indexes https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/index-settings/

The real issue however is the fullfacet endpoint, where counts are calculated across all documents. I have thought about few potential approaches, such as returning approximate counts or showing filter options without counts, but these would require quite big refactoring

src/datasets/datasets.service.ts
Junjiequan: Good point, I will include batch here

src/opensearch/opensearch.service.ts
Junjiequan: After 3 init fails attempts, BE will just throw error and fallback to mongo query, so restart is required.
Are you suggesting to run health check every n seconds and automatically re-connect if OS is available?

src/opensearch/opensearch.service.ts
Junjiequan: the bulk API has its own internal batch logic
https://docs.opensearch.org/latest/clients/javascript/helpers/?utm_source=chatgpt.com

Junjiequan:
@minottic 
Opensearch was previously used by datasets v3 only so this PR only included changes for datasets v3. For V4 integration with opensearch I think its better to have new PR for that, I think this PR is too big already.
Docker composes needs to be cleaned up. I know `docker-compose.api.yaml` and `docker-compose.local.yaml` is in use but not sure about the others

.github/workflows/test.yml
Junjiequan: No env is required for github jobs and local development, the values are hard coded

minottic:
> Opensearch was previously used by datasets v3 only so this PR only included changes for datasets v3

ah ok, thanks. I was a bit puzzled because the tests are modified for the v4 controllers, but I did not see changes in the code that made me think opensearch was applied to v4. I agree v4 changes can come later if so

src/datasets/datasets.service.ts
minottic: maybe an in-between is using [Pick](https://docs.nestjs.com/openapi/mapped-types#pick) (starting from one of the Datasets' DTO) with the list from a const?

src/datasets/datasets.service.ts
minottic: I think the users submitting no limits it's still a thread for fullquery, even if rare. Wouldn't it be an option to chunk the find() using batches of returned pids and then concat? (for the mongo part, the os part, maybe we need indeed to limit the size to a max of N)

minottic: pagination from os would also solve the fullfacet problem if I am not mistaken? I mean here that if we retrieve the OS documents in batches then we can feed these batches to mongo and concat. I have the feeling, as before, that the batch size of OS will be bigger than the size limit of the list of pids

src/datasets/datasets.service.ts
minottic: sorry.

Essentially in the controller you do some IFs that you then duplicate to some extent here. I would see working, removing the IF in the controller and simply calling `opensearchQuery` having added to line 313:

```TS
|| (this.configService.get<string>("opensearch.enabled") || "no")
|| !filters.fields?.text
```

src/datasets/datasets.service.ts
minottic: see comment in fullquery duplication

src/opensearch/opensearch.service.ts
minottic: no, I meant that maybe we should think to test connection and eventually re-establish when opensearch operations are triggered

test/DatasetV4.js
```diff
@@ -314,7 +314,7 @@ describe("2500: Datasets v4 tests", () => {
     it("0126: adds a new dataset with scientificMetadata", async () => {
       return request(appUrl)
         .post("/api/v4/datasets")
-        .send(TestData.ScientificMetadataForElasticSearchV4)
+        .send(TestData.ScientificMetadataForOpensearchV4)
```
minottic: minor: I find this mock name a bit misleading, if v4 is not using opensearch, why the data stub is called `ScientificMetadataForOpensearchV4`?

Junjiequan: I'm not sure who created this, but it was there already. Since we are going to extend opensearch for dataset v4 I just left it as it is.

minottic: I guess since we are renaming it in this PR, we could rename it to something more meaningful ?

src/datasets/datasets.service.ts
Junjiequan: For fullQuery we can use OS built in pagination logic to improve performance in the future if needed, iterate query multiple times with OS and concat until returned result meets the requested limit:
```

user requests: fullquery?limits={"limit":1000,"skip":0}&fields={"text":"text"}
1. Query OS in batches and returns pids 
2. Query MongoDB {$in: [pids], ..otherFilters} save results in cache
3. Concat to cache until limit met, then stop
4. If results < limit, fetch next OS batch and repeat
```

Fullfacet on the other hand, use the following query `fullfacet?facets=["creationLocation","ownerGroup","type","keywords"]&fields={"text":"text"}`, we don't pass any limit, as it needs to scan all documents to get accurate facets.

The response with fullfacet endpoint:

image

We can still use batch logic, but it essentially won't fix the fact that the query needs to scan all documents

minottic: ah yes, on that I agree, I was raising more the problem of loading so many entries in memory if the OS return is huge

src/datasets/datasets.service.ts
Junjiequan: I see, it looks cleaner that way.
Change included in [ad04376](https://github.com/SciCatProject/backend/pull/2562/commits/ad043764db58568394deca48f1a8fa62878609d3)

test/DatasetV4.js
Junjiequan: yes, change included in [3a1d1e5](https://github.com/SciCatProject/backend/pull/2562/commits/3a1d1e56b7e0bf32753c87195c37dd797e92688e)

minottic: thanks and sorry for the pedantic comment!

src/datasets/datasets.service.ts
Junjiequan: if fullquery and fullfacet with OS interception becomes issue, I will open a PR to address those issues. Meanwhile, we could also discuss if current fullfacet behavior, calculating all facetcounts, is really necessary.
For this PR I'd like to keep it as is, it's getting too large

minottic: works well for me, thanks! Just to clarify further, when I mentioned batching in my comment I meant mostly batching requests to mongo with PID batches, I think that's actually the more realistic issue. I thikn a relative small number of pids will make the where filter too large and the call to mongo fail.

Junjiequan: Yes, I'm totally agree with you. max and I have also discussed about the large array within $in clause cases. That will be the next improvement for opensearch integration.

src/datasets/datasets.controller.ts
Junjiequan: addressed in https://github.com/SciCatProject/backend/pull/2562/commits/7180eb3e91b99085e48683babc285491c3b07b5f

src/opensearch/opensearch.service.ts
Junjiequan: I have increased retry attempt up to 10 times, and moved retry to run on the background instead of  onModuleInit, so it does not block the application from startup 

[8b93dfe](https://github.com/SciCatProject/backend/pull/2562/commits/8b93dfe6081f3ddbaf3d9e66efb780bc29fc37f9)

src/datasets/datasets.controller.ts
```diff
@@ -985,12 +990,12 @@ export class DatasetsController {
...
-    const parsedFilters: IFilters<DatasetDocument, IDatasetFields> = {
-      fields: fields,
-      limits: JSON.parse(filters.limits ?? "{}"),
-    };
+    const isAdmin = canViewAny;
 
-    const datasets = await this.datasetsService.fullquery(parsedFilters);
+    const datasets = await this.datasetsService.opensearchQuery(
+      parsedFilters,
+      isAdmin,
```
minottic: sorry, I overlooked this before. Why is the `isAdmin` needed here? From the function definition I thought that userGroups would be enough?

Junjiequan: You are right, since `canViewAny` check blocks userGroups to be passed down to the filter query for admin, `isAdmin`  will not be needed here. Nice catch, thanks

src/datasets/datasets.service.ts
Junjiequan: added in [5c6b650](https://github.com/SciCatProject/backend/pull/2562/commits/5c6b6506174276bb81c0ca1ea7b21c9164ea3c98)

minottic: thanks! sorry, for being pedantic, but the implementation I think it's not very efficient, especially because of the gt filter. I think we can obtain the same bulking by using a cursor and an accumulator that we flush every batchsize.

```TS
try {
  await this.opensearchService.checkIndexExists(index);

  const bulkOperationFinalResult: BulkStats = {
    total: 0,
    failed: 0,
    retry: 0,
    successful: 0,
    noop: 0,
    time: 0,
    bytes: 0,
    aborted: false,
  };

  const cursor = this.datasetModel
    .find({}, DATASET_OPENSEARCH_EXCLUDE_FIELDS_QUERY)
    .lean()
    .cursor();

  let batch: DatasetClass[] = [];
  let isCursorExhausted = false;

  while (!isCursorExhausted) {
    const doc = await cursor.next();
    
    if (doc) {
      batch.push(doc as DatasetClass);
    } else {
      isCursorExhausted = true;
    }

    // Condition: Is the batch full OR are we at the very end with a non-empty tail?
    const isBatchReady = batch.length >= this.osSyncBatchSize;
    const isFinalTail = isCursorExhausted && batch.length > 0;

    if (!isBatchReady && !isFinalTail) {
      continue;
    }

    // Single source of truth for the bulk operation
    const bulk = await this.opensearchService.performBulkOperation<DatasetClass>(
      batch,
      index,
    );

    // Aggregate stats
    bulkOperationFinalResult.total += bulk.total;
    bulkOperationFinalResult.failed += bulk.failed;
    bulkOperationFinalResult.retry += bulk.retry;
    bulkOperationFinalResult.successful += bulk.successful;
    bulkOperationFinalResult.noop += bulk.noop;
    bulkOperationFinalResult.time += bulk.time;
    bulkOperationFinalResult.bytes += bulk.bytes;
    bulkOperationFinalResult.aborted = bulk.aborted;

    Logger.log(
      `Synced ${bulkOperationFinalResult.total} datasets to OpenSearch (Tail: ${isFinalTail})`,
      "OpensearchSync",
    );

    batch = []; // Reset
  }

  return bulkOperationFinalResult;
} catch (error: unknown) {
  const errorMessage = error instanceof Error ? error.message : "Unknown error";
  Logger.error(`Sync failed: ${errorMessage}`, "OpensearchSync");
  throw error;
}
```

Junjiequan: if I'm not mistaken, there's no memory related performance difference between using `cursor` or `$gt with limit`? from what I read about cursor, it is essientially using the index filter logic internally.
I'm open to use cursor here, as it seems to me almost the same. But the code above is a bit complex to me

Junjiequan: The only difference is that, with `cursor` we make db query only once and the rest happenes internally, the downside is we need to cache each doc in the memory, on the contrast, with `$gt + limit`, we make multiple queries to the DB.

minottic: doesn't mongo need to do a O(logN) operation on every find query to find the starting position of the lastIndex? while with the cursor we simply start from the first document and retrieve them (so basically O(1))

minottic: > downside is we need to cache each doc in the memory

don't we cache the same number of documents with the 2 approaches, i.e. an array of length batchSize (either from the limit or from the push)? the cursor is consumed on every next, so in practice (not considering the accumulator) there's always one doc in memory

Junjiequan: Yes, the cost behind O(1) is that we need to keep cache everything in the memory with additional logic to handle that.
As for O(logN) cost I think it depends how many times it needs to operate index lookup I think, but I believe mongo is very efficient for index lookup

minottic: only the pointer is cached not the full documents. And there's not much additional logic tbh, it's a while loop (as the current one) and building a local array rather than needing to manage the lastIndex

We have 500k datasets, so with the current implementation the mongo search would be O(logN * (500k / batchSize  - 1)) while the search with the cursor one is O(1) (plus the O(N) needed to unpack the documents but that happens in both cases)

Junjiequan: Ah, I just saw the new code. In the previous snippet I remember the batch stores the entire docs, thats why i mentioned cachcing everything in the memory.
new code looks cleaner to me but I wonder should we use [cursor.batchSize()](https://www.mongodb.com/docs/manual/reference/method/cursor.batchSize/) or not, or I will just keep it as configurable 

Based on the doc

> Note
> If the batch size is too large, the cursor allocates more resources than it requires, which can negatively impact query performance. If the batch size is too small, the cursor requires more network round trips to retrieve the query results, which can negatively impact query performance.

minottic: ah good idea! I would say we can hardcode (or default a reasonable size) e.g. 1000 documents and eventually add in the config an option to change? Though I doubt anyone would need to decrease that

Junjiequan: New commits has been added accordingly based on dicussion. 
I still think cursor uses `$gt + limit `logic just internally, considering it's default batchSize is 101 or 16Mb, but I'm not going to dig deep into it : D thanks for the good discussion

minottic: thanks a lot, and apologies for being pedantic!

src/datasets/datasets.service.ts
minottic: I mean here that we could define:

```TS
export class OpenSearchDTO extends OmitType(DatasetDocument, DATASET_OPENSEARCH_EXCLUDED_FIELDS) {}
```

and then replace `sanitizeDatasetForOpensearch<DatasetDocument>(savedDataset.toObject())`

with a usual `plainToInstance(OpenSearchDTO, savedDataset.toObject())`

and then get rid of the sanitizeDatasetForOpensearch function

(or similar)

Junjiequan: Yes, that is almost the same as what I have included here

[fb59db5](https://github.com/SciCatProject/backend/pull/2562/commits/fb59db50bde02cd0add130b33fe3c8c23f56ca5c)

I thought about using OmitType but considering the query searches only datasetName and description at the moment using Pikc is fine to me

minottic: thanks a lot! Yes, this works very well I believe, sorry, I added the comment just before your push

minottic (approved):
thanks a lot for the feature and for the patience!

src/opensearch/dto/dataset-opensearch.dto.ts
```diff
@@ -1,9 +1,17 @@
...
 import { DATASET_OPENSEARCH_FIELDS } from "src/opensearch/utils/dataset-opensearch.utils";
 
-export class DatasetOpenSearchDto extends PickType(
-  UpdateDatasetDto,
+export class DatasetOpenSearchDto extends OmitType(
+  OutputDatasetDto,
   DATASET_OPENSEARCH_FIELDS,
-) {}
+) {
+  @Expose() pid: string;
```
minottic, line 11: I was thinking (not blocking), would it make sense to review the dataset dto and add the expose there?

src/opensearch/dto/dataset-opensearch.dto.ts
```diff
@@ -1,9 +1,17 @@
...
+import { OmitType } from "@nestjs/swagger";
+import { Expose } from "class-transformer";
+import { OutputDatasetDto } from "src/datasets/dto/output-dataset.dto";
 
-import { UpdateDatasetDto } from "src/datasets/dto/update-dataset.dto";
 import { DATASET_OPENSEARCH_FIELDS } from "src/opensearch/utils/dataset-opensearch.utils";
 
-export class DatasetOpenSearchDto extends PickType(
-  UpdateDatasetDto,
+export class DatasetOpenSearchDto extends OmitType(
```
minottic: should this be a PickType?

src/opensearch/dto/dataset-opensearch.dto.ts
Junjiequan: I think it's better to keep the logic seperated, not mixing them. Keep OS interception in its module, and dataset should not care about what OS collects

src/opensearch/dto/dataset-opensearch.dto.ts
Junjiequan: Yes, It'd make sense to have PickType here

src/opensearch/dto/dataset-opensearch.dto.ts
minottic: but shouldn't these fields be exposed anyway in the dataset DTO? What I mean is that the dataset DTO is anyway incomplete, no?

Junjiequan: I'm not sure I understand the context you are talking about here. Are you saying to add `expose` for all the fields in dataset dto? or to `expose` fields that are used by OS only

minottic: I am advocating to add `expose` whenever needed in DTOs (in general a review of the DTOs would not hurt), since  if we will at any point need `plainToInstance(DataestDTO, mongoJson)` this will also include non desired fields, unless we have marked properly the DTO. And this would allow to remove the expose here, since it would be already in the dataset DTO. 

But ofc this can happen in any other PR, not here, just thought to comment here since I saw the change

Junjiequan: I agree that we need to review the dtos, there's some cleanup needs to be done I suppose.

It’s probably harder to decide in the main DTO which fields are needed for different use cases, but i'm not sure, maybe its not that hard to decide.

PR #2569, opened 2026-03-02
feat(publish): add support for ajv dynamic defaults

fpotier:
## Description

[Implements this idea](https://github.com/orgs/SciCatProject/discussions/1#discussioncomment-15825520)
Note: `PublishedData.metadata.publicationYear` is now set using the new dynamic default feature. This required a change in the schema that currently breaks the frontend. I can remove it if it's not desired but I thought it was an improvement as the previous behavior only set a default value on creation that could be overwritten with updates.

## Motivation

See dedicated [discussion](https://github.com/orgs/SciCatProject/discussions/1#discussion-9487421)
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation


Junjiequan:
Looks good to me.
I guess we need to provide customized ajv on the frontend with correponding packages `ajv-formats` and `ajv-keywords` ? reading from https://jsonforms.io/docs/validation

src/published-data/published-data.v4.controller.ts
```diff
@@ -373,6 +375,7 @@ export class PublishedDataV4Controller {
       }
     }
 
+    await this.validatorService.validate(updatePublishedDataDto);
```
minottic: don't we need to raise errors here if there are during the validation?

src/published-data/validator.service.ts
```diff
@@ -0,0 +1,148 @@
...
+    this.ajv = new Ajv2019({
+      useDefaults: true,
+      allErrors: true,
+      strict: false,
+    });
+    addFormats(this.ajv);
+    addKeywords(this.ajv);
+
+    const modulePath = this.configService.get<string>("ajvCustomDefinitions");
+    if (!isEmpty(modulePath)) {
```
minottic: minor:

```TS
if (isEmpty(modulePath)): return
```

src/published-data/validator.service.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  async validate(
+    publishedData:
+      | CreatePublishedDataV4Dto
+      | UpdatePublishedDataV4Dto
+      | PartialUpdatePublishedDataV4Dto,
+  ) {
+    await this.loadDynamicDefaultFunctions(publishedData);
+    const config = this.configService.get<Record<string, unknown>>(
```
minottic: this will be loaded on every validate, maybe better in the constructor?

minottic (commented):
can we add a file with an example of the js dynamic default files?

fpotier:
> Looks good to me. I guess we need to provide customized ajv on the frontend with correponding packages `ajv-formats` and `ajv-keywords` ? reading from https://jsonforms.io/docs/validation

I linked the corresponding PR

src/published-data/validator.service.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  async validate(
+    publishedData:
+      | CreatePublishedDataV4Dto
+      | UpdatePublishedDataV4Dto
+      | PartialUpdatePublishedDataV4Dto,
+  ) {
+    await this.loadDynamicDefaultFunctions(publishedData);
+
+    const validateFn = this.ajv.compile(this.config.metadataSchema as Schema);
```
minottic, line 103: this can heppen in the constructor?

minottic (approved):
minor comment, but otherwise very nice, thanks! 

We might need to add docs for this I believe.

I always get confused by what the difference is between keywords and dynamicDefaults, can I not replace keywords with dynamicDefaults and simplify the ajv js file?

minottic:
ah I forgot one comment... don't we need some API tests in the test folder that show the formPopulate rendering from the js file?

src/published-data/validator.service.ts
fpotier: no because I inject the publisheddata into the functions, mainly to access the datasets pids.

minottic: ah so you need the load before compile if I understand right? Thanks for explaining!

fpotier:
> minor comment, but otherwise very nice, thanks!
> 
> We might need to add docs for this I believe.
> 
> I always get confused by what the difference is between keywords and dynamicDefaults, can I not replace keywords with dynamicDefaults and simplify the ajv js file?

I added a test and some docs.

dynamicDefaults are a specialization of keywords, it's a built-in keyword that is executed on a specific case (the value is empty).
Keywords have a broader scope, eg. throw additional errors, modify data.

src/published-data/validator.service.ts
```diff
@@ -0,0 +1,157 @@
...
+>;
+export type ReadOnlyAttachmentsService = Pick<
+  AttachmentsService,
+  "findOne" | "findAll" | "count"
+>;
+
+@Injectable()
+export class ValidatorService {
+  private ajv: Ajv2019;
+  private static logger: Logger = new Logger(ValidatorService.name);
```
Junjiequan: Can we use a global logger instead of manually creating new logger instances so that all logs are sent to the external central logging system?

Junjiequan (approved):
Just one comment, otherwise looks good to me.

I think we need to manage the example files in a dedicated folder soon

Issue #2583, opened 2026-03-03
Basic filters broken in v3


sbliven:
## Summary

Filters seem to be broken on the `/api/v3/datasets?filter=` endpoint

## Steps to Reproduce

Here's a pretty basic curl call:

```
curl 'https://dacat.psi.ch/api/v3/datasets?filter=%7B%22where%22%3A+%7B%22ownerGroup%22%3A+%22a-35266%22%7D%2C+%22fields%22%3A+%5B%22id%22%5D%2C+%22limits%22%3A+%7B%22limit%22%3A+200%2C+%22skip%22%3A+0%2C+%22sort%22%3A+%7B%22datasetName%22%3A+%22asc%22%7D%7D%7D'  -H 'accept: application/json' -H "Authorization: Bearer $SCICAT_TOKEN"
```

This returns an empty body. It works on v4.

sbliven:
Apparently not a problem with all filters, but only those containing "date-like" queries under javascript's extremely broad definition of dates.

minottic:
closed by #2584

PR #2584, opened 2026-03-03
fix: stricter date validation for encode

minottic:
## Description
This will try to convert to dates only if they start with YYYY-MM-DD. This caused strings like "a-1234" to be converted to 1234 Jan 1


minottic:
closes issue #2583

sbliven (approved):
Even the [Date.parse docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format) are salty about the poor state of date parsing. I appreciated the link to 'Broken Parser – A Web Reality Issue'. I think your solution here is pragmatic. I don't think anyone was relying on dates like "Feb 30, 2026" parsing.

PR #2585, opened 2026-03-03
feat(ldap-auth): make querying user groups possible

cchndl:
The ldap authenticator did not query the groups from ldap until now.

Add the required config options for getting the groups from ldap. For this we just expose the existing options from node-ldapauth-fork/passport-ldapauth from the SciCat config.
Add a group extraction service for the LDAP response and register it to the servicefactory if enabled.

Additional fixes:
The `AccessGroupFromPayloadService` did not correctly check whether the elements in `payload[accessGroupsProperty]` are strings.

### DIFF
--- a/src/auth/access-group-provider/access-group-from-ldap.service.ts
+++ b/src/auth/access-group-provider/access-group-from-ldap.service.ts
@@ -13,7 +13,7 @@ export class AccessGroupFromLdapService extends AccessGroupService {
   }
 
   async getAccessGroups(userPayload: UserPayload): Promise<string[]> {
-    let accessGroups: string[] = [];
+    const accessGroups: string[] = [];
 
     const accessGroupsProperty = userPayload.accessGroupProperty;
     if (accessGroupsProperty) {
@@ -24,9 +24,9 @@ export class AccessGroupFromLdapService extends AccessGroupService {
       ) {
         for (const group of payload[accessGroupsProperty]) {
           if (
-              typeof group === "object" &&
-              "cn" in group &&
-              typeof group["cn"] === "string"
+            typeof group === "object" &&
+            "cn" in group &&
+            typeof group["cn"] === "string"
           ) {
             accessGroups.push(group["cn"]);
           }
--- a/src/auth/access-group-provider/access-group-from-ldap.service.ts
+++ b/src/auth/access-group-provider/access-group-from-ldap.service.ts
@@ -1,5 +1,4 @@
 import { Injectable, Logger } from "@nestjs/common";
-import { ConfigService } from "@nestjs/config";
 import { UserPayload } from "../interfaces/userPayload.interface";
 import { AccessGroupService } from "./access-group.service";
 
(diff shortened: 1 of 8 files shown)


src/auth/access-group-provider/access-group-from-ldap.service.ts
```diff
@@ -0,0 +1,39 @@
...
+        payload !== undefined &&
+        Array.isArray(payload[accessGroupsProperty])
+      ) {
+        for (const group of payload[accessGroupsProperty]) {
+          if (
+            typeof group === "object" &&
+            "cn" in group &&
+            typeof group["cn"] === "string"
+          ) {
+            accessGroups.push(group["cn"]);
```
fpotier: Maybe it would make sense to make this configurable, the ldap backend used in scicatlive seems to use `ou` rather than `cn`. The type in the payload is also not `object` but a simple string.
```
memberOf: [
      'ou=group-test,ou=groups,dc=facility',
      'ou=scicatlive,ou=groups,dc=facility'
    ]
```

src/auth/access-group-provider/access-group-from-ldap.service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  async getAccessGroups(userPayload: UserPayload): Promise<string[]> {
+    const accessGroups: string[] = [];
+
+    const accessGroupsProperty = userPayload.accessGroupProperty;
+    if (accessGroupsProperty) {
+      const payload: Record<string, unknown> | undefined = userPayload.payload;
+      if (
+        payload !== undefined &&
+        Array.isArray(payload[accessGroupsProperty])
```
fpotier: Is it correct if the user is only part of one group? scicatlive returns:
```
memberOf: 'ou=scicatlive,ou=groups,dc=facility'
```

fpotier (commented):
I tried the feature with scicatlive but couldn't make it work out of the box.
It's maybe because we use a [simplified ldap provider](https://glauth.github.io/).
Left a few comments but I'm not sure it's worth implementing as you seem to be the only facility using this.

src/auth/access-group-provider/access-group-from-ldap.service.ts
cfelder: In principle you are right, that making the ldap attribute configurable would make this code more general. However, the GLAuth in scicatlive uses the ldap object class ``groupOfUniqueNames`` that also has the ``cn`` attribute. The entry for the scicatlive group is:

```
[DN: ou=scicatlive,ou=groups,dc=facility - STATUS: Read - READ TIME: 2026-05-11T17:02:51.168567
     cn: scicatlive
     description: scicatlive
     gidNumber: 5501
     objectClass: groupOfUniqueNames
                  top
     ou: scicatlive
     uniqueMember: cn=ldap-user,ou=scicatlive,ou=users,dc=facility]
```

I don't understand why one would like to introduce an ``ou`` subtree in the ldap tree for every group. Nevertheless this patch works out of the box when using the following settings in scicatlive.

``services/backend/services/v4/config/.ldap.env``:

```
ACCESS_GROUPS_LDAPPAYLOAD_ENABLED=true
LDAP_ACCESS_GROUPS_PROPERTY=_groups
LDAP_GROUP_SEARCH_BASE=ou=groups,dc=facility
LDAP_GROUP_SEARCH_FILTER="(uniqueMember=cn={{username}},*ou=users,dc=facility)"
```

and ``LDAP_ENABLED=true`` in the ``.env``file.

src/auth/access-group-provider/access-group-from-ldap.service.ts
cfelder: Verified w/ scicatlive and the two users ``ldap-user`` and ``scicatlive-svc``

src/auth/access-group-provider/access-group-from-ldap.service.ts
fpotier: My bad, I did not realize that the library was populating the `_groups` property and was confused on how to make that patch work using `memberOf`.

fpotier (approved):

Junjiequan (approved):

Junjiequan:
Can we merge it or is there any blocker for this to be merged?

cchndl:
I missed that the environment variables were also in `docs/index.md`, but otherwise nothing from my side.

cfelder:
Besides making the ldap attribute for determining the group name configurable we also need this for determining the username. For non-ad setups e.g. using ``uid`` instead of ``displayName`` is still a local fix on our end.

Imho we can do it as a small addition within this PR. If you want to have it separate please let us know.

Junjiequan:
> Imho we can do it as a small addition within this PR. If you want to have it separate please let us know.

It looks find to me to have them in this PR. This PR is quite small and the implementation is straightforward, I'm happy for it to be merged

cchndl:
no points from my side anymore, +1 to the squash :)

cfelder:
Added a patch for configuring ldap w/ the GLAuth backend in scicatlive accordingly.

https://github.com/SciCatProject/scicatlive/pull/825

PR #2609, opened 2026-03-10
fix: move runNumber in legacy scientificMetada to top

### LINKED ISSUE #2592: Dataset's runNumber inconsistencies
RunNumber in datasets is sometimes considered part of [scientificMetadata others as a top dataset element](https://github.com/search?q=repo%3ASciCatProject%2Fbackend%20runnumber&type=code). In the old backend it was part of [scientificMetadata](https://github.com/search?q=repo%3ASciCatProject%2Fbackend-v3%20runNumber&type=code). The frontend treats it as [top level element](https://github.com/search?q=repo%3ASciCatProject%2Ffrontend%20runNumber&type=code).

I would suggest we move it back to scientificMetadata (and the FE fetching accordingly), and have a migration which moves back

minottic:
## Description
Since runNumber was moved from scientificMetadata to a top level dataset property #1485 , this PR adds a migration for legacy data and a fallback when posting or patching v3 datasets which moves scientificMetadata.runNumber to runNumber. This 2nd change is to facilitate customers' adoption as they would need to change their ingestions.

Closes #2592 

## Tests included

- [x] Passing? 

## Documentation


src/datasets/dto/update-dataset-obsolete.dto.ts
```diff
@@ -277,6 +278,7 @@ export class UpdateDatasetObsoleteDto extends OwnableDto {
   @IsOptional()
   @IsObject()
   @Transform(({ value }) => encodeScientificMetadataKeys(value))
+  @Transform(({ value }) => omit(value, "runNumber"), { toClassOnly: true })
```
bpedersen2: Again I would probably keep it duplicated for now

migrations/20260309161001-move-run-number-to-top.js
```diff
@@ -0,0 +1,56 @@
...
+    const stats = { modifiedCount: 0, unModifiedCount: 0 };
+    const collection = db.collection("Dataset");
+    for await (const dataset of collection
+      .find({ "scientificMetadata.runNumber": { $exists: false }, "runNumber": { $exists: true } })) {
+      bulkOps.push({
+        updateOne: {
+          filter: { _id: dataset._id },
+          update: {
+            $set: { "scientificMetadata.runNumber": dataset.runNumber },
+            $unset: {runNumber: ""},
```
bpedersen2: for the down path that seems ok.

migrations/20260309161001-move-run-number-to-top.js
```diff
@@ -0,0 +1,56 @@
...
+    const stats = { modifiedCount: 0, unModifiedCount: 0 };
+    const collection = db.collection("Dataset");
+    for await (const dataset of collection
+      .find({ "scientificMetadata.runNumber": { $exists: true }, "runNumber": { $exists: false } })) {
+      bulkOps.push({
+        updateOne: {
+          filter: { _id: dataset._id },
+          update: {
+            $set: { runNumber: dataset.scientificMetadata.runNumber },
+            $unset: {"scientificMetadata.runNumber": ""},
```
bpedersen2: I would actually keep the old field for now and delete it later (if at all) as also non-scicat task may rely on this field

minottic: but doesn't that call more for a field in the DTO rather than in the DB (which I will be happy to add)? Also, all datasets created with a payload that had runNumber as a top element will miss the one in scientificMetadata, unless we add a migration for that one as well

migrations/20260309161001-move-run-number-to-top.js
Junjiequan: Does this script addresses the following runNumber format?
```
                    "runNumber" : {
                        "value" : NumberInt(482),
                        "unit" : ""
                    },
```

minottic: is anyone doing that? runNumber is simply an integer in the top level schema so I guess noone did that? I think we would have a hard time trying to imagine how everyone could have used runNumber in scientificMetadata, but if you reckon this shape is something they used, I can add to the migration

migrations/20260309161001-move-run-number-to-top.js
minottic: added as part of the v3 return DTO

migrations/20260309161001-move-run-number-to-top.js
minottic: also, what should the migration do if there are `unit`? Simply concat?

Junjiequan: I just checked our DB and some scientificMetadata contains the format above. 
This change does not affect our side, as we are going to use high level run number anyways.
But I'm not sure would it affect others or not

minottic: ok, happy to add that!

nitrosx:
I am not oppose to this migration script but it can only be optional.
Even if we have an high level field for run number, it is only a best practice to use this field.
We cannot assume what a facility decides to do  with run number: use the high level field or store it in a scientific metadata entry.

Also it is a facility decision if they want to migrate legacy data or not and how to convert it from quantity (if they have it as quantity) to a numeric field.

minottic:
> I am not oppose to this migration script but it can only be optional.
> Even if we have an high level field for run number, it is only a best practice to use this field.
> We cannot assume what a facility decides to do with run number: use the high level field or store it in a scientific metadata entry.
> 
> Also it is a facility decision if they want to migrate legacy data or not and how to convert it from quantity (if they have it as quantity) to a numeric field.
> 

the migration will run only if there are datasets with scientificMetadata.runNumber. I am not sure I understand the comment fully, doesn't this apply to all migrations then? This is a change in behaviour between the old backend and the new one, so a migration is needed. For example the FE runNumber in the datasets overview used to rely on scientificMetadata.runNumber and now fetches the top level run number. People will always be able to use scientificMetadata.runNumber in the v3 endpoints and this will move it to the topLevel runNumber. One could argue why this field was moved to a top level field without checking the old behaviour, but that's another question...

nitrosx:
> > I am not oppose to this migration script but it can only be optional.
> > Even if we have an high level field for run number, it is only a best practice to use this field.
> > We cannot assume what a facility decides to do with run number: use the high level field or store it in a scientific metadata entry.
> > Also it is a facility decision if they want to migrate legacy data or not and how to convert it from quantity (if they have it as quantity) to a numeric field.
> 
> the migration will run only if there are datasets with scientificMetadata.runNumber. I am not sure I understand the comment fully, doesn't this apply to all migrations then? This is a change in behaviour between the old backend and the new one, so a migration is needed. For example the FE runNumber in the datasets overview used to rely on scientificMetadata.runNumber and now fetches the top level run number. People will always be able to use scientificMetadata.runNumber in the v3 endpoints and this will move it to the topLevel runNumber. One could argue why this field was moved to a top level field without checking the old behaviour, but that's another question...

The migration should run only if the facility decided to do so.
It is not up to SciCat project to force a facility to use high level field `runNumber`, although we can suggest it as best practice.
If a facility has datasets with scientific metadata entry named `runNumber`, we can not assume that they want to migrate and the migration makes sense.
I will be happy to discuss in person

src/datasets/dto/output-dataset-obsolete.dto.ts
```diff
@@ -257,5 +257,12 @@ export class OutputDatasetObsoleteDto extends UpdateDatasetObsoleteDto {
...
   })
   @Transform(({ value }) => decodeScientificMetadataKeys(value))
+  @Transform(
+    ({ value, obj }) => {
+      if (!obj?.runNumber) return value;
+      return { runNumber: obj.runNumber, ...value };
+    },
+    { toClassOnly: true },
+  )
   declare scientificMetadata?: Record<string, unknown>;
```
nitrosx, line 268: Could you please explain better what this snippet of code?
If you are removing a field named runNumber from scientific metadata, we cannot allow this, as some users can have a field named `runNumber` which might have a different meaning in their context and is also a valid entry

nitrosx (dismissed):
This change, although technical and syntactically correct, needs more community discussion

minottic:
> The migration should run only if the facility decided to do so.

but why not for all migrations then? Also, noone forces anyone to run migrations, they are not run on startup. We do the same exact thing [here](https://github.com/SciCatProject/backend/blob/master/migrations/20250827150710-jobs-v3-to-v4-ids.js)

Say I am a user and I have a look at the POST datasets DTO. I will see a runNumber field. So I will add my runNumber there. Then all historical metadata used to use scientificMetadata.runNumber. If no migration I will have inconsistent behaviour

If you think we should not standardise how to use runNumber as a top level element, I think this will create a lot of confusion for the use, but I can manage that from my side

src/datasets/dto/output-dataset-obsolete.dto.ts
minottic: good point, I think it misses indeed an omit if missing logic

minottic: I removed the omit. There is a bit of info duplication, but we will not incur into different meaning assumptions, thanks (see commit)! 

About the code here, the spread operator will override runNumber if in value, so we are not overriding unless empty

minottic:
just to summarise the change, given the latest commit:

1. on PATCH/POST on v3 endpoint runNumber is populated from scientificMetadata.runNumber if runNumber is missing. Both are kept (we might think to remove the scientificMetadata.runNumber in this case but I am fine keeping it)
2. on GET on v3 scientificMetadata.runNumber is populated from runNumber if missing

since the migration is a source of discussion, I am happy to manage it from my side and not have it in this repo

minottic:
any further reason this is blocked @nitrosx ?

minottic:
@nitrosx , @Junjiequan any news here?

src/datasets/dto/output-dataset-obsolete.dto.ts
Junjiequan: Maybe move spreading to the left would be more intuitive here? I was confused a bit - why overwritting after value is assigned here
```js
      return { runNumber: obj.runNumber, ...value };
```

Junjiequan (dismissed):
Looks good to me

src/datasets/dto/output-dataset-obsolete.dto.ts
minottic: done [here](https://github.com/SciCatProject/backend/pull/2609/changes/52c958d834c41a5f9afbfe52266a04c74f5c6359)

Junjiequan:
If this PR is urgent we can dismiss review from @nitrosx , as he will mostly not available untill the week after next week.
I don't see any other blockers with this PR

minottic:
> If this PR is urgent we can dismiss review from @nitrosx , as he will mostly not available untill the week after next week.
> I don't see any other blockers with this PR

If you are happy with the changes, I would indeed proceed with the merge

Junjiequan:
I just thought about one potential issues here. if we allow overwritting runNumber from scientificMetadata.runNumber if not exists, what about old datasets? wouldn't it causing incosistent datasets?
For example, when creating or updating a dataset, dataset.runNumber may not be provided, but the user might set scientificMetadata.runNumber for a different purpose than the actual meaning of runNumber.

minottic:
> I just thought about one potential issues here. if we allow overwritting runNumber from scientificMetadata.runNumber if not exists, what about old datasets? wouldn't it causing incosistent datasets?

That's exactly why we will run a migration at PSI. I think it's indeed not a trivial matter, because: 

1. the FE consumes top level runNumber. Before it used to consume scientificMetadata.runNumber. Already this, kinda supports the argument that they have the same meaning
2. If I want runNumber to be displayed in FE I need to use the top level field
3. Even if I did not want the FE to use this field, any new user who would see the DTO of Datasets will put runNumber as a top level field, as that's the most intuitive thing if having a look at the schema (they would not think they need to use the scientificMetadata.runNumber)
4. Given that we have legacy users and data, we will very likely have historic users using "scientificMetadata.runNumber" and new users using "runNumber"
5. This will create inconsistencies
6. To overcome this, PSI will run a migration of its data moving scientificMetadata.runNumber to runNumber and the  change of this PR would keep new data consistent

I am happy with other solutions but this bug is causing complains from our users.

Also, I honestly think asking ourselves "what if the user meant something by leaving runNumber empty and used scientificMetadata.runNumber" is a sensible question, but in reality a bit too theoretic, I believe anyone seeing runNumber would interpret it in the same way as scientificMetadata.runNumber and simply leave one of the 2 out (likely the nested one as it's not in the DTO)

Last point, when we moved scientificMetadata.runNumber to runNumber not only we caused this possible inconsistence, but we also introduced a breaking change, as any client would not know anymore what field to consume.

Junjiequan:
After this change I'm afraid It still remains the root issue that what if user’s intent for scientificMeata.runNumber is not meant to be used for dataset.runNumber.

minottic:
sorry, I updated the comment [here](https://github.com/SciCatProject/backend/pull/2609#issuecomment-4098665990) quite a bit

minottic:
and also, we've done these defaultings a lot of time it's done in publishedData DTOs and job DTOs.

But maybe to be pragmatic, what if I make another PR where I expose conditionally runNumber as a top level field? In this way, PSI could opt out from having it as a top level field and keep the old behaviour or I can make the change in this PR configurable via ENV?

Junjiequan:
Is the complaining comes from mainly for FE display ? if it's the case the table config accepts `scientificMetadata.runNumber.value` and displays like so:
image
also within labelLocalization the label can be renamed to runNumber

minottic:
no, the main problem will be this

> Even if I did not want the FE to use this field, any new user who would see the DTO of Datasets will put runNumber as a top level field, as that's the most intuitive thing if having a look at the schema (they would not think they need to use the scientificMetadata.runNumber)

Junjiequan:
In this PR I think the output part looks fine to me, my only concern was the create/update  dto part which tries to predict/ twist user’s intent that it assumes dataset.scientificMeatadata.runNumber = dataset.runNumber

Junjiequan:
If I understand the whole situation corretctly, the issue is that legacyusers or services that dont want to use or see the top level runNumber , as it can be confusing which one to use.

if that's the case, please go ahead and add env variables within this PR and expose the field conditionally. And we can treat top level runNumber as a new feature in v4

minottic:
> In this PR I think the output part looks fine to me, my only concern was the create/update dto part which tries to predict/ twist user’s intent that it assumes dataset.scientificMeatadata.runNumber = dataset.runNumber

I think the change cannot happen only in the get dto though, we have dashboarding tools that query the db directly, I'd like to avoid querying for both fields in each query. 

But that's valuable feedback, I'll propose something different next week

nitrosx (changes requested):
@minottic apologies for being so demanding!!!
can you add a configuration variable in order to enable or disable this behavior?

src/datasets/dto/output-dataset-obsolete.dto.ts
```diff
@@ -257,5 +257,13 @@ export class OutputDatasetObsoleteDto extends UpdateDatasetObsoleteDto {
     description: "JSON object containing the scientific metadata.",
   })
   @Transform(({ value }) => decodeScientificMetadataKeys(value))
+  @Transform(
+    ({ value, obj }) => {
```
sbliven, line 261:
```suggestion
    // Add scientificMetadata.runNumber for backwards compatibility
    ({ value, obj }) => {
```

src/datasets/dto/update-dataset-obsolete.dto.ts
```diff
@@ -334,6 +334,10 @@ export class UpdateDatasetObsoleteDto extends OwnableDto {
   })
   @IsOptional()
   @IsString()
+  @Expose()
+  @Transform(({ value, obj }) => value ?? obj.scientificMetadata?.runNumber, {
```
sbliven, line 338:
```suggestion
  @Transform(
    // Copy scientificMetadata.runNumber to top level for backwards compatibility
    ({ value, obj }) => value ?? obj.scientificMetadata?.runNumber, {
```

sbliven: This should also check `obj.scientificMetadata?.runNumber?.value` first. And it should check the type, and only copy numbers.

sbliven (approved):
I think the current version is quite reasonable, and addresses the core issue that **v4 broke backwards compatibility on runNumber**. With v3 endpoints we have generally held a policy that optional adding fields to the DTO is allowed, but that the behavior of the legacy endpoints should be preserved as much as possible. This PR follows this philosophy. If users want unambiguous DTOs they should update clients (and possibly the frontend, but that's a bigger discussion) to use v4 endpoints.

Some more comments would be helpful, as the Transform logic wasn't obvious to me at first. Basically, this adds:

- When posting or patching a dataset via /api/v3 endpoints, the `Dataset.runNumber` will be set to the first field found among:
  1. `dto.runNumber`
  2. `dto.scientificMetadata.runNumber`
- When getting a dataset using /api/v3/ endpoints, the `dto.runNumber` field is new so v3 clients will ignore it, and instead look for `dto.scientificMetadata.runNumber`. Thus, if there is a top-level runNumber set it should also be returned at the old location.

> After this change I'm afraid It still remains the root issue that what if user’s intent for scientificMeata.runNumber is not meant to be used for dataset.runNumber.

There are no such cases for v3 endpoints. Since the field was added in v4, only v4 endpoints should distinguish the fields semantically. The only edge case where this PR would even give an unintended result is if the users intentionally submitted a dataset with an _empty_ top-level runNumber and a non-empty scientificMetadata.runNumber; this seems vanishingly unlikely. If they filled both fields but with different values then both are preserved by this PR, even in v3.

> can you add a configuration variable in order to enable or disable this behavior?

I am strongly opposed to making this behavior configurable.

PR #2649, opened 2026-04-05
feat: published data schema update

nitrosx:
## Description
This PR adds fields to store proposals and samples in the published data record.

## Motivation
Some data policies requires that a DOi is assigned to the whole proposal as soon as it is approved. This PR will allow to assign and register a new DOI as soon as a proposal is accepted and saved in SciCat.
Also, more and more, DOI can be register for a non homogeneous group of information which can contain proposals, samples and datasets alike.

## Changes:
* published data schema

## Tests included
N/A

## Documentation


HayenNico (changes requested):
The code looks fine, but I think we should not make the `datasetPids` field optional at this time. The DOI registration at DataCite is currently hardcoded to use `Dataset` as the `resourceTypeGeneral` (see e.g.  [here](https://github.com/SciCatProject/backend/blob/ed032625dbeaecbcfc88fcca912fc303758fdd99/src/published-data/published-data.v4.controller.ts#L772) in the v4 controller). This is not suitable if a `PublishedData` instance were to include only samples or proposals.
It would be fine to add links to sample and proposal documents under a `publishedData` instance so long as these do not have their own DOIs (standalone sample DOI publications should for example use [IGSNs](https://ev.igsn.org/)).

nitrosx:
> The code looks fine, but I think we should not make the `datasetPids` field optional at this time. The DOI registration at DataCite is currently hardcoded to use `Dataset` as the `resourceTypeGeneral` (see e.g. [here](https://github.com/SciCatProject/backend/blob/ed032625dbeaecbcfc88fcca912fc303758fdd99/src/published-data/published-data.v4.controller.ts#L772) in the v4 controller). This is not suitable if a `PublishedData` instance were to include only samples or proposals. It would be fine to add links to sample and proposal documents under a `publishedData` instance so long as these do not have their own DOIs (standalone sample DOI publications should for example use [IGSNs](https://ev.igsn.org/)).

I'm fine to keep `datasetPids` mandatory for now, but for our data policy, we will have registered published data with only proposal until the datasets acquired under the proposal will become public. I assume that we would need to update the logic for `resourceTypeGeneral` if we would like to accomodate this usecase or register a second published data once the datasets are public.
Thoughts?

nitrosx:
@HayenNico which `resourceTypeGeneral` would you pick for published data with proposal, datasets and potentially samples? `Collection`?
You can find the list [here](https://datacite-metadata-schema.readthedocs.io/en/4.7/appendices/appendix-1/resourceTypeGeneral)

Junjiequan:
From what I found on the list, they have sample-ish (PhysicalObject) and proposal-ish (Project) resourcesType.

HayenNico:
@nitrosx @Junjiequan Which `resourceTypeGeneral` you choose depends a lot on what the main content of the resource you're publishing is. I think that for `publishedData` documents this should always be `Dataset`, that's what the name and schema communicate.
And your use case would to my understanding still ultimately be a publication of all datasets belonging to a proposal - just with the added requirement that the datasets do not exist at the point of `publishedData` creation, correct? Bundling the proposal with that data is not an issue in my opinion, the main resource is still a `Dataset`. `Collection` could be a suitable fallback, but you lose a layer of granularity and semantic annotation.

My concern is that the schema in this form could be used to register `publishedData` with no data. Standalone publications of samples or proposals should probably live in their own collections (e.g. `publishedSample`, `publishedProposal`) if we want to allow that. Samples should generally be `PhysicalObject`, for proposals `StudyRegistration` fits best.

Perhaps an easy workaround to support your use case would be to make `datasetPids` optional, but throw an error when trying to register a DOI for a `publishedData` instance with an empty list there? Or does your data policy explicitly require the DOI to be registered with DataCite immediately?

nitrosx:
@HayenNico great points. Let's discuss at the meeting.

As we stand now, we would like to create a published data record and register it as soon as the proposal is accepted.
Datasets will be added when they become public (end of embargo period or manually published). So the end objectives is to have a published data which will contains datasets and relative proposal and maybe samples.

Regarding published data, I always interpreted with a wider meaning. It is public and registered, I do not know what it contains. It can be a mix of datasets, proposals or samples.

I like your suggestion that we can make datasets, proposals and samples optional with the condition that when I make them public and register the published data record, the combined list of the three has to contain at least one element.

HayenNico:
@nitrosx Following up on the discussion at the meeting: For your use case, I think you'd want `resourceTypeGeneral` to be `Dataset` for the final result, the question for me is what should be done in the intermediate stage where you only have the proposal.

DataCite supports different status levels for registered DOIs: `Draft` (DOI registered at DataCite, but *not* the global Handle system), `Registered` (Registered globally, but not discoverable via DataCite search) and `Published`. The current implementation in SciCat only uses the `Published` status, maybe adding some form of pre-registration with the `Registered` status would make sense? This is assuming that the initial `publishedData` with just the proposal is not really something you want incorporated in downstream applications like knowledge graphs (which depend on correct resource annotation with `resourceTypeGeneral` to be useful).

We need to have a look at the `publishedData` lifecycle in any case - as far as I remember, the `datasetPids` list is supposed to get locked down and be uneditable after DOI registration, which would block your workflow. Idk if that was actually implemented, but will need to be revisited if so.

HayenNico:
As an aside:
For samples specifically, we use standalone persistent identifiers called IGSNs in DAPHNE - these are DataCite DOIs with some narrower restrictions. The two main ones: `resourceTypeGeneral` *must* be `PhysicalObject`, and IGSNs must have a dedicated DOI-prefix space not used for other resources.
I had been considering an integration of IGSN with SciCat, and due to the second restriction they would have to be a separate entity `publishedSample` since they can't share the same DOI-prefix as `publishedData`. An additional aspect is that we'd want a separate DataCite schema configuration, since a site's standard descriptions of a sample and a dataset would most likely have different requirements.
Given the size of this project, I hadn't brought it up yet since I can't realistically dedicate enough time

nitrosx:
@HayenNico maybe should do a brainstorming session and continue the discussion somewhere else.
Can review this PR and see if your comments have been addressed?

src/published-data/schemas/published-data.schema.ts
```diff
@@ -104,16 +104,36 @@ export class PublishedData extends QueryableClass {
     type: [String],
     required: true,
     description:
-      "Array of one or more Dataset persistent identifier (pid) values that" +
-      " make up the published data.",
+      "Array of one or more datasets' persistent identifier values that" +
+      " are part of the published data record.",
   })
-  @Prop({ type: [String], required: true })
+  @Prop({ type: [String], required: false })
```
HayenNico: Should be `required: true` here to match `ApiProperty` information

src/published-data/schemas/published-data.schema.ts
```diff
@@ -104,16 +104,36 @@ export class PublishedData extends QueryableClass {
...
 
+  @ApiProperty({
+    type: [String],
+    required: false,
+    description:
+      "Array of one or more proposals identifier values that" +
+      " are part of this published data record.",
+  })
+  @Prop({ type: [String], required: false })
+  proposalIds: string[];
```
HayenNico: Should be `proposalIds?: string[]` if optional;
could also make this required for the schema with a default empty array

src/published-data/schemas/published-data.schema.ts
```diff
@@ -104,16 +104,36 @@ export class PublishedData extends QueryableClass {
...
+
+  @ApiProperty({
+    type: [String],
+    required: false,
+    description:
+      "Array of one or more samples identifier values that" +
+      " are part of this published data record.",
+  })
+  @Prop({ type: [String], required: false })
+  sampleIds: string[];
```
HayenNico: Same as proposalIds

src/published-data/dto/update-published-data.v4.dto.ts
```diff
@@ -6,50 +6,97 @@ import {
...
+  })
   @IsString()
   readonly abstract: string;
 
-  /**
-   * Array of one or more Dataset persistent identifier (pid) values that make up the published data.
-   */
+  @ApiProperty({
+    type: [String],
+    required: false,
```
HayenNico: Should go back to `required: true` to match schema

HayenNico (approved):
@nitrosx There's a few spots with minor inconsistencies for the handling of the different pid arrays. Looking good otherwise.
I'll pre-approve since the fixes are small.

src/published-data/published-data.v4.controller.ts
```diff
@@ -101,6 +101,7 @@ export class PublishedDataV4Controller {
   async create(
     @Body() createPublishedDataDto: CreatePublishedDataV4Dto,
   ): Promise<PublishedData> {
+    console.log("published data v4 create");
```
fpotier: should probably be removed from the final pr

fpotier (commented):
The auth mechanism is not changed in this PR but I think it would make sense to add a migration to back populate the owner/accessGroups from the datasets (I don't think instrumentGroup is used but I might be wrong)

nitrosx:
@fpotier good idea!!! Can we add the script in a different PR?

fpotier (changes requested):
After further review, I think this PR is not a feature but a breaking change to the api. Adding a new required field (`ownerGroup`) makes publishing through the frontend impossible right now.

nitrosx:
@fpotier I am wondering if we can enforce the ownership only in api/v4. Would that be acceptable?

If the ownership needs more discussion, I will split this PR in two: one for the new schema fields (which is the change that I need the most) and one for the ownership.

Thoughts?

nitrosx:
@fpotier I think I have the solution. Please review the following plan. Most likely it needs few more changes, but I think it will address your concerns.

# Plan

## 1. Migration Strategy for Old Records
- Add a database migration to set a default `ownerGroup` for all existing records without one (e.g., a system group like `"public"` or `"legacy"`).
- Set default `status` to `PUBLIC` for old records (since they should be accessible to everyone).

---

## 2. Access Control Logic
Modify `publishedDataEndpointAccess` in `casl-ability.factory.ts` to:

- **For unauthenticated users:** Allow read access to records where:
  - `status` is `PUBLIC`, `REGISTERED`, or `AMENDED` (everyone can access)
  - **OR** `ownerGroup` is not set (legacy records).

- **For authenticated users:**
  - `PRIVATE`: Only owner can access (check `ownerGroup` matches user's groups).
  - `PUBLIC`/`REGISTERED`/`AMENDED`: Everyone can access.
  - Legacy (no `ownerGroup`): Treat as `PUBLIC`.

---

## 3. Status-Specific Rules
   **Status**   | **Access**                                                                 |
 |--------------|----------------------------------------------------------------------------|
 | PRIVATE      | Only owner can read/write                                                |
 | PUBLIC       | Everyone can read, only owner can write                                  |
 | REGISTERED   | Everyone can read, only owner can write, has DOI                         |
 | AMENDED      | Everyone can read, only owner can write, has DOI, marked invalid         |

---
## 4. Implementation Files
- `src/casl/casl-ability.factory.ts` – Update `publishedDataEndpointAccess`.
- Create migration script for existing data.

---
**Would you like me to proceed with implementing this plan?**

---

# Plan for Ownership & Status Access Control

## Migration Strategy
- Database migration: Set default `ownerGroup: "public"` and `status: PUBLIC` for all existing records without these fields.

---

## Access Control Logic (`casl-ability.factory.ts`)
Modify `publishedDataEndpointAccess` to implement:
 | **Status**   | **Unauthenticated** | **Authenticated (Non-owner)** | **Owner**          |
 |--------------|---------------------|--------------------------------|--------------------|
 | PRIVATE      | ❌                  | ❌                             | ✅ Read/Write      |
 | PUBLIC       | ✅ Read             | ✅ Read                        | ✅ Read/Write      |
 | REGISTERED   | ✅ Read             | ✅ Read                        | ✅ Read/Write      |
 | AMENDED      | ✅ Read             | ✅ Read                        | ✅ Read/Write      |
 | Legacy       | ✅ Read             | ✅ Read                        | ✅ Read/Write      |

---
### Key Implementation
- Legacy records (missing `ownerGroup`) → treated as `PUBLIC`.
- Status check: `status !== PRIVATE` allows public read access.
- Owner check: user's `currentGroups` includes record's `ownerGroup`.

src/casl/casl-ability.factory.ts
```diff
@@ -1093,24 +1094,53 @@ export class CaslAbilityFactory {
...
-      user.currentGroups.some((g) => this.accessGroups?.delete.includes(g))
-    ) {
-      /*
-        / user that belongs to any of the group listed in DELETE_GROUPS
-        */
-      can(Action.Delete, PublishedData);
+      // Update access: owner of the record OR legacy records (no ownerGroup)
+      can(Action.Update, PublishedData, {
+        ownerGroup: { $in: user.currentGroups },
+      });
```
HayenNico, line 1132: This condition needs to go in the InstanceAccess function, since conditions can't be evaluated on the endpoint auth. A blanket `can(Action.Update, PublishedData);` should remain here.

Additionally, the "no owner group" case is not covered here - post-migration this would be an additional ability in the instance function:
```
can(Action.Update, PublishedData, {
  ownerGroup: "public",
});
```

migrations/20260605150000-add-ownership-to-published-data.js
```diff
@@ -0,0 +1,50 @@
...
+        ],
+      })
+      .forEach(async (publishedData) => {
+        const pid = publishedData._id;
+        
+        // Set default ownerGroup to "public" for backward compatibility
+        // Set default status to "public" if not set or is null
+        const update = {
+          $set: {
+            ownerGroup: "public",
```
HayenNico, line 22: Since the schema now implements Ownable, it will also include the `isPublished` field, which should also be set to true for legacy records

migrations/20260605150000-add-ownership-to-published-data.js
```diff
@@ -0,0 +1,50 @@
...
+      });
+  },
+  async down(db, client) {
+    // Remove ownerGroup from records that were added by this migration
+    // We can't easily revert this, but we'll unset ownerGroup for records that had it set to "public"
+    await db
+      .collection("PublishedData")
+      .updateMany(
+        { ownerGroup: "public" },
+        { $unset: { ownerGroup: true } },
```
HayenNico, line 45: Also unset `isPublished`

HayenNico (changes requested):
The `PublishedData` schema inherits the `isPublished` field from `OwnableClass`, which also should be properly handled. Its purpose is entirely duplicated with what `status` already does, so I'd suggest it's false for PRIVATE and true otherwise to match other collections.

This also requires a change in the controller to correctly update `isPublished` when status changes (could be an on-save trigger in the module as well).
How should the backend react when:
1. An update comes in with only `isPublished = true` (status ambiguous)?
2. An update comes in with conflicting `status` and `isPublished` info?

fpotier:
> @fpotier I think I have the solution. Please review the following plan. Most likely it needs few more changes, but I think it will address your concerns.

Sorry for the late feedback I was away.
The plan sounds good to me but my concern over adding a new mandatory field still remains when creating new records.
My preferred solution would be to either:
- make the `ownerGroup` field optional at the api level (maybe default it if we want to make it mandatory in the DB) but I don't think this is possible when relying on inheritance like we currently do.
or
- make it required but bump the api version to 5

#2747 was a similar case and imo when adding a required field, the api version should be bumped.

PR #2661, opened 2026-04-08
feat: update relationship schema

omkar-ethz:
## Description

Add fields to relationship schema used by `relationships` field in the dataset schema.

Testing:
PATCH /api/v4/datasets/:pid
```json
{
  "relationships": [
    {
      "identifier": "https://scilog.development.psi.ch/logbooks/6895bea625f055bca783dfdd",
      "identifierType": "URL",
      "entityType": "Logbook",
      "externalId": "6895bea625f055bca783dfdd"
    },
    {
      "identifier": "10.1016/j.epsl.2011.11.037",
      "identifierType": "DOI",
      "entityType": "JournalArticle"
    }
  ]
}
```
Updated dataset contains:
```json
"relationships": [
  {
    "identifier": "https://scilog.development.psi.ch/logbooks/6895bea625f055bca783dfdd",
    "identifierType": "URL",
    "relationship": "IsReferencedBy",
    "entityType": "Logbook",
    "externalId": "6895bea625f055bca783dfdd",
    "_id": "69d7b7e797c6654f48cab9df"
  },
  {
    "identifier": "10.1016/j.epsl.2011.11.037",
    "identifierType": "DOI",
    "relationship": "IsReferencedBy",
    "entityType": "JournalArticle",
    "_id": "69d7b7e797c6654f48cab9e0"
  }
]
```

A corresponding frontend feature (widget / tab on dataset detail page) will follow.

## Motivation

Previously, RelationshipClass was dataset specific. However, there is a need to link to related entities outside of SciCat, e.g. a SciLog logbook.
So we add `relatedEntityType` (inspired by datacite's [resourceTypeGeneral](https://datacite-metadata-schema.readthedocs.io/en/4.7/properties/relatedidentifier/#f-resourcetypegeneral)) . The existing `relationship` field is roughly  [relationType](https://datacite-metadata-schema.readthedocs.io/en/4.7/properties/relatedidentifier/#b-relationtype)

## Fixes

* NA

## Changes:

* generalize description of pid and relationship fields from "dataset" to "entity"
* add additional fields to the schema as indicated above

## Tests included

- [ ] Passing? 

## Documentation

### DIFF
--- a/src/datasets/dto/create-relationship.dto.ts
+++ b/src/datasets/dto/create-relationship.dto.ts
@@ -1,5 +1,5 @@
-import { ApiProperty } from "@nestjs/swagger";
-import { IsString } from "class-validator";
+import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
+import { IsOptional, IsString, IsUrl } from "class-validator";
 
 export class CreateRelationshipDto {
   @ApiProperty({
@@ -10,11 +10,30 @@ export class CreateRelationshipDto {
   @IsString()
   readonly pid: string;
 
-  @ApiProperty({
+  @ApiPropertyOptional({
     type: String,
-    required: true,
-    description: "Relationship between this dataset and the related one.",
+    description: "Relationship between this dataset and the related entity.",
+    default: "is related to",
   })
   @IsString()
-  relationship: string;
+  @IsOptional()
+  readonly relationship?: string;
+
+  @ApiPropertyOptional({
+    type: String,
+    description:
+      "Type of the related entity (e.g., 'Dataset', 'Logbook', 'Other').",
+    default: "Other",
+  })
+  @IsString()
+  @IsOptional()
+  readonly relatedEntityType?: string;
+
+  @ApiPropertyOptional({
+    type: String,
+    description: "URL to access the related entity, if applicable.",
+  })
+  @IsString()
+  @IsUrl()
+  readonly url?: string;
 }
--- a/src/datasets/dto/create-relationship.dto.ts
+++ b/src/datasets/dto/create-relationship.dto.ts
@@ -1,19 +1,33 @@
 import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
-import { IsOptional, IsString, IsUrl } from "class-validator";
+import { IsOptional, IsString, Validate } from "class-validator";
+import { RelatedIdentifierMatchesType } from "../utils/related-identifier-validator.util";
 
 export class CreateRelationshipDto {
   @ApiProperty({
     type: String,
     required: true,
-    description: "Persistent identifier of the related dataset.",
+    description:
+      "Identifier of the related entity (e.g. 'https://example.org/datasets/123', '10.1016/j.epsl.2011.11.037', 'arXiv:0706.0001').",
   })
+  @Validate(RelatedIdentifierMatchesType)
   @IsString()
-  readonly pid: string;
+  readonly relatedIdentifier: string;
 
   @ApiPropertyOptional({
     type: String,
-    description: "Relationship between this dataset and the related entity.",
-    default: "is related to",
+    description:
+      "Type of the related identifier (e.g., 'URL', 'DOI', 'arXiv', 'Other').",
+    default: "Other",
+  })
+  @IsString()
+  @IsOptional()
+  readonly relatedIdentifierType?: string;
+
+  @ApiPropertyOptional({
+    type: String,
+    description:
+      "Relationship between this dataset and the related entity (e.g., 'IsReferencedBy', 'IsSupplementTo', 'IsCitedBy').",
+    default: "IsReferencedBy",
   })
   @IsString()
   @IsOptional()
@@ -31,9 +45,9 @@ export class CreateRelationshipDto {
 
   @ApiPropertyOptional({
     type: String,
-    description: "URL to access the related entity, if applicable.",
+    description: "Identifier of the related entity in the external system.",
   })
   @IsString()
-  @IsUrl()
-  readonly url?: string;
+  @IsOptional()
+  readonly targetId?: string;
 }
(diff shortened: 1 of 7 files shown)


src/datasets/schemas/relationship.schema.ts
```diff
@@ -9,18 +9,34 @@ export class RelationshipClass {
...
-  @ApiProperty({
+  @ApiPropertyOptional({
     type: String,
-    required: true,
-    description: "Relationship between this dataset and the related one.",
+    description: "Relationship between this dataset and the related entity.",
+    default: "is related to",
   })
-  @Prop({ type: String, required: true })
+  @Prop({ type: String, default: "is related to" })
```
minottic: very minor comment... since we are taking inspiration from datacite, should we use their terms and fomrat (at least for the default)? Namely, 

```TS
@ApiPropertyOptional({
    type: String,
    description: "Relationship between this dataset and the related entity (e.g., IsReferencedBy, IsSupplementTo, IsDescribedBy).",
    default: "IsReferencedBy",
  })
  @Prop({ type: String, default: "IsReferencedBy" })
```

minottic (approved):
LGTM, thanks! 

I am wondering if it's [relatedIdentifierType](https://datacite-metadata-schema.readthedocs.io/en/4.7/properties/relatedidentifier/#a-relatedidentifiertype) might be valuable? Apparently, e.g. for DOIs, that's useful later for datacite to count citations and might become valuable when we publish the dataset?

src/datasets/schemas/relationship.schema.ts
HayenNico: I'd support this approach - we probably don't want to restrict this to only the terms defined by DataCite (list of definitions [here](https://datacite-metadata-schema.readthedocs.io/en/4.5/properties/relatedidentifier/#b-relationtype)), but we should match that vocabulary's style.

src/datasets/dto/create-relationship.dto.ts
```diff
@@ -10,11 +10,30 @@ export class CreateRelationshipDto {
   @IsString()
   readonly pid: string;
```
HayenNico: Could we consider renaming this to `relatedEntityPid` or something similar? I think that would make it more clear that this is not the pid of this relationship

src/datasets/dto/create-relationship.dto.ts
```diff
@@ -10,11 +10,30 @@ export class CreateRelationshipDto {
...
+  readonly relatedEntityType?: string;
+
+  @ApiPropertyOptional({
+    type: String,
+    description: "URL to access the related entity, if applicable.",
   })
   @IsString()
-  relationship: string;
+  @IsUrl()
+  readonly url?: string;
```
HayenNico: Same as above, maybe this could be named `relatedEntityUrl` instead?

omkar-ethz:
> LGTM, thanks!
> 
> I am wondering if it's [relatedIdentifierType](https://datacite-metadata-schema.readthedocs.io/en/4.7/properties/relatedidentifier/#a-relatedidentifiertype) might be valuable? Apparently, e.g. for DOIs, that's useful later for datacite to count citations and might become valuable when we publish the dataset?

Thanks for the review! I think this is a good idea.
Currently, the assumption was that [relatedIdentifierType](https://datacite-metadata-schema.readthedocs.io/en/4.7/properties/relatedidentifier/#a-relatedidentifiertype) is always URL, so we had the `url` field in the model.

But we can indeed align it more closely with DataCite. So we can remove the url field, as @HayenNico suggests, rename `pid` to `relatedIdentifer`. This would contain the URL if `relatedIdentifierType` is URL, a DOI if `relatedIdentifierType` is doi and so on.

We could add an optional `externalId` field that stores the id of the external entity (e.g. logbookId), and have an index on both externalId and relatedIdentifier?
```json
"relationships": [
  {
    "relationship": "IsReferencedBy",
    "relatedEntityType": "Logbook",
    "externalId": "6895bea625f055bca783dfdd",  // id of the related entity in the external system, opaque for SciCat
    "relatedIdentifier": "https://scilog.development.psi.ch/logbooks/6895bea625f055bca783dfdd",
    "relatedIdentifierType": "URL" 
  }
]
```

*Question:* As we are renaming the fields, do we need a migration script? Or we can do without, as it seems the class wasn't used before

src/datasets/dto/create-relationship.dto.ts
omkar-ethz: Good point, we can change it to `relatedIdentifier` . As per Carlo's suggestion, we're also considering adding a `relatedEntityType` which means this field becomes "polymorphic" based on the the relatedEntityType - please see the detailed comment below

src/datasets/dto/create-relationship.dto.ts
omkar-ethz: Considering getting rid of the field altoghether (comment below)

nitrosx (approved):

nitrosx:
It looks good to me.
Do we want to use the DataCite naming which is long?
Also does this cover if I want to express an arbitrary relationship between two different datasets within the same SciCat instance? If so, how would we do that? Do you already have something in mind?

omkar-ethz:
> It looks good to me. Do we want to use the DataCite naming which is long? Also does this cover if I want to express an arbitrary relationship between two different datasets within the same SciCat instance? If so, how would we do that? Do you already have something in mind?

Thanks a lot! Yes the DataCite names were quite verbose, I dropped the `related-` prefix from the field names.

For SciCat entities, I have proposed using `identifierType: Local`, and then if `entityType` is `Dataset` we can interpret the `identifier` as the PID of a SciCat dataset. Similarly we can use  `identifierType: Local`,  with entityType of `Proposal` / `Instrument` to link to other local entities.

So we could express a derived dataset relationship as:
```json
{
    "identifier": "PID.SAMPLE.PREFIX/psi_ds3",
    "identifierType": "Local",
    "relationship": "IsDerivedFrom",
    "entityType": "Dataset"
}
```

I added this in the documentation of the  `identifierType` field for now, but we can also enforce this in the validation in the future, or use this info in frontend to treat local identifiers specially if needed

PR #2685, opened 2026-04-15
fix: fix bug in optimistic concurrency control implementation

### LINKED ISSUE #2684: Bug in Optimistic Concurrency Control implementation
## Summary
The intended behavior in case of concurrent PATCHes to a resource is described [here](https://github.com/SciCatProject/backend/pull/2138#issuecomment-3250247754). While the current implementation protects from stale writes (i.e. client holds an older representation and attempts a PATCH after an extended period of time, within which the resource has changed on the server), it suffers from the [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use#Description) issue. That is, even if requests 1 and 2 come a second apart, if the DB latency is more than a second, one of the updates will be lost.
To fix this, the updateOne command to the db must be conditional on the resources `updatedAt` being less older than one in the header. i.e. we need to move the unmodified since check from the application level to the DB level.

e.g. in dataset.service.ts:
```ts
async findByIdAndUpdate(
    id: string,
    updateDatasetDto:
      | PartialUpdateDatasetDto
      | PartialUpdateDatasetWithHistoryDto,
    ifUnmodifiedSince?: Date, // **Pass the header date down from the controller**
  ): Promise<DatasetDocument | null> {
...
```

and use it as:
```ts
const patchedDataset = await this.datasetModel
      .findOneAndUpdate(
        {pid: id, updatedAt: {"$lte": ifUnmodifiedSince }}, // **Only update if resource not modified**
        addUpdatedByField(
          updateDatasetDto as UpdateQuery<DatasetDocument>,
          username,
        ),
        { new: true },
      )
      .exec();
    if (!patchedDataset) {
// ** here is the correct place to throw the exception rather than the controller
      throw new PreconditionFailedException(
        `failed if-unmodified-since check due to concurrent update.`
      );
    }
```
## Steps to Reproduce

Please find the script below (run with `node <script.js>`) to trigger this behavior.

## Current Behaviour

In case of multiple concurrent PATCH requests, all except one will fail silently

## Expected Behaviour

In case of multiple concurrent PATCH requests, one succeeds and all others get HTTP 412 Precondition Failed.

omkar-ethz:
## Description

https://github.com/SciCatProject/backend/issues/2684

## Motivation

In the original OCC implementation, the check for whether the resource to be updated has changed after the `if-unmodified-since` timestamp passed by the client was done in the controller, but the actual DB update did not enforce this condition. This meant, two concurrent requests could pass this check in the controller, and update the DB, leading to "lost" or silently failed updates.

By enforcing this precondition in the DB filterQuery, we ensure that exactly one of the concurrent requests succeeds, and all other get 412 Precondition Failed.

## Fixes

* #2684

## Changes:

* moved the OCC check from controller to the atomic `findOneAndUpdate` DB operation
* deleted the now unused util function for checkUnmodifiedSince

## Tests included

- [x] Passing? 

## Documentation

## Summary by Sourcery

Implement atomic optimistic concurrency control across update endpoints by moving conditional modification checks into database findOneAndUpdate operations and standardizing precondition failure handling.

Bug Fixes:
- Ensure concurrent modifications are detected correctly by using conditional findOneAndUpdate filters on updatedAt instead of controller-level timestamp checks.
- Propagate and surface PreconditionFailed exceptions consistently from services to controllers when resources have changed since the client’s If-Unmodified-Since timestamp.

Enhancements:
- Introduce a reusable parseDate utility for safely handling optional and invalid timestamp headers.
- Refine services for datasets, proposals, samples, instruments, attachments, and origdatablocks to validate resource existence before updates and to throw domain-appropriate errors for missing or concurrently modified resources.
- Extend and adjust unit tests to cover the new optimistic concurrency behavior and error propagation across controllers and services.

Tests:
- Add and update unit tests for controllers and services to assert correct usage of findOneAndUpdate with unmodifiedSince filters and proper PreconditionFailedException handling.
- Add tests for the new parseDate utility to verify correct parsing and handling of invalid or missing date strings.

## Summary by Sourcery

Enforce optimistic concurrency control at the database layer for update endpoints and adjust controllers to delegate conditional updates and surface precondition failures consistently.

Bug Fixes:
- Prevent lost updates by enforcing If-Unmodified-Since preconditions directly in atomic findOneAndUpdate operations for datasets, proposals, samples, instruments, attachments, and origdatablocks.
- Ensure controllers propagate PreconditionFailed and NotFound errors thrown by services instead of re-mapping them to generic HTTP exceptions.
- Handle invalid or missing If-Unmodified-Since headers gracefully to avoid incorrect concurrency checks.

Enhancements:
- Introduce a reusable parseDate utility for robust parsing of optional If-Unmodified-Since header values.
- Update services to check resource existence before updates and to throw domain-specific errors when resources are missing or have been concurrently modified.
- Standardize service method naming to findOneAndUpdate for conditional updates across resources and adjust corresponding controller and test usage.
- Extend unit tests to cover optimistic concurrency behavior, error propagation, and new date parsing logic across controllers and services.

Tests:
- Add and update unit tests for controllers and services to verify correct use of findOneAndUpdate with unmodifiedSince filters and PreconditionFailed error handling.
- Add unit tests for the parseDate utility to validate correct behavior for valid, invalid, and missing date strings.


omkar-ethz:
FYI the original authors of the feature as I cannot add you as reviewers: @cfelder @cchndl @joe-baudisch

cfelder:
@omkar-ethz Thanks for reaching out. We'll have a look at this. @FloKo2669 is also involved here.

src/datasets/datasets.controller.spec.ts
```diff
@@ -90,14 +90,23 @@ describe("DatasetsController", () => {
       ).rejects.toThrow(NotFoundException);
     });
 
-    it("should throw PreconditionFailed if header date <= updatedAt", async () => {
+    it("should throw PreconditionFailed if datasetsService throws it", async () => {
```
FloKo2669, line 93: Is there a reason why in this test the _toHaveBeenCalledWith_ assertion for the date propagation is missing compared to the other controller tests? (Also some other tests are missing this) Without it, the test does not add too much value right?

src/instruments/instruments.controller.spec.ts
```diff
@@ -88,7 +103,7 @@ describe("InstrumentsController", () => {
```
FloKo2669: This test does not really check anything anymore right?

FloKo2669 (commented):
Overall the PR looks very good. I only noticed some minor inconsistencies in the tests. (see comments)  I am also wondering if some e2e tests would make sense? (Now the actual logic happens on db level and is not really tested). Let me know what you think, apart from that, I am happy to approve!

src/datasets/datasets.controller.spec.ts
omkar-ethz: i think i just missed it, thank you! added it in this and the "missing header" case, and also in v4 controller tests now.

src/instruments/instruments.controller.spec.ts
omkar-ethz: i think it checks the happy path for when the header is provided, and the service also returns successfully (the test above checks service raising PreconditionFailed, this one is complementary to that). updated the test name to reflect this

omkar-ethz:
> Overall the PR looks very good. I only noticed some minor inconsistencies in the tests. (see comments) I am also wondering if some e2e tests would make sense? (Now the actual logic happens on db level and is not really tested). Let me know what you think, apart from that, I am happy to approve!

thanks for the review!
yes i also think e2e tests were warranted as it's not simply a controller behavior anymore - added them now!

Junjiequan (approved):
looks good to me

src/datasets/datasets.service.ts
```diff
@@ -508,27 +509,26 @@ export class DatasetsService {
...
-    }
-
     const username = (this.request.user as JWTUser).username;
 
     // NOTE: When doing findByIdAndUpdate in mongoose it does reset the subdocuments to default values if no value is provided
     // https://stackoverflow.com/questions/57324321/mongoose-overwriting-data-in-mongodb-with-default-values-in-subdocuments
+    const queryFilter: FilterQuery<DatasetDocument> = { pid: id };
+    if (unmodifiedSince !== undefined) {
+      queryFilter.updatedAt = { $lte: unmodifiedSince };
+    }
```
cfelder: can this filter part be refactored in it's own function to reduce code duplication?

cfelder:
One remark inline about refactoring to reduce code duplication.

Otherwise /lgtm

and first tests on our dev deployment are fine.

omkar-ethz:
> One remark inline about refactoring to reduce code duplication.
> 
> Otherwise /lgtm
> 
> and first tests on our dev deployment are fine.

@cfelder thanks! I don't see your comment unfortunately, perhpas you forgot to publish your review?

src/datasets/datasets.service.ts
omkar-ethz: good point, refactored all six services to call a `withOCCFilter` utility function

PR #2697, opened 2026-04-20
refactor(metadatakeys): merge duplicate keys across datasets and track group references

Junjiequan:
## Description
MetadataKeys are now deduplicated, instead of one entry per key per dataset, a single entry is shared across all datasets containing the same key, with `userGroups`, `usageCount`, and `userGroupCounts` merged.

### userGroupCounts
New internal field that tracks how many datasets per group reference this key:
```js
dataset1: {ownerGroup:"group1", userGroups:["group2"]},
dataset2: {ownerGroup:"group1"}
dataset3: {ownerGroup:"group1"}

"userGroupCounts": { "group-1": 3, "group-2": 1 },
"userGroups": ["group-1", "group-2"],
"usageCount": 3
```
When a group's count hits 0 it is removed from userGroups. When usageCount hits 0 the entry is deleted.

## Changes
- Parent document (`dataset`, `proposal`) permission changes (`ownerGroup`, `accessGroups`) update userGroups and userGroupCounts accordingly
- If any parent document has `isPublished: true`, the metadata key will have `isPublished:true` as well and is visible to all users
- Migration script included with documentation explaining the aggregation pipeline to populate MetadataKeys from existing Dataset documents — must be run manually before deploying (duration depends on dataset count and scientificMetadata count per dataset). Keys used by more than 1000 datasets are automatically made public.

## Fixes

* Bug fixed (#X)

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Track metadata keys at a global level across datasets and other entities, aggregating usage and access control by key rather than per-source document.

New Features:
- Expose metadata key usage counts and per-group reference counts, enabling queries over how often and by whom metadata keys are used.
- Add a migration to backfill the MetadataKeys collection from existing dataset scientificMetadata, merging keys across datasets and computing group usage.
- Extend metadata key search capabilities with filtering on humanReadableName, regex matching, field projection, and pagination semantics.

Enhancements:
- Refine metadata key aggregation to use sourceType plus key and humanReadableName as the identifier, allowing the same key to be merged across multiple datasets and groups.
- Update datasets, samples, proposals, and instruments services to emit unified MetadataSourceDoc structures that carry resolved userGroups and drive consistent metadata key bookkeeping.
- Strengthen access-control semantics so metadata keys respect publication status and user group visibility while avoiding exposure of group-private keys to unauthorized users.
- Improve documentation for the MetadataKeys module and its migration, including examples of filters and an architectural overview of synchronization behavior.

Documentation:
- Expand developer documentation for the MetadataKeys module to describe the new schema, synchronization flow, filter options, and initial migration requirements.
- Document the new migration that rebuilds the MetadataKeys collection from dataset scientificMetadata, including pipeline details and operational guidance.

Tests:
- Add extensive unit tests for MetadataKeysService covering aggregation pipelines, insert/delete/replace behaviors, and group/publish state transitions.
- Add end-to-end tests around the /metadatakeys API to validate access control, search behavior, pagination, projection, and usageCount semantics across different user roles.


migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.js
```diff
@@ -0,0 +1,313 @@
+const SOURCE_COLLECTIONS = ["Dataset"];
+
+function buildPipeline(sourceType) {
```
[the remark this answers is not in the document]
Junjiequan: The pipeline is documented with input/output examples for each steps with detailed expalanation. I don't think seperate the pipe into multiple chunks would help without clear explanation anyway.

src/metadata-keys/metadatakeys.service.ts
```diff
@@ -79,59 +92,254 @@ export class MetadataKeysService {
...
+        ...(addedGroups.length > 0 && {
+          $addToSet: { userGroups: { $each: addedGroups } },
+        }),
+        $inc: {
+          ...(addedGroups.length > 0 && this.groupCountDeltas(addedGroups, 1)),
+          ...(removedGroups.length > 0 &&
+            this.groupCountDeltas(removedGroups, -1)),
+        },
+        ...(publishedFlippedOn && { $set: { isPublished: true } }),
+      });
```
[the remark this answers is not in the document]
Junjiequan: > Starting in MongoDB 5.0, [mongod](https://www.mongodb.com/docs/manual/reference/program/mongod/#mongodb-binary-bin.mongod) no longer raises an error when you use an update operator like [$inc](https://www.mongodb.com/docs/manual/reference/operator/update/inc/#mongodb-update-up.-inc) with an empty operand expression ( { } ). An empty update results in no changes and no [oplog](https://www.mongodb.com/docs/manual/reference/glossary/#std-term-oplog) entry is created (meaning that the operation is a no-op).

https://www.mongodb.com/docs/manual/reference/operator/update/inc/

src/metadata-keys/metadatakeys.service.ts
```diff
@@ -79,59 +92,254 @@ export class MetadataKeysService {
...
+  private async updateSharedKeys(
+    sharedKeys: string[],
+    oldDoc: MetadataSourceDoc,
+    newDoc: MetadataSourceDoc,
+  ): Promise<void> {
+    const { sourceType } = newDoc;
+    const ids = sharedKeys.map((k) => {
+      const humanReadableName =
+        (newDoc.metadata[k] as ScientificMetadataEntry)?.human_name ?? "";
+      return this.buildId(sourceType, k, humanReadableName);
```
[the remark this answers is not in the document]
Junjiequan: addressed

test/MetadataKeys.js
```diff
@@ -2,173 +2,583 @@
...
+          key: "rename_test_key",
+          humanReadableName: "Renamed Name",
+        },
+      },
+      accessTokenAdmin,
+    );
+    afterNew.body.should.have.lengthOf(1);
+    afterNew.body[0].usageCount.should.equal(1);
+
+    await deleteDataset(pid);
```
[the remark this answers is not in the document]
Junjiequan: There's no need to test accessGroups explicitly, since ownerGroup and accessGroups will be merged into userGroups before passed to the metadataKeys service

```js
  private createMetadataKeysInstance(
    doc: UpdateQuery<DatasetDocument>,
  ): MetadataSourceDoc {
    const source: MetadataSourceDoc = {
      sourceType: this.datasetModel.collection.name,
      userGroups: Array.from(
        new Set([doc.ownerGroup, ...(doc.accessGroups ?? [])].filter(Boolean)),
      ),
      isPublished: doc.isPublished || false,
      metadata: doc.scientificMetadata ?? {},
    };
    return source;
  }
```

docs/developer-guide/metadatakeys-module.md
```diff
@@ -17,74 +17,179 @@ The previous implementation in the Datasets service lacked a permission-based fi
 - Stability: Crashes occurred when retrieval limits were missing or improperly configured.
 - Risks: Users could see metadata keys they did not have permissions to access.
 
+---
```
rkweehinzmann, line 20: Could you explain in which cases this can happen that a user sees metadata keys they do not have permissions for?

src/metadata-keys/metadatakeys.service.ts
```diff
@@ -79,59 +92,266 @@ export class MetadataKeysService {
...
 
     Logger.log(
-      `Created ${docs.length} MetadataKeys from source ${doc.sourceType} with ID ${doc.sourceId}`,
+      `Decremented or deleted MetadataKeys for ${sourceType}: ${keys.join(", ")}`,
     );
+  }
+  /**
+   * Called when a dataset is updated.
+   *
+   * Diffs the old and new metadata to produce three disjoint key sets:
```
rkweehinzmann: Who will know what old and new metadata is in 2 months or 2 years (or even now)? Could you add some short description of what is called old and what new, please? Thank you.

rkweehinzmann (changes requested):
Thanks for the reimplementation of this service!

One other question:
Are nested scientific metadata values handled that have themselves nested values?

Thank you.

docs/developer-guide/metadatakeys-module.md
Junjiequan: All the metadata keys returned from `/api/v3/datasets/metadataKeys` are not filterd by permission

src/metadata-keys/metadatakeys.service.ts
Junjiequan: old metadata is the metadata befor change, new metadata is the metadata after change

so if I change metadata from `{"metadataKey1":"text"}` to `{"meatadataKey2":"text"}`
the former is old metadata, latter is the new metadata

docs/developer-guide/metadatakeys-module.md
rkweehinzmann: ok, but what does it mean, eg if `/api/v4/datasets/metadatakeys` is used they will always be filtered by permissions? From your initial description I thought it can have to do with setting limits wrongly ?

src/metadata-keys/metadatakeys.service.ts
rkweehinzmann: Did you want me to see that `metadataKey1` and `metadataKey2` are different names, but both refer to the same value (which is the actual meta data and did not change)? I meant if you call something old or new, you're referring of to an old/new way of handling things. What was the old way, what the new way? Apologies, probably too obvious for the programmer!

docs/developer-guide/metadatakeys-module.md
Junjiequan: The new endpoint `/api/v4/metadataKeys/findAll?{query}` will always return filtered metadatakeys by permissions. 

I'm not sure I understand your second quesion

src/metadata-keys/metadatakeys.service.ts
Junjiequan: Just to clarify what it is used for.

Consider a data provider A and a data receiver B.
Initially, A sends `{key1}[old]` to B, and B stores it. Later, A renames the key from `{key1}[old]` to `{key2}[new]`. If A only sends `{key2}[new]`, B ends up holding both `{key1}[old]` and `{key2}[new]` — even though the old key no longer exists on A's side. This creates a sync inconsistency.
To prevent this, when A changes `{key1}[old]` to `{key2}[new]`, it must provide both the old and new state to B, effectively saying:

"Here is what changed — discard [old] and replace it with [new]."

This is exactly why replaceManyFromSource requires both oldDoc (pre-update) and newDoc (post-update) — so B can recompute the difference and stay in sync with A.

Junjiequan: I also updated the comment to state that, old is before change and new is after change. There's many terms to describe it, but old and new is probably is the shortest term and it should be self-explanatory

docs/developer-guide/migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.md
```diff
@@ -0,0 +1,311 @@
+# 20260417145401 — Sync Dataset scientificMetadata to MetadataKeys
+
+## What this migration does
+
+Rebuilds the `MetadataKeys` collection from scratch by scanning all `Dataset` documents and extracting every key found in `scientificMetadata`.
+
+Each unique `(sourceType, key, humanReadableName)` combination becomes one `MetadataKey` document. If the same key appears across multiple datasets, their `userGroups` and counts are merged into a single entry.
```
minottic, line 7: why do we need humanReadableName? Isn't `sourceType, key` already unique?

docs/developer-guide/migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+---
+
+## Output shape
+
+```js
+// MetadataKey document
+{
+  _id: "Dataset_temperature_Temperature",  // ${sourceType}_${key}_${humanReadableName}
```
minottic: as discussed offline, I would keep uuid as ID and eventually add an additional field with the concat. This will make easier to have multiple documents with the same combination in case we will ever want to split big documents

docs/developer-guide/migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.md
```diff
@@ -0,0 +1,311 @@
...
+  key: "temperature",
+  humanReadableName: "Temperature",
+  sourceType: "Dataset",
+  isPublished: true,
+  usageCount: 2,
```
minottic, line 52: why do you need this? Isn't userGroupCounts enough?

migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.js
minottic, line 1: what about this:

```TS

function buildPipeline(sourceType) {
  return [
    {
      $project: {
        isPublished: 1,
        groups: { $setUnion: [["$ownerGroup"], { $ifNull: ["$accessGroups", []] }] },
        meta: { $objectToArray: "$scientificMetadata" },
      }
    },
    { $unwind: "$meta" },
    { $unwind: "$groups" },
    { $match: { groups: { $ne: null, $ne: "" } } },
    {
      $group: {
        _id: { 
          k: "$meta.k", 
          g: "$groups", 
          n: { $ifNull: ["$meta.v.human_name", ""] } 
        },
        count: { $sum: 1 },
        isPublished: { $max: "$isPublished" }
      }
    }
  ];
}

module.exports = {
  async up(db) {
    const start = Date.now();
    const elapsed = () => `${((Date.now() - start) / 1000).toFixed(1)}s`;

    await db.collection("MetadataKeys").deleteMany({});
    await db.collection("MetadataKeys").createIndex({ metaKeyId: 1 }, { unique: true });

    for (const collectionName of SOURCE_COLLECTIONS) {
      console.log(`[${elapsed()}] Starting stream for ${collectionName}...`);

      // 1. Open a single cursor for the entire collection
      const cursor = db.collection(collectionName).aggregate(buildPipeline(collectionName), { 
        allowDiskUse: true 
      });

      let operations = [];
      let totalProcessed = 0;

      // 2. Iterate over the cursor without toArray()
      for await (const doc of cursor) {
        const metaKeyId = `${collectionName}_${doc._id.k}_${doc._id.n}`;
        
        operations.push({
          updateOne: {
            filter: { metaKeyId },
            update: {
              $set: {
                key: doc._id.k,
                humanReadableName: doc._id.n,
                sourceType: collectionName,
                updatedBy: "migration",
                updatedAt: new Date()
              },
              $setOnInsert: { createdAt: new Date(), createdBy: "migration" },
              $max: { isPublished: doc.isPublished },
              $addToSet: { userGroups: doc._id.g },
              $inc: { 
                usageCount: 1,
                [`userGroupCounts.${doc._id.g.replace(/\./g, '_')}`]: doc.count 
              }
            },
            upsert: true
          }
        });

        // 3. Periodic Flush: execute bulk write every BATCH_SIZE items
        if (operations.length >= BATCH_SIZE) {
          await db.collection("MetadataKeys").bulkWrite(operations);
          totalProcessed += operations.length;
          console.log(`[${elapsed()}] ${collectionName}: Flushed ${totalProcessed.toLocaleString()} keys...`);
          operations = []; // Clear the bucket
        }
      }

      // 4. Final flush for any remaining operations
      if (operations.length > 0) {
        await db.collection("MetadataKeys").bulkWrite(operations);
        totalProcessed += operations.length;
      }

      console.log(`[${elapsed()}] ✅ ${collectionName} complete. Total keys: ${totalProcessed}`);
    }

    // Clean up: switch metaKeyId to the official _id
    await db.collection("MetadataKeys").updateMany({}, [
      { $set: { id: "$metaKeyId" } }, 
      { $unset: "metaKeyId" }
    ]);
  }
};
```

src/metadata-keys/schemas/metadatakey.schema.ts
```diff
@@ -76,28 +76,40 @@ export class MetadataKeyClass extends QueryableClass {
...
-  @Prop({ type: Boolean, required: true })
-  isPublished: boolean;
+  @Prop({ type: Map, of: Number, default: {} })
+  userGroupCounts: Map<string, number>;
 }
 
 export const MetadataKeySchema = SchemaFactory.createForClass(MetadataKeyClass);
 
-MetadataKeySchema.index({ sourceType: 1, sourceId: 1 });
+MetadataKeySchema.index({ sourceType: 1, isPublished: 1, key: 1 });
```
minottic: why do we need so many indexes?

src/samples/samples.service.ts
```diff
@@ -45,10 +45,10 @@ export class SamplesService {
     doc: UpdateQuery<SampleDocument>,
   ): MetadataSourceDoc {
     const source: MetadataSourceDoc = {
```
minottic: this snippet is replicated in all collections, worth making a util function that takes doc and name as input?

src/metadata-keys/metadatakeys.service.ts
minottic, line 1: what about something like this:

```TS
@Injectable({ scope: Scope.REQUEST })
export class MetadataKeysService {
  constructor(
    @InjectModel(MetadataKeyClass.name)
    private metadataKeyModel: Model<MetadataKeyDocument>,
  ) {}

  // Centralized logic for changing counts
  private async adjustCounts(doc: MetadataSourceDoc, delta: 1 | -1): Promise<void> {
    const { sourceType, userGroups, metadata } = doc;
    const ops = Object.entries(metadata).map(([key, entry]) => {
      const humanName = (entry as ScientificMetadataEntry).human_name ?? "";
      const metaKeyId = `${sourceType}_${key}_${humanName}`;

      return {
        updateOne: {
          filter: { metaKeyId },
          update: {
            $set: { key, sourceType, humanReadableName: humanName },
            $inc: {
              usageCount: delta,
              ...Object.fromEntries(userGroups.map(g => [`userGroupCounts.${g.replace(/\./g, '_')}`, delta]))
            },
            ...(delta === 1 && { $max: { isPublished: doc.isPublished } }),
            $setOnInsert: { createdAt: new Date() }
          },
          upsert: delta === 1, // Only upsert when adding
        }
      };
    });

    if (ops.length > 0) {
      await this.metadataKeyModel.bulkWrite(ops);
      
      // Cleanup: Only run when decrementing
      if (delta === -1) {
        await this.metadataKeyModel.deleteMany({ usageCount: { $lte: 0 } });
      }
    }
  }

  async insertManyFromSource(doc: MetadataSourceDoc) {
    await this.adjustCounts(doc, 1);
  }

  async deleteMany(doc: MetadataSourceDoc) {
    await this.adjustCounts(doc, -1);
  }

  async replaceManyFromSource(oldDoc: MetadataSourceDoc, newDoc: MetadataSourceDoc) {
    // Standard "out with the old, in with the new"
    // MongoDB handles the interleaving safely
    await this.deleteMany(oldDoc);
    await this.insertManyFromSource(newDoc);
  }
}
```

minottic: (to be changed with the right key inplace of this)

```TS
      const metaKeyId = `${sourceType}_${key}_${humanName}`;
```

migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.js
minottic: (to be changed with the right key inplace of this)

```TS
      const metaKeyId = `${sourceType}_${key}_${humanName}`;
```

Junjiequan: I would avoid to use cursor for the migration script, it will be extremly slow to retrive data store in the memory and process them. since each doc transferred into the mongoDB needs to be encoded

Example of comparison of the performances  using cursor vs aggregate + $mege to handle everything:

image

This testing was using 20k datasets with 5~10 ScientificMetadata per datasets

minottic: but doesn't that depend largely on the BATCH_SIZE? The thing is that the current migration is hardly maintainable, it's extremely complex if we need to change anything

Junjiequan: In this case, the bigger BATCH_SIZE the faster with aggregate + $merge,  .

I would expect this migration script not to be touched and any further maintainance should be a new migration patch on top of this migration. 
But TBH, I don't really mind the script to be slow, as long as it's robust, since we need it to run only once. So, if you suggest that readability should be more prioratized in this case, Im fine with it. 
I also felt like current implementation is hard to read and that's why I included a doc to explain each step.

src/metadata-keys/metadatakeys.service.ts
Junjiequan: I think the code you suggested here looks very good, I will test it locally and see if there's any missing edge cases

migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.js
Junjiequan: > I would expect this migration script not to be touched 

By this I meant once the migration is merged I expect it not to be changed, not that im saying current code.

docs/developer-guide/migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.md
Junjiequan: Same key might have completely different humanReadableName, so the uniqueness is based on sourceType+key+humanReadableName

docs/developer-guide/migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.md
Junjiequan: I will update the documentation later, I changed the id to use UUID but forgot to change doc here.

docs/developer-guide/migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.md
Junjiequan: This avoids edge cases where a dataset has no groups but isPublished: true (unlikely, but ownerGroup is not a required field in OwnableDto). 
Additionally, userGroupCounts cannot replace usageCount because a dataset can belong to multiple groups, whereas usageCount counts each dataset only once.

src/metadata-keys/schemas/metadatakey.schema.ts
Junjiequan: The indexs are to handle queries from unauthenticated users and authenticated users.

for authenticated users the query will be 
```js
MetadataKey.find({
  sourceType: 'dataset',
  $and: [
    {
      $or: [
        { isPublished: true },
        { userGroups: { $in: user.groups } }
      ]
    },
    {
      $or: [
        { key: job_id },
        { humanReadableName: JOB_ID }
      ]
    }
  ]
});
```

for unauthenticated users the query will be:
```js
MetadataKey.find({
  sourceType: 'dataset',
  $and: [
    {
      $or: [
        { isPublished: true },
      ]
    },
    {
      $or: [
        { key: job_id },
        { humanReadableName: JOB_ID }
      ]
    }
  ]
});
```

src/samples/samples.service.ts
Junjiequan: Yes, maybe it's worth to have a shared function here

docs/developer-guide/migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.md
Junjiequan: docs updated

src/samples/samples.service.ts
Junjiequan: addressed in https://github.com/SciCatProject/backend/pull/2697/commits/1d4f99ad2f7448f0f0eddcf0df42b0fc3570ef46

src/metadata-keys/metadatakeys.service.ts
Junjiequan: addressed in https://github.com/SciCatProject/backend/pull/2697/changes/32f95400b1fda14b2d4e9c1fc6dfca8052895ec2

migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.js
Junjiequan: The readability is improved by removing some unnecessary fields, such as check existing fields then overwrite logic, which is not needed, since we wipe everything when migration starts. 
Documentation is also updated with better explanation for aggregation pipe

src/metadata-keys/schemas/metadatakey.schema.ts
minottic: mmm I am a bit concerned by this, every index is space the DB allocates and this collection is quite big already.

Junjiequan: I see your concerns, but considering datasets and other schemas with `schema.index({ "$**": "text" })` running without any issues, I think the metadata schema with 4 indexes should be fine for a very long time, especially considering each metadata doc is very small.

nitrosx (approved):
The PR looks good to me.
We might be able to improve the aggregation pipeline and performance, but we can address those issue in a following PR

migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.js
minottic: thanks. I still find the other approach superior in terms of maintainability, but happy with this one too!

src/metadata-keys/schemas/metadatakey.schema.ts
minottic: but since we have few records per (key, sourceType) isn't it enough to have these indexes only?

```TS
{ sourceType: 1, key: 1 }
{ sourceType: 1, humanReadableName: 1 }
```

Junjiequan: Just did some research, I think you are right, I will change it

Junjiequan: addressed in [3f98ce7](https://github.com/SciCatProject/backend/pull/2697/commits/3f98ce777dd5155a97e02b760fd3338357174d42)

minottic (approved):
thanks!

migrations/20260420145401-sync-dataset-scientificMetadata-to-metadatakeys.js
```diff
@@ -0,0 +1,204 @@
...
+    {
+      $addFields: {
+        metaKeyId: "$_id",
+        generatedId: {
+          $function: {
+            body: "function() { return UUID().toString().replace('UUID(\"', '').replace('\")', ''); }",
+            args: [],
+            lang: "js",
+          },
+        },
```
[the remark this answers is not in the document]
Junjiequan: it's been tested and UUID works

src/datasets/datasets.service.ts
```diff
@@ -62,19 +63,15 @@ import {
...
-  MetadataSourceDoc,
-} from "src/metadata-keys/metadatakeys.service";
+import { MetadataKeysService } from "src/metadata-keys/metadatakeys.service";
 import { OpensearchService } from "src/opensearch/opensearch.service";
-import type { IndexSettings } from "@opensearch-project/opensearch/api/_types/indices._common";
-import type { TypeMapping } from "@opensearch-project/opensearch/api/_types/_common.mapping";
-import { BulkStats } from "@opensearch-project/opensearch/lib/Helpers";
+import { BulkStats } from "@opensearch-project/opensearch/lib/Helpers.js";
+import { IndexSettings } from "@opensearch-project/opensearch/api/_types/indices._common.js";
+import { TypeMapping } from "@opensearch-project/opensearch/api/_types/_common.mapping.js";
```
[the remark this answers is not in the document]
Junjiequan: not related to this PR

PR #2699, opened 2026-04-24
fix(attachmentV4): allow elevated users to access attchment endpoint

Junjiequan:
## Description
This PR adds back the missing CASL ability check for elevated attachment group users to access attachment endpoints.
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Bug Fixes:
- Allow elevated attachment group users to access attachment create, update, and delete endpoints via CASL ability checks.


jl-wynen:
Are there tests for these endpoints with normal user accounts? It seems like you should have caught this bug without my help.

Junjiequan:
@jl-wynen 
All the attachmentV4 endpoint tests were using admin account, I think I forgot to add tests for privileged accounts
Allowing normal users to upload attachments needs to be discussed, I will get back to you once there's decision.

src/casl/casl-ability.factory.ts
```diff
@@ -423,6 +423,29 @@ export class CaslAbilityFactory {
...
+          this.accessGroups?.attachment.includes(g),
+        ) ||
+        this.accessGroups?.attachment.includes("#all")
+      ) {
+        // -------------------------------------
+        // users belonging to any of the group listed in ATTACHMENT_GROUPS
+        //
+        can(Action.AttachmentCreateEndpoint, Attachment);
+        can(Action.AttachmentUpdateEndpoint, Attachment);
+        can(Action.AttachmentDeleteEndpoint, Attachment);
```
nitrosx, line 448: Shouldn't this be `cannot`?
As it is now, `attachment privileged` and `attachment` do not have difference.

nitrosx (changes requested):

src/casl/casl-ability.factory.ts
Junjiequan: Endpoint wise there shouldnt be differences, both group users should be able to access the endpoint. The difference should be in the instance level

nitrosx: OK.
I will consolidate the two `if` branches.

Junjiequan: I suggeste we open a new PR and consolidates all of the `if`s that has same permission logic. Instead of just do it for the attachmentEndpointAccess

nitrosx (approved):

PR #2712, opened 2026-04-29
fix: allow retrieve of public and owned ds

minottic:
## Description
The computation of job.ownerGroup, error thrown if empty, suffered from having
jobs trying to access public and owned datasets because public datasets might
be not owned by the user submitting the job. It also includes a refactor to
simplify maintenance as well as more verbose error messages

## Fixes

* public datasets excluded from intersection computation

## Changes:

* refactor
* more verbose error messages/cleaner error codes
* speed up dataset retrieval with a findAll rather than looping on all datasets

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Adjust job creation mapping and validation to correctly handle public and owned datasets while improving ownership checks and error reporting.

Bug Fixes:
- Exclude published datasets from group intersection when resolving ownerGroup for #datasetAccess jobs so mixed public and owned datasets can be used.
- Ensure ownerGroup resolution for #datasetOwner jobs validates that all datasets share the same owner group and fails with a clear error otherwise.
- Align HTTP status codes for invalid ownership, group membership, and anonymous job email validation with semantic 403 and 422 responses instead of generic 400s.

Enhancements:
- Refactor the v3 job creation interceptor into smaller helpers for building job parameters, owner user, and owner group.
- Relax immutability on CreateJobDto ownership fields so they can be populated server-side before persistence.
- Improve error messages around job ownership and group validation for clearer diagnostics.

Tests:
- Extend job authorization tests to cover mixed public and owned datasets, differing owner groups, and updated error status codes and messages.
- Add new test datasets to verify published datasets do not block dataset access jobs via group intersection.

### DIFF
--- a/src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
+++ b/src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
@@ -15,6 +15,12 @@ import { UsersService } from "src/users/users.service";
 import { DatasetsService } from "src/datasets/datasets.service";
 import { JobConfigService } from "src/config/job-config/jobconfig.service";
 import { JobsControllerUtils } from "src/jobs/jobs.controller.utils";
+import { omit } from "lodash";
+import { CreateJobAuth } from "../types/jobs-auth.enum";
+import { DatasetDocument } from "src/datasets/schemas/dataset.schema";
+import { ConfigService } from "@nestjs/config";
+import { AccessGroupsType } from "src/config/configuration";
+import { FilterQuery } from "mongoose";
 
 interface JobParams {
   datasetList: DatasetListDto[];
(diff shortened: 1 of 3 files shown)


src/jobs/interceptors/create-job-v3-mapping.interceptor.ts
```diff
@@ -47,115 +59,96 @@ export class CreateJobV3MappingInterceptor implements NestInterceptor {
...
+    }
+    const datasets = await this.datasetsService.findAll(datasetsFilter);
+    if (datasets.length !== datasetList.length) return undefined;
+    if (datasets.length === 0) return undefined;
+    if (
+      jobConfigCreateAuth === CreateJobAuth.DatasetAccess &&
+      datasets.every((dataset) => dataset?.isPublished)
+    )
+      return jobUserCurrentGroups?.[0];
+    const nonPublishedDatasets = datasets.filter(
```
Junjiequan, line 149: I'm not very familiar with the job concept, but this logic looks slightly off to me.
If each job requires ownerGroup, instead of picking arbitary one from userGroups array, how about forcing user to provide one? if possible

minottic: I completely agree, though this is mostly a refactor (logic was [here](https://github.com/SciCatProject/backend/pull/2712/changes#diff-2c5665d678dccbe64d2186fc6bd0d1cd5fab516d09e7a3a08f63faa6d57218f1L129) already), making sure the existing tests don't break (TBH I am also not super familiar with JOBS internals, so my strategy is to keep existing logic while fixing bugs and keeping tests passing). The olny addition to this PR Is that before my change the user was given 400 when trying to submit a job with public and owned DS because the intersection of ownerGroups from public ones with the users' might be empty, so they are now excluded when computing the intersection

Junjiequan (approved):

PR #2713, opened 2026-05-01
fix: rewrite instance access func with proper auth

minottic:
## Description
It changes from a model based on job payload fields on creation to solely relying on user groups. This fixes the bug that prevented users to submit mixed ownerGroup datasets (part of the user). It also refactors the god function to smaller units, split by responsability

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Refine job creation authorization to rely on user group memberships, enforce dataset access consistently, and clarify error handling for job submission failures.

Bug Fixes:
- Allow job submission with datasets owned by multiple groups as long as the submitting user belongs to all relevant groups.
- Correct HTTP status codes and error messages for invalid or unauthorized job creation scenarios.
- Fix access group configuration parsing so privileged and delete-job groups are correctly interpreted from comma-separated environment variables.

Enhancements:
- Refactor job creation authorization logic into smaller, focused helper methods for initializing jobs, determining user privileges, and validating dataset access.

Tests:
- Update existing job authorization tests to match the new error semantics and status codes, and add coverage for mixed-owner dataset scenarios with multi-group users.

### DIFF
--- a/src/config/configuration.ts
+++ b/src/config/configuration.ts
@@ -277,9 +277,13 @@ const configuration = () => {
       attachmentPrivileged: attachmentPrivilegedGroups
         .split(",")
         .map((v) => v.trim()),
-      createJobPrivileged: createJobPrivilegedGroups,
-      updateJobPrivileged: updateJobPrivilegedGroups,
-      deleteJob: deleteJobGroups,
+      createJobPrivileged: createJobPrivilegedGroups
+        .split(",")
+        .map((v) => v.trim()),
+      updateJobPrivileged: updateJobPrivilegedGroups
+        .split(",")
+        .map((v) => v.trim()),
+      deleteJob: deleteJobGroups.split(",").map((v) => v.trim()),
     },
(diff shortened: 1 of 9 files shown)


Junjiequan (commented):
It seems that the jobs system has its own auth managing logic per user, instead of solely relying on CASL ability, which we should avoid. Once we start doing that, we lose the point of using CASL ability as the single source of truth for access management. 
This is not related to this PR, since the old one was already doing that. I think at some point we will have to deal with the tech debt of having a separated access logic implementation. But this refactor seems to be mitigating that cost.

Junjiequan (approved):
It's hard to thoroughly understand the entire changes since I'm not very familiar with the whole job system, but it seems to be mainly refactoring and breaking down big functions into smaller pieces. Additionally, the job system has quite decent test coverage, so as long as the current tests pass I'm happy for it to be merged.

minottic:
>  think at some point we will have to deal with the tech debt of having a separated access logic implementation

I could not agree more

PR #2714, opened 2026-05-04
BREAKING CHANGE: simplify job validation and reduce db calls

minottic:
## Description
Refactor job validation by reducing DB calls and using class-transformer validation. Also use more specific error codes.

Depends on #2713 

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Refactor job validation logic to rely on DTO validation and more specific HTTP errors while reducing database usage for dataset and file checks.

Bug Fixes:
- Ensure dataset and file validation errors return HTTP 422 Unprocessable Entity instead of generic bad request responses.

Enhancements:
- Leverage class-transformer and class-validator to validate dataset list DTOs instead of manual checks.
- Optimize dataset PID and file existence checks by reducing and consolidating database queries and projecting only required fields.

Tests:
- Update job-related integration tests to expect HTTP 422 status codes for invalid job requests.

### DIFF
--- a/src/jobs/jobs.controller.utils.ts
+++ b/src/jobs/jobs.controller.utils.ts
@@ -43,6 +43,8 @@ import {
 import { toObject } from "src/config/job-config/actions/actionutils";
 import { loadDatasets } from "src/config/job-config/actions/actionutils";
 import { DatasetClass } from "src/datasets/schemas/dataset.schema";
+import { validate } from "class-validator";
+import { plainToInstance } from "class-transformer";
 
 @Injectable()
 export class JobsControllerUtils {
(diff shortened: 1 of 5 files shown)


src/jobs/jobs.controller.utils.ts
```diff
@@ -81,63 +83,30 @@ export class JobsControllerUtils {
...
-    }
-    if (datasetList.length == 0) {
-      throw new HttpException(
-        {
-          status: HttpStatus.BAD_REQUEST,
-          message: "List of passed datasets is empty.",
-        },
-        HttpStatus.BAD_REQUEST,
+    if (!Array.isArray(datasetList))
+      throw new UnprocessableEntityException("Invalid dataset list");
```
pozsa: `UnprocessableEntityException` is throwing `422`, instead of the old `BAD_REQUEST` `400`. That's a breaking api change.

minottic: what's your suggestion here, keep the generic error code or improve it and introduce a breaking change?

minottic: actually, I can change the PR title to be BREAKING so the semantic release will pick it up at the next release

pozsa: My opinion, specific error code is better and breaking change is fine as long as semantic versioning is respected. i.e. this should be a major version bump if the API spec is changed.

src/jobs/jobs.controller.utils.ts
```diff
@@ -81,63 +83,30 @@ export class JobsControllerUtils {
...
-          },
-          HttpStatus.BAD_REQUEST,
-        );
-      }
+    const datasetListDtos: DatasetListDto[] = plainToInstance(
+      DatasetListDto,
+      datasetList,
+    );
+    const nestedErrors = await Promise.all(
+      datasetListDtos.map((dto) => validate(dto)),
```
pozsa: Old code rejected items with keys outside `{pid, files}` via `keys.length !== 2 || !keys.every(...)`. The new path:

  ```ts
  const datasetListDtos = plainToInstance(DatasetListDto, datasetList);
  const nestedErrors = await Promise.all(datasetListDtos.map((dto) => validate(dto)));
  ```

  `plainToInstance` without `excludeExtraneousValues: true` copies everything, and `validate()` without `{ whitelist: true, forbidNonWhitelisted: true }` allows unknown properties. So `{
  pid, files, anyExtraField: ... }` is now accepted where it was previously rejected.

This should preserve the current behaviour:
  ```ts
validate(dto, { whitelist: true, forbidNonWhitelisted: true })
  ```

minottic: nice catch! changing it

pozsa:
Inconsistent error body shapes across the three throws.

  - `validateDatasetList` (validation path): `new UnprocessableEntityException("Invalid dataset list. " + JSON.stringify(minimalErrors))` — sentence-with-embedded-JSON; clients have to
  substring-parse `.message`.
  - `checkDatasetPids`: plain string.
  - `checkDatasetFiles`: `{ message, error: JSON.stringify(checkResults) }`.

  Pick one shape — preferably structured — and use it everywhere.
  
Example:
  ```ts
  throw new UnprocessableEntityException({
    message: "Invalid dataset list",
    errors: minimalErrors,
  });
  ```

minottic:
> checkDatasetPids: plain string.
> checkDatasetFiles: { message, error: JSON.stringify(checkResults) }.

thanks! these 2 I think are essentially the same, (the first one would have the default `"error": "Unprocessable Entity"`). Changing the JSON one though, thanks!

pozsa (approved):
Don't forget to mark the breaking change in the title/body.

Junjiequan (approved):
Looks fine to me.
If you squash merge please remember to include the breaking change PR title in the body to trigger major bump not just in the title

PR #2715, opened 2026-05-04
fix: use $text and $search for published data search and use correct count access check

abdimo101:
## Description
This PR updates the `AppendFieldsToFilterPipe` used for published data filter to use $text and $search so the search would work and uses correct count access check.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Bug Fixes:
- Ensure published data search requests using a `text` field correctly map to `$text` / `$search` queries.

### DIFF
--- a/src/published-data/pipes/filter.pipe.ts
+++ b/src/published-data/pipes/filter.pipe.ts
@@ -18,7 +18,7 @@ class AppendFieldsToFilterPipe implements PipeTransform {
     const filter = value.filter ?? {};
 
     if ("text" in value.fields) {
-      const search = value.fields.text;
+      const search = String(value.fields.text ?? "").trim();
       delete value.fields.text;
       (value.fields as FilterQuery<PublishedDataDocument>).$text = {
         $search: search,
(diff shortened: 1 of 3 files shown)


src/published-data/pipes/filter.pipe.ts
```diff
@@ -16,6 +16,15 @@ class AppendFieldsToFilterPipe implements PipeTransform {
   }) {
     if (isEmpty(value.fields)) return value;
     const filter = value.filter ?? {};
+
+    if ("text" in value.fields) {
```
minottic: since text could be "", shouldn't we check this:

```TS
    if (value.fields["text"]) {
```

abdimo101: changed in [d227be0](https://github.com/SciCatProject/backend/pull/2715/commits/d227be09f4a98cb07807dbca8820839fb9563335)

minottic (approved):
the second if seems redundant to me, but happy to keep if it's for some type guarding

PR #2724, opened 2026-05-08
fix(config): parse UPDATE_DATASET_LIFECYCLE_GROUPS as string array

alubbock:
## Description

The updateDatasetLifecycle access group was stored as a raw string rather than split into an array like every other group config entry. The CASL factory calls .includes() on this value to check membership; on a string, .includes() does substring matching, so a user in group "lifecycle" would pass the check when only "lifecycle-managers" is configured, granting unauthorised DatasetLifecycleUpdate access.

Adds a regression test that exercises the real configuration() function to catch any future reversion.

## Changes:

- Convert UPDATE_DATASET_LIFECYCLE_GROUPS from a comma-separated string of groups to an array
- Add regression test

## Tests included

- [X] Included for each change/fix?
- [X] Passing? 

## Summary by Sourcery

Ensure dataset lifecycle update access groups are parsed consistently with other group-based permissions to prevent unauthorized access.

Bug Fixes:
- Parse UPDATE_DATASET_LIFECYCLE_GROUPS as a trimmed string array instead of a raw string to avoid substring-based authorization matches.

Tests:
- Add a regression test around the CASL ability factory using the real configuration() to verify correct handling of dataset lifecycle update groups.


Junjiequan (approved):
lgtm

Issue #2250, opened 2025-10-03
Updating number of files/size of dataset when creating datablocks leads to double counting


omkar-ethz:
## Summary
When creating `origDatablocks`, the dataset's `size` and `numberOfFiles` are updated ([code](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/datasets.controller.ts#L2091)).
The dataset's size / number of files should be the sum of its origDataBlocks files - this part works as expected.

During archiving,`datablocks` are created. However, the dataset size is updated again ([code](https://github.com/SciCatProject/scicat-backend-next/blob/master/src/datasets/datasets.controller.ts#L2372)). This leads to double counting the number of files and false increase in size.

## Expected Behaviour
Only update packedSize and numberOfFiles archived while creating datablocks

bpedersen2:
Sounds reasonable

PR #2726, opened 2026-05-08
fix(origdatablocks): replace JS map/reduce with aggregation in updateDatasetSizeAndFiles

alubbock:
## Description

Replaces the `updateDatasetSizeAndFiles` helper in `OrigDatablocksController` with a single MongoDB `$group` aggregation.

### Correctness

The previous implementation called `findAll` with no `limits`, which caused `parseLimitFilters` to apply its default of `limit: 100`. For any dataset backed by more than 100 origdatablocks, only the first 100 were included in the sum, silently understating `size` and `numberOfFiles` on the dataset record. This bug is triggered on every origdatablock create, update, and delete.

The aggregation has no such cap -- it processes every matching document.

### Performance

Beyond correctness, the previous approach fetched full origdatablock documents into application memory, including their complete `dataFileList` arrays, and computed the totals in JavaScript. The replacement pushes both calculations to the database using `$sum` and returns only the two scalars needed.

Related upstream issues: #2250 (double-counting of size/files when creating datablocks) and #1688 (size reset to 0 after dataset PATCH) both involve the same size-tracking logic.

## Changes:

* `src/origdatablocks/origdatablocks.service.ts` -- new `aggregateSizeAndFileCount(datasetId)` method using `$match` + `$group`; `dataFileList` wrapped in `$ifNull` so documents with a missing or null field return 0 rather than erroring; returns `{ size, numberOfFiles }`, stripping the internal `_id: null` field from the `$group` result
* `src/origdatablocks/origdatablocks.controller.ts` -- `updateDatasetSizeAndFiles` reduced to a single call to the new service method; unused `PartialUpdateDatasetDto` import removed

## Tests included

- [x] Passing?

Three new service-level tests for `aggregateSizeAndFileCount`:
* returns summed totals when origdatablocks exist
* returns `{ size: 0, numberOfFiles: 0 }` when none exist for the dataset
* passes the correct `datasetId` to the `$match` stage

Two new controller-level tests for `updateDatasetSizeAndFiles`:
* verifies the aggregation result flows through to `datasetsService.findByIdAndUpdate`
* verifies zero totals are propagated correctly

## Documentation
- [x] swagger documentation updated (required for API changes) -- n/a, no API change
- [x] official documentation updated -- n/a, internal implementation change only

## Summary by Sourcery

Replace dataset size and file count recalculation for origdatablocks with a MongoDB aggregation-based implementation to ensure correct totals for all datasets.

Bug Fixes:
- Fix dataset size and numberOfFiles being understated when more than 100 origdatablocks exist by eliminating the implicit query limit during recalculation.

Enhancements:
- Introduce an aggregateSizeAndFileCount service method that computes dataset size and file count in MongoDB using a $group aggregation and handles missing dataFileList fields.
- Simplify OrigDatablocksController.updateDatasetSizeAndFiles to delegate to the new aggregation-based service method and update the dataset in a single step.

Tests:
- Add unit tests for aggregateSizeAndFileCount covering normal aggregation, empty results, and correct datasetId matching in the aggregation pipeline.
- Add controller tests to verify updateDatasetSizeAndFiles uses the aggregation results to update datasets, including propagation of zero totals.

### DIFF
--- a/src/origdatablocks/origdatablocks.controller.spec.ts
+++ b/src/origdatablocks/origdatablocks.controller.spec.ts
@@ -22,6 +22,7 @@ class CaslAbilityFactoryMock {}
 describe("OrigDatablocksController", () => {
   let controller: OrigDatablocksController;
   let origDatablocksService: OrigDatablocksServiceMock;
+  let datasetsService: DatasetsServiceMock;
 
   beforeEach(async () => {
     const module: TestingModule = await Test.createTestingModule({
@@ -37,10 +38,10 @@ describe("OrigDatablocksController", () => {
     controller = module.get<OrigDatablocksController>(OrigDatablocksController);
     origDatablocksService = module.get<OrigDatablocksService>(
       OrigDatablocksService,
-    );
-
-    // Mock internal methods to isolate the handler under test
-    controller["updateDatasetSizeAndFiles"] = jest.fn();
+    ) as unknown as OrigDatablocksServiceMock;
+    datasetsService = module.get<DatasetsService>(
+      DatasetsService,
+    ) as unknown as DatasetsServiceMock;
   });
 
   it("should be defined", () => {
@@ -57,6 +58,11 @@ describe("OrigDatablocksController", () => {
       updatedAt: new Date(Date.now() - 1000),
       datasetId: "ds1",
     };
+
+    beforeEach(() => {
+      // Isolate the update handler from the side-effect helper
+      controller["updateDatasetSizeAndFiles"] = jest.fn();
+    });
 
     it("should throw NotFoundException if datablock not found before update", async () => {
       origDatablocksService.findOne.mockResolvedValue(null);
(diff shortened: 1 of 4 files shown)


alubbock:
@sourcery-ai review

alubbock:
@sourcery-ai Thanks for the follow-up review. Two responses:

**`$ifNull` on `size`** — not needed here. MongoDB's `$sum` already treats missing or null numeric fields as 0, so the behaviour is identical to wrapping in `$ifNull`. This is different from the `dataFileList` case, where a missing *array* causes `$size` to error. `size` is also declared `required` in the schema, making the defensive case unlikely. No behaviour change would result from adding it.

**Explicit `$project` stage** — the field stripping is already handled in TypeScript before the result is returned:

```ts
return result
  ? { size: result.size, numberOfFiles: result.numberOfFiles }
  : { size: 0, numberOfFiles: 0 };
```

The `$group` stage only ever emits `_id`, `size`, and `numberOfFiles`, so there are no intermediary fields to leak regardless. Adding a `$project` would be a no-op pipeline stage.

Junjiequan (approved):

PR #2736, opened 2026-05-18
fix(Jobs): fix jobs fullfacet access with text

abdimo101:
## Description
This PR places a check if the pipeline contains a text and merges access to the text match so MongoDB accepts the pipeline and access is still enforced.

## Motivation
When searching in Jobs table, the fullfacet endpoint throws an error: 

"{status: 400, message: "$match with $text is only allowed as the first pipeline stage"}
message
: 
"$match with $text is only allowed as the first pipeline stage"
status
: 
400"

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Included for each change/fix?
- [ ] Passing? 

## Documentation
- [ ] swagger documentation updated (required for API changes)
- [ ] official documentation updated

## Summary by Sourcery

Bug Fixes:
- Fix jobs fullfacet aggregation error when combining access filtering with $text search by correctly positioning and merging the $match stage.

PR #2745, opened 2026-05-20
fix: sort on get all jobs

nitrosx:
## Description
This PR implement and fix the sorting of jobs when retrieving a number of jobs using `/api/v4/jobs`. It also improves the building process for the mongo pipeline used to retrieve the jobs in the jobs service.

## Motivation
Working on an ESS specific task, I realized that the jobs were returned in the wrong order when requesting them from the most recent to the oldest. The Filter passed to the GET `/api/v4/jobs` is:
```
{
   "where": {
      "type": "embargo_period",
      "statusCode": "jobSubmitted"
    },
    "fields":["createdAt"],
    "limits": {
      "limit": 10,
      "skip": 0,
      "sort": {
        "createdAt": "desc"
      }
    }
  }
```

## Fixes
* fixed job service to correctly use the sort in the mongo pipeline

## Changes:
* refactored casl ability function that check access to jobs, so it does not add unnecessary stages to the mongo pipeline 

## Tests included
- [ ] Passing? 

## Documentation

## Summary by Sourcery

Fix job listing sort behavior and refine MongoDB query construction for job access control and retrieval.

Bug Fixes:
- Ensure job retrieval applies the requested sort, limit, and skip correctly when building the MongoDB aggregation pipeline.

Enhancements:
- Simplify and optimize the jobs access MongoDB query by filtering out impossible conditions, handling unrestricted access, and avoiding unnecessary $or and $match stages.


src/casl/casl-ability.factory.ts
```diff
@@ -1926,12 +1926,37 @@ export class CaslAbilityFactory {
...
+    ];
+
+    // Remove impossible queries
+    const meaningfulQueries = queries.filter(
+      (q) =>
+        JSON.stringify(q) !==
+        JSON.stringify({
+          $expr: { $eq: [0, 1] },
+        }),
+    );
```
Junjiequan: maybe name the function like `queryValidation`,  `falsyQueryValidation` would be more clear?

src/jobs/jobs.service.ts
```diff
@@ -157,12 +157,15 @@ export class JobsService {
...
     }
 
-    const sort = filter.sort ?? parseLimitFilters(filter.order).sort;
-    if (sort && !isEmpty(sort)) pipeline.push({ $sort: sort });
-    if (filter.limits?.limit) pipeline.push({ $limit: filter.limits.limit });
-    if (filter.limits?.skip) pipeline.push({ $skip: filter.limits.skip });
+    const limits = parseLimitFilters(filter.limits);
+    if (!isEmpty(limits.sort)) {
+      const sort = parsePipelineSort(limits.sort);
+      pipeline.push({ $sort: sort });
```
Junjiequan, line 163: Maybe it's worth to have a default sort?

src/casl/casl-ability.factory.ts
```diff
@@ -1926,12 +1926,37 @@ export class CaslAbilityFactory {
...
+    if (meaningfulQueries.length === 0) {
+      return { $expr: { $eq: [0, 1] } };
+    }
+
+    // Single condition doesn't need $or
+    if (meaningfulQueries.length === 1) {
+      return meaningfulQueries[0];
+    }
+
+    return { $or: meaningfulQueries };
```
Junjiequan, line 1991: I'd avoid having this kind of query building logic in the casl-ability. If it is intended to be used by jobService only I'd include them in the service file with a queryBuilder helper function, such as `parseAccessFilters`, otherwise put them in the util file

src/casl/casl-ability.factory.ts
nitrosx: A dedicated function has been created

src/jobs/jobs.service.ts
nitrosx: I added a default sort

src/casl/casl-ability.factory.ts
nitrosx: I agree but I would address this in a future PR.
This functions were already in casl ability.
I think we would need to review the all file to be able properly address your concern.

Junjiequan (approved):
I'm approving it as it's a bug fix and fixes the root cause that PR #2736 did not address, please merge after the failign tests are resolved.


A few things maybe worth noting for future considerations:
- `meaningfulQueries` could be renamed to something more descriptive based on the context, such as `validQueries`
- The mongoQuery building logic `isImpossibleQuery`, `isEmptyObject`, `meaningfulQueries` etc. might be worth moving to utility layer, so we can scope down casl-ability.ts to rule definitions only.
- `{ $expr: { $eq: [0, 1] } }` could benefit from a short comment. e.g, `// it's generated by accessibleBy() when no rules match, forcing the query to return no results.`

nitrosx:
@Junjiequan I addressed your concerns, except moving the functions to `utils`.
I will do that in a future PR

PR #2761, opened 2026-05-27
feat: make user's accessToken available for jobv4

belfhi:
Add accessToken to jobClass so it can be reused in actions via job.accessToken in handlebars

## Description

the `accessToken` with which the Jobv4 was submitted is added to the jobClass so that in a jobConfig 
you can use `authorization: "Bearer {{job.accessToken}}"` for authorization.
This is especially useful if the job action performs a Scicat API call as the user which is not possible otherwise.

## Motivation
configurable actions and jobs using urlactoin make it possible to perform arbitrary API calls, and with the proposed changes these calls can be done as the user that is currently logged in a submits a job to scicat.

## Summary by Sourcery

Generate short-lived user JWTs for job executions and expose them to job templates to enable authenticated user-scoped actions.

New Features:
- Provide a derived user JWT (userToken) in the job execution context so jobv4 templates can perform authenticated actions on behalf of the job owner.
- Allow job configurations to use a dedicated JOB_TOKEN_EXPIRES_IN setting for controlling job-specific JWT lifetimes.

Bug Fixes:
- Fix UsersService to store the correct user identifier from user identities and align JWT payload structure with JWTUser.

Enhancements:
- Relax the JobClass results field type to SchemaTypes.Mixed for greater flexibility in stored results.
- Add utility and tests for generating job user tokens and for using userToken within Handlebars-based job templates.

Documentation:
- Document the JOB_TOKEN_EXPIRES_IN environment variable and its fallback behavior in the README.

Tests:
- Add unit tests for job Handlebars utilities to ensure userToken is available in templates and behaves as expected.
- Add unit tests for job token utility functions to cover success and failure scenarios when generating user JWTs.


bpedersen2:
Hmm, while I see the idea, maybe a different implementation would be better:
 -  just store the user id
 -  generate a shortlived scicat JWT token for this user on job execution and inject that instead
 
If  we need the token to access non-scicat resources, it will get a bit more tricky as either a token exchange is needed anyway (the token audience field would not match) or maybe the new keycloak  Identity Assertion JWT Grant ( see https://www.keycloak.org/2026/07/keycloak-2670-released) may later be used.

belfhi:
I like the idea of a short-lived token that is generated, that could also work. I'll investigate.
Regarding non-scicat resources, that's a very different field but also very intriguing, I hadn't even considered that yet.

belfhi:
@bpedersen2 I changed the approach and at least in a first test it seems to work as expected. Does that look better from a safety perspective?

bpedersen2 (commented):
Looks better to me

bpedersen2:
@sourcery-ai review

belfhi:
concering the change of the payload from groups to currentGroups:

>The only callers of createUserJWT are:
>- token.utils.ts (our job token generation)
>- users.controller.ts (admin endpoint /users/createUserJWT)
>
>The users.controller.ts caller passes the request.user which is already a JWTUser from JwtStrategy.validate() — so signing the full JWTUser is fine there too.
>
>For the payload change: the old format { username, groups } was unique to createUserJWT (the login JWT was always the full User object via auth.service.ts). Since both callers are internal and the payload is consumed by JwtStrategy.validate() which expects the full User-like shape, the new format is correct.

belfhi:
Any more comments @bpedersen2 ? Maybe you also want to take a look @sbliven ?

PR #2764, opened 2026-05-28
fix(Dataset): allow public dataset access in v4 findById endpoint for unauthenticated users

abdimo101:
## Description
This PR allows unauthenticated users to access published datasets through the V4 `findById` endpoint. The `@CheckPolicies `decorator now checks for both DatasetRead and DatasetReadOnePublic permissions just like in the V3 `findById` .

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Bug Fixes:
- Fix v4 dataset findById authorization to permit access to publicly published datasets without authentication, consistent with v3 behavior.

### DIFF
--- a/src/datasets/datasets.v4.controller.ts
+++ b/src/datasets/datasets.v4.controller.ts
@@ -675,8 +675,11 @@ export class DatasetsV4Controller {
   // GET /datasets/:id
   //@UseGuards(PoliciesGuard)
   @UseGuards(PoliciesGuard)
-  @CheckPolicies("datasets", (ability: AppAbility) =>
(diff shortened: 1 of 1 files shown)


HayenNico (commented):
Hi @abdimo101,
the DatasetReadOnePublic action/permission is only granted to unauthenticated users, who are forbidden from using any endpoints in `datasets.v4.controller.ts` by design. An unauthenticated user should use the /api/v4/datasets/public endpoints defined in `datasets-public.v4.controller.ts` instead

PR #2782, opened 2026-06-15
feat(publish): run validation on create

fpotier:
## Description

Runs ajv validation during creation and update of PublishedData records, ignoring any validation error that may occur.
## Motivation

We use ajv validation process to auto-fill some properties, for instance the full name of a creator is derived from the given and the family names.

This caused some confusion to our users. The frontend uses `metadata.creators[].name` to display the list of creators but the full names are not computed on creation/updates.

In combination with #2749, it would also allow us to keep metadata accurate (e.g. computing the full size of the published data if the dataset list changes)

## Tests included

- [x] Passing? 

## Documentation

## Summary by Sourcery

Run metadata validation across published-data creation, updates, and resynchronization to keep derived fields accurate.

New Features:
- Run AJV validation during published-data creation and updates so derived metadata fields are populated consistently.
- Apply validation to full and partial resync operations as well as standard updates.

Enhancements:
- Merge existing records with partial updates before validation to preserve complete published-data metadata.

Tests:
- Add endpoint coverage verifying derived publication-year metadata during create, update, partial update, and resync flows.

### DIFF
--- a/src/published-data/dto/update-published-data.v4.dto.ts
+++ b/src/published-data/dto/update-published-data.v4.dto.ts
@@ -52,7 +52,7 @@ export class UpdatePublishedDataV4Dto {
    */
   @IsObject()
   @IsOptional()
-  readonly metadata?: Record<string, unknown> = {};
+  readonly metadata?: Record<string, unknown>;
 }
 
 export class PartialUpdatePublishedDataV4Dto extends PartialType(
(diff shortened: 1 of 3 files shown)


minottic (approved):
thanks! I wonder if this might be a problem in some edge cases, I will add an issue to raise that and suggest a possible later solution

fpotier:
> thanks! I wonder if this might be a problem in some edge cases, I will add an issue to raise that and suggest a possible later solution

what edge case do you have in mind?
If the dto doesn't have any `metadata` property, then it will not be initialized (see the 1st sourcery comment)

PR #2789, opened 2026-06-15
fix(Jobs): correct aggregation pipeline stage order for text search with access filters

abdimo101:
## Description
This PR fixes an issue in the jobs fullfacet endpoint where text search queries combined with access control filters were failing. 

Before this fix, when searching with text queries, the pipeline would place the access filter at the first stage and push the text search to the second stage. So the pipeline structure would like this:
``` 
[
  {
    "$match": {
      "$or": [
        {
          "ownerUser": "user5.2",
          "type": "acquisition_script"
        },
        {
          "ownerGroup": {
            "$in": [
              "group5"
            ]
          },
          "type": "acquisition_script"
        },
        {
          "ownerUser": "user5.2",
          "type": "visualization"
        },
        {
          "ownerGroup": {
            "$in": [
              "group5"
            ]
          },
          "type": "visualization"
        },
        {
          "ownerUser": "user5.2",
          "type": "embargo_period"
        },
        {
          "ownerGroup": {
            "$in": [
              "group5"
            ]
          },
          "type": "embargo_period"
        }
      ]
    }
  },
  {
    "$match": {
      "$or": [
        {
          "$text": {
            "$search": "embargo_period"
          }
        }
      ]
    }
  },
  {
    "$facet": {
      "all": [
        {
          "$match": {}
        },
        {
          "$count": "totalSets"
        }
      ]
    }
  }
]
```

Which would throw this error: 

``` 
response: {
    status: 400,
    message: '$match with $text is only allowed as the first pipeline stage'
  }
```
According to MongoDB documentation on aggregation pipeline constraints: https://www.mongodb.com/docs/manual/tutorial/text-search-in-aggregation/, the $text operator must always be the first stage in an aggregation pipeline. No $match conditions can come before it

The fix introduces a utility function in `utils.ts` called `addAccessMatchToPipeline` that checks if a text search field exists in the request. If text search is present, the access filter is inserted at index 1 instead of being unshifted to index 0, keeping the text search as the first stage. If no text search exists, the access filter is unshifted normally to the beginning.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Fix the jobs aggregation pipeline so text search stages remain first while still applying access control filters.

Bug Fixes:
- Ensure jobs fullfacet aggregation pipelines with text search comply with MongoDB’s requirement that $text be the first $match stage to prevent query failures.

Enhancements:
- Introduce a shared utility to insert access control $match stages into aggregation pipelines while preserving correct stage ordering.

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -13,7 +13,7 @@ import { ScientificRelation } from "./scientific-relation.enum";
 import { DatasetType } from "src/datasets/types/dataset-type.enum";
 import { isPlainObject, mapValues, omit, pickBy, some } from "lodash";
 import { MetadataSourceDoc } from "src/metadata-keys/metadatakeys.service";
-import { IJobFields } from "src/jobs/interfaces/job-filters.interface";
+import type { IJobFields } from "src/jobs/interfaces/job-filters.interface";
 
 // add Å to mathjs accepted units as equivalent to angstrom
 const isAlphaOriginal = Unit.isValidAlpha;
@@ -1382,7 +1382,7 @@ export function addAccessMatchToPipeline<T>(
   access: FilterQuery<T>,
   fields: IJobFields,
 ) {
-  if (fields["text"]) {
+  if ("text" in fields) {
     pipeline.splice(1, 0, {
       $match: access,
     });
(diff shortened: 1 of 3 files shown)


Junjiequan:
@abdimo101 could you take a look for the copliot review and if the suggestions does not make sense just archive them

Junjiequan (approved):

PR #2795, opened 2026-06-22
feat(OrigDatablock): file count endpoints with isPublished backfill

abdimo101:
## Description
This PR adds new endpoints to count files across origdatablocks linked to a specific dataset. It also includes API tests for the public and private file count endpoints.

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Add API support to count files across origdatablocks, including public and authenticated variants, and cover them with tests.

New Features:
- Expose a secured /api/v4/origdatablocks/files/count endpoint to return the number of files matching optional filters.
- Expose a public /api/v4/origdatablocks/public/files/count endpoint to return the number of public files matching optional filters.
- Add a service-layer aggregation to count files across origdatablocks with support for filters and projections.

Tests:
- Add authenticated and unauthenticated tests for the private origdatablocks files count endpoint.
- Add tests for the public origdatablocks files count endpoint and adjust existing test case numbering.

### DIFF
--- a/src/origdatablocks/origdatablocks-public.v4.controller.ts
+++ b/src/origdatablocks/origdatablocks-public.v4.controller.ts
@@ -14,7 +14,11 @@ import {
   IOrigDatablockFields,
   IOrigDatablockFiltersV4,
 } from "./interfaces/origdatablocks.interface";
-import { FullFacetFilters, FullFacetResponse } from "src/common/types";
+import {
+  FullFacetFilters,
+  FullFacetResponse,
+  CountApiResponse,
+} from "src/common/types";
 import { getSwaggerOrigDatablockFilterContent } from "./types/origdatablock-filter-content";
 import {
   OrigDatablockLookupKeysEnum,
@@ -236,5 +240,53 @@ export class OrigDatablocksPublicV4Controller {
     });
 
     return origdatablock;
+  }
+
+  // GET /origdatablocks/public/files/count
+  @AllowAny()
+  @Get("/files/count")
+  @ApiOperation({
+    summary: "It returns the count of public files",
+    description:
+      "It returns the total number of public files across all origdatablocks matching the provided filter.",
+  })
+  @ApiQuery({
+    name: "filter",
+    description:
+      "Database filters to apply when retrieving count of public files",
+    required: false,
+    type: String,
+    content: getSwaggerOrigDatablockFilterContent({
+      where: true,
+      include: false,
+      fields: false,
+      limits: false,
+    }),
+  })
+  @ApiResponse({
+    status: HttpStatus.OK,
+    type: CountApiResponse,
+    description:
+      "Return the number of public files in the following format: { count: integer }",
+  })
+  async countFilesPublic(
+    @Query(
+      "filter",
+      new FilterValidationPipe(
+        ALLOWED_ORIGDATABLOCK_KEYS,
+        ALLOWED_ORIGDATABLOCK_FILTER_KEYS,
+        {
+          where: true,
+          include: false,
+          fields: false,
+          limits: false,
+        },
+      ),
+    )
+    queryFilter: string,
+  ) {
+    const parsedFilter = JSON.parse(queryFilter ?? "{}");
+
+    return this.origDatablocksService.countFiles(parsedFilter);
   }
 }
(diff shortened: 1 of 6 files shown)


src/origdatablocks/origdatablocks-public.v4.controller.ts
```diff
@@ -237,4 +241,52 @@ export class OrigDatablocksPublicV4Controller {
...
+          include: false,
+          fields: false,
+          limits: false,
+        },
+      ),
+    )
+    queryFilter?: string,
+  ) {
+    const parsedFilter = JSON.parse(queryFilter ?? "{}");
+
```
HayenNico, line 289: For the /public endpoint, I think the controller should still inject an `{ "isPublished": true }` filter into parsedFilter before passing it to the service layer (the same stage addAccessBasedFilters is applied in the nonpublic endpoint). That way, the OrigDatablocksService does not need to distinguish which endpoint the request came from.
The current  implementation sends a filter with no access restrictions to the service. I believe there's still some auth logic on the service layer when calculating lookups, but this likely produces an overcount since it may pick up unpublished documents

test/OrigDatablockV4.js
```diff
@@ -909,31 +909,73 @@ describe("2800: OrigDatablock v4 endpoint tests", () => {
...
+      };
+
+      return request(appUrl)
+        .get("/api/v4/origdatablocks/files/count")
+        .query({ filter: JSON.stringify(filter) })
+        .expect(TestData.AccessForbiddenStatusCode)
+        .expect("Content-Type", /json/);
+    });
+
+    it("0601: should be able to fetch the count of origdatablocks files", async () => {
```
HayenNico: Would be nice to have two tests - one for adminIngestor and one for user1 (or one of the others). The last expectation count > 0 should be refined to match the exact number of expected documents, and that number should then differ for adminIngestor with elevated access and user1.

test/OrigDatablockV4Public.js
```diff
@@ -373,8 +373,33 @@ describe("2900: OrigDatablock v4 public endpoint tests", () => {
...
+      return request(appUrl)
+        .get("/api/v4/origdatablocks/public/files/count")
+        .query({ filter: JSON.stringify(filter) })
+        .expect(TestData.SuccessfulGetStatusCode)
+        .expect("Content-Type", /json/)
+        .then((res) => {
+          res.body.should.be.a("object");
+          res.body.should.have.property("count");
+          res.body.count.should.be.a("number");
+          res.body.count.should.be.greaterThan(0);
```
HayenNico: Related to the first two comments: Just checking count > 0 is imprecise and will not reveal the potential overcount.

HayenNico (changes requested):
Generally solid and a good addition to the odb controllers - needs a (potential) fix for access restrictions on the public v4 endpoint and a bit more test coverage, approved outside those two aspects :+1:

src/origdatablocks/origdatablocks-public.v4.controller.ts
Junjiequan: addressed in [cf4c680](https://github.com/SciCatProject/backend/pull/2795/commits/cf4c6800a20231a7add10b9d2b9a30002388b42c)

test/OrigDatablockV4.js
Junjiequan: addresed in [6432d84](https://github.com/SciCatProject/backend/pull/2795/commits/6432d848725b28651f405e0f736a00cca1b72159)

test/OrigDatablockV4Public.js
Junjiequan: addressed in [6432d84](https://github.com/SciCatProject/backend/pull/2795/commits/6432d848725b28651f405e0f736a00cca1b72159)

Junjiequan (commented):
I took over the PR and made some changes as abdi has left.

- Improved OrigDatablockV4 and OrigDatablockV4Public api tests 
- Added bakfill migration script for legacy origDatablocks to inherit isPublished from parent dataset ( included the script in this PR, as many old origDatablocks don't have isPublished field)
- simplified `countFiles` service to keep only the unwind pipe for the combined total file count and removed other unnecessary pipes

HayenNico (approved):

PR #2825, opened 2026-07-09
fix: allow multiple sorts

Junjiequan:
## Description
parseOrderLimits only parsed one field:direction and dropped the rest. It now splits on commas, so creationTime:desc,pid:asc works.
when rows share the same creationTime (common with batch ingestion), sorting by creationTime alone leaves their order undefined. MongoDB can arrange tied rows differently per request, so the same dataset appears on two pages and others get skipped. Adding a unique field (e.g, pid) as the last sort key fixes their position.

Backward compatible: a single key with no comma parses as before.
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Support comma-separated sort fields in order parsing to ensure stable pagination and avoid duplicate/omitted rows across pages.

Bug Fixes:
- Fix parseOrderLimits dropping additional sort fields by correctly handling comma-separated field:direction pairs.

Tests:
- Add unit tests covering single and multiple sort fields in parseOrderLimits for stable pagination behavior.

### DIFF
--- a/src/common/utils.spec.ts
+++ b/src/common/utils.spec.ts
@@ -1,5 +1,5 @@
 // Currently only covers converToSI
-import { convertToSI, parseDate } from "./utils";
+import { convertToSI, parseDate, parseOrderLimits } from "./utils";
 
 describe("convertToSI", () => {
   it("should convert a known unit to SI successfully", () => {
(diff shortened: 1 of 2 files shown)


sbliven (changes requested):
I wasn't aware that the 'order' term could take a field. This is not documented in the endpoint descriptions. Could you update the documentation strings (eg `fullQueryDescriptionLimits`, `filterDescriptionSimplified`, etc).

This is only used for v3, since v4 uses the structured `sort` instead of the string `order`. This seems like a reasonable way to add multi-sort in a backwards-compatible way. However, this is a good indication that we need to stop calling /v3 endpoints from the frontend and start using /v4 features.

The sorcery suggestion to strip whitespace seems reasonable.

PR #2827, opened 2026-07-13
feat(auth): implement OIDC RP-initiated logout and token refresh

### LINKED ISSUE #1924: OIDC RP Initiated Logout
## Summary
OIDC Logout should include either `id_token_hint` with the id_token or `client_id` and `post_logout_redirect_uri` as specified in the specs 

https://openid.net/specs/openid-connect-rpinitiated-1_0.html

## Steps to Reproduce
use oidc_auto_logout

## Current Behaviour
the logout endpoint is called without any parameters

## Expected Behaviour
url parameters are sent with the logout endpoint

belfhi:
## Description
In the Scicat backend, there is some logic to call the oidc `end_session_endpoint` to log the user out of the IdP as well as out of scicat. 
This is currently broken, since the oidc spec https://openid.net/specs/openid-connect-rpinitiated-1_0.html
requires additional URI parameters to be added in the request.

Also, since the `id_token` needs to be a valid token for the logout, token refresh was implemented
because for most setups the token lifetime is shorter than the typical Scicat JWT duration.

The accessGroups are also fetched from the userinfo endpoint from the idp, which makes it much easier to update those, since the information is not only passed to scicat at login time. 
It also allows the oidc auth endpoint to accept access_tokens, which would be closer to the oidc spec.

## Motivation
OIDC Logout should work, Tokens should be small and always up to date.

## Fixes
#1924 

## Changes:
* fix oidc logout
* add token refresh
* get accessGroups also from userinfo endpoint

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Implement OIDC RP-initiated logout, session-based token refresh, and dynamic access group updates using the userinfo endpoint.

New Features:
- Introduce a session-based OIDC token refresh mechanism that periodically refreshes ID and access tokens.
- Fetch and merge access group-related claims from the OIDC userinfo endpoint to keep user permissions up to date.

Bug Fixes:
- Fix OIDC logout by generating a compliant RP-initiated end session URL based on the configured client and callbacks.

Enhancements:
- Persist OIDC tokens in the session and clear them on logout to better align with OIDC flows and improve session management.
- Update OIDC validation to accept access tokens and prefer userinfo responses while gracefully falling back to ID token claims.
- Refresh stored user identities and access groups when tokens are refreshed to keep user profiles synchronized.

Tests:
- Extend auth-related unit tests and mocks to cover the new token refresh service integration and session refresh handling.


bpedersen2:
Test failures are due to: 
``` OIDC issuer discovery failed: only requests to HTTPS are allowed ```

which is legit as OIDC requires https

**Comment by bpedersen2** [comment]
In the test you probably need to use allowInsecureRequests at some place:
https://github.com/panva/openid-client/blob/main/docs/functions/allowInsecureRequests.md#function-allowinsecurerequests

File: src/config/configuration.ts [hunk 1]
```diff
@@ -358,6 +358,12 @@ const configuration = () => {
       scope: process.env.OIDC_SCOPE, // Example: "openid profile email"
```
bpedersen2, line 358: maybe mention (either here or in the config doc) that `offline_access` is required as well for getting a  refresh token.

belfhi: Well that's not true, without the`offline_access` you get a refresh_token, too. It's shorter lived (same as the access_token) and bound to the SSO session but it _does_ exist. An offline session should not be required, even be discouraged, since we only want to do refresh while the JWT / scicat session is valid (typically ~12h).

belfhi:
> Test failures are due to: `OIDC issuer discovery failed: only requests to HTTPS are allowed`
> 
> which is legit as OIDC requires https

Ah, yes I changed the oauh2-mock-server config to run with (self-signed) https. 

The e2e tests revealed some problems in the azp validation but that should be fixed, too.

sbliven:
Thanks for this. The logout fix will be nice to have.

Is there anything that should be configured on the OIDC Provider (keycloak) side to match this? Eg if `OIDC_USERINFO_ENABLED=yes` should we remove access group claims from the access token? Are there additional settings?

I didn't follow how the refresh_token works. Does scicat store this for each user, just so logout works? What happens if they log in from multiple sessions simultaneously?

I'm not sure if this is related to this PR, but it's a bit awkward that scicat issues its own tokens at all. Would any of these issues be easier to solve if a OIDC token were already valid for use in `Authorization: Bearer` headers, rather than needing to be exchanged for a scicat token first?

PR #2860, opened 2026-07-29
perf: avoid recompute to optimize size updates

minottic:
## Description
The size and number of files was computed in a way that required the full datablocks scan by datasetId. The new implementation increments the sizes by simply adding the new entries. This is much more efficient

## Tests included

- [x] Included for each change/fix?
- [x] Passing? 

## Documentation
- [ ] swagger documentation updated (required for API changes)
- [ ] official documentation updated

## Summary by Sourcery

Optimize dataset size and file count maintenance by incrementally applying changes from datablocks and origdatablocks instead of recomputing aggregates.

New Features:
- Support incremental dataset size and file count updates when creating, updating, or deleting datablocks and origdatablocks.
- Expose new service and controller methods that operate on filters and pass old/new datablock documents for size update calculations.

Bug Fixes:
- Prevent dataset size/file updates when the targeted datablock or origdatablock is missing or disappears during update, returning appropriate HTTP exceptions instead.
- Ensure origdatablock and datablock update endpoints respect optimistic concurrency and permission checks while avoiding unintended dataset mutations.

Enhancements:
- Refine DatasetsService to compute and apply size and file count deltas directly on the dataset model using $inc operations.
- Improve service APIs for datablocks and origdatablocks to more clearly express combined update-plus-dataset-size maintenance flows.
- Strengthen mocking in origdatablocks unit tests to better cover create, update, and delete behaviors tied to dataset size updates.

Tests:
- Add unit tests for DatasetsService.updateDatasetSizeAndFiles to validate delta calculations for create, update, and delete scenarios.
- Extend datablocks and origdatablocks service and controller tests to cover incremental size updates and error handling when documents are missing.
- Introduce an integration test verifying dataset packedSize and numberOfFilesArchived stay consistent when a datablock is modified.

PR #2862, opened 2026-07-30
fix: stop patches from resetting dataset size fields to 0

minottic:
## Description
Having explicit defaults makes plainToInstance attach fields with 0s even when missing in the patch payload. This then updates the document's values in mongo to 0 even when missing in the payload

Raw and derived datasets have dedicated DTOs so this fixes the bug for other types

## Tests included

- [x] Passing? 

## Documentation

## Summary by Sourcery

Prevent dataset size-related fields from being unintentionally reset when applying partial updates to v3 custom datasets.

Bug Fixes:
- Ensure size, packedSize, numberOfFiles, and numberOfFilesArchived are not defaulted to 0 when omitted from update payloads.

Tests:
- Add regression tests verifying dataset size-related fields are preserved across unrelated PATCH operations on v3 custom datasets.

### DIFF
--- a/src/datasets/dto/update-dataset.dto.ts
+++ b/src/datasets/dto/update-dataset.dto.ts
@@ -105,7 +105,7 @@ export class UpdateDatasetDto extends OwnableDto {
   })
   @IsOptional()
   @IsInt()
-  readonly size?: number = 0;
+  readonly size?: number;
 
   @ApiProperty({
     type: Number,
@@ -116,7 +116,7 @@ export class UpdateDatasetDto extends OwnableDto {
   })
   @IsOptional()
   @IsInt()
-  readonly packedSize?: number = 0;
+  readonly packedSize?: number;
 
   @ApiProperty({
     type: Number,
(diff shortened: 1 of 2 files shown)


fpotier (commented):
Does the size update mechanism handle null?
For instance if I create a dataset without specifying a size and then add an origdatablock (I believe this triggers a recompute of the size)
Would be a breaking change but since the backend does the book keeping for size, packedSize etc. would it make sense to remove/ignore them from create/update dtos?

minottic:
good question, let me add a test for that and we see that

minottic:
test added, codewise they are defaulted in mongo [here](https://github.com/SciCatProject/backend/blob/master/src/datasets/schemas/dataset.schema.ts#L137-L179) before reaching the DB and in size improved in this pr #2860 and addressed with the current behaviour [here](https://github.com/SciCatProject/backend/blob/master/src/datablocks/datablocks.controller.ts#L108-L109) and [here](https://github.com/SciCatProject/backend/blob/master/src/datasets/datasets.service.ts#L755-L777) and datablock tests covered [here](https://github.com/SciCatProject/backend/blob/master/test/Datablock.js#L29-L65)

fpotier (approved):

PR #2877, opened 2026-08-06
fix: apply dataset types check

minottic:
## Description
The datasetTypes check was skipped due to a wrong type comparison. This PR should fix it together with simplifying the variables creation.

Since the check was never enforced, it's a bit hard to know what the expected behaviour should be when the datasetTypes.json file is missing. 

1. should all custom types be allowed (as assumed by the tests)?
2. or should no custom types be allowed (as assumed by the controller, even if buggy)?

This version implements the 1 , but I am happy to change it.

## Fixes

* dataset type comparison wrong check

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Validate dataset creation against configured custom dataset types and adjust configuration accordingly.

Bug Fixes:
- Correct dataset type validation to ensure unsupported custom types are rejected during dataset creation.

Enhancements:
- Rename and streamline configuration and controller handling from generic datasetTypes to customDatasetTypes for clearer semantics.

Tests:
- Add an integration test verifying that datasets with unsupported types are rejected with a bad request error.

### DIFF
--- a/src/datasets/datasets.controller.ts
+++ b/src/datasets/datasets.controller.ts
@@ -152,12 +152,16 @@ export class DatasetsController {
     this.datasetCreationValidationRegex = this.configService.get<string>(
       "datasetCreationValidationRegex",
     );
-    this.datasetTypes = this.configService.get<string>("datasetTypes");
+    const customDatasetTypes =
+      this.configService.get<Record<string, string>>("customDatasetTypes");
+    this.customDatasetTypes = customDatasetTypes
+      ? Object.values(customDatasetTypes)
+      : [];
   }
   private accessGroups;
   private datasetCreationValidationEnabled;
   private datasetCreationValidationRegex;
-  private datasetTypes;
+  private customDatasetTypes;
 
   updateMergedFiltersForList(
     request: Request,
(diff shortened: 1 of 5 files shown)


src/datasets/datasets.controller.ts
```diff
@@ -745,23 +749,17 @@ export class DatasetsController {
     const outputDatasetDto = plainToInstance(dto, inputDatasetDto);
 
     if (
-      outputDatasetDto instanceof
-      (CreateRawDatasetObsoleteDto ||
-        CreateDerivedDatasetObsoleteDto ||
-        CreateDatasetDto)
+      this.customDatasetTypes.length > 0 &&
+      outputDatasetDto instanceof CreateDatasetDto &&
+      !this.customDatasetTypes.includes(outputDatasetDto.type)
```
HayenNico, line 676: I believe this condition will throw incorrectly if a `CreateDatasetDto` (API v4 style) is sent to the create/patch/put v3 endpoints with type `raw` or `derived`, while at least one custom dataset type is defined.

Could you add a test for `POST /api/v3/datasets` where a `CreateDatasetDto` with the same information content as a `CreateDatasetObsoleteDto` is sent (replacing `proposalId` -> `proposalIds` or adding `sampleIds`, `instrumentIds`)? That should confirm whether both entities are handled the same

HayenNico (changes requested):

src/datasets/datasets.controller.ts
minottic: ah true, this will indeed throw I think, thanks! Applying the change

HayenNico (approved):

src/datasets/datasets.controller.ts
minottic: I have added some more tests, but I checked this case and it can actually not happen because of [this](https://github.com/SciCatProject/backend/blob/master/src/datasets/datasets.controller.ts#L583-L602)

HayenNico: You're right - but I think that section is not 100% working as intended. Should be separate from this PR though in any case; I'll do some testing on my end what works and what doesn't when mixing API v3 endpoints and v4 DTOs

PR #2884, opened 2026-08-12
fix: enforce dataset permissions on v3 publisheddata register

knQzx:
The v3 publisheddata register endpoint updated datasets through the service without checking dataset level permissions, so any authenticated user could flip isPublished and lifecycle fields on arbitrary datasets by supplying their pids. This routes the update through the dataset controller so the dataset update permission is enforced.

## Summary by Sourcery

Bug Fixes:
- Enforce dataset update permissions when the v3 published data register endpoint updates datasets.


src/published-data/published-data.controller.spec.ts
```diff
@@ -7,7 +7,7 @@ import { DatasetsService } from "src/datasets/datasets.service";
 import { ProposalsService } from "src/proposals/proposals.service";
 import { PublishedDataController } from "./published-data.controller";
 import { PublishedDataService } from "./published-data.service";
-import { DatasetsController } from "src/datasets/datasets.controller";
+import { DatasetsV4Controller } from "src/datasets/datasets.v4.controller";
```
omkar-ethz, line 10: I had added the v3 controller in https://github.com/SciCatProject/backend/pull/2887/changes#diff-91782c7cfd65c8dcac1eea9e838d9943ac240fdc4ddf655a572ebafe393c0243, but seems it's not needed.
Should we remove the export (and provider) from datasets.module?

omkar-ethz (approved):

src/published-data/published-data.controller.spec.ts
knQzx: yes, it can go. after the switch to the v4 controller nothing injects `DatasetsController` anymore - both `formpopulate` and `register` go through `DatasetsV4Controller` now, and the only remaining mentions are in `datasets.module.ts` itself. keeping it in `providers` also makes nest instantiate the controller a second time next to the one it already builds from `controllers`.

pushed a commit that drops it from `exports` and `providers` and leaves it in `controllers`, so the v3 routes are untouched - it is exactly the revert of what #2887 added there.

checked locally: build and lint clean, 366 unit tests, the 34 jest e2e tests that boot the full `AppModule`, and the whole mocha api suite (1797 passing) all green. happy to split it into its own PR if you'd rather not carry it here.

PR #2885, opened 2026-08-12
fix: apply access filter to samples findOne

knQzx:
The samples findOne endpoint ran the caller's filter with only a class level read guard, so any authenticated user could read any sample by crafting the filter. This applies the same access filter the list endpoint uses.

## Summary by Sourcery

Enforce sample access controls when retrieving individual samples.

Bug Fixes:
- Apply the same access filtering to the samples findOne endpoint as to sample listing, preventing users from retrieving samples they are not authorized to access.

Tests:
- Add authorization coverage for findOne access by sample owner, unauthorized users, and access-group members.


omkar-ethz (approved):
LGTM! added missing e2e tests for findOne (similar to the list endpoint)

Junjiequan (approved):

PR #2886, opened 2026-08-12
fix: add access filter and status guard to v3 publisheddata operations

knQzx:
The v3 publisheddata update had no policy check beyond authentication, so any authenticated user could patch any published data record. This adds an access scoped lookup and a status guard so users can only change their own private records.

## Summary by Sourcery

Enforce ownership and lifecycle-state protections across v3 PublishedData operations.

Bug Fixes:
- Restrict v3 PublishedData reads and mutations to records users are authorized to access, preventing cross-user access to private records.
- Prevent updates to registered or amended records while allowing authorized updates to private records.

Enhancements:
- Apply consistent access scoping to PublishedData listing, counting, retrieval, registration, and resynchronization operations, with administrators retaining broader access.

Tests:
- Add end-to-end coverage for ownership-based access, administrator privileges, private-record visibility, mutation restrictions, and status-based update guards.


omkar-ethz:
The original PR was only to fix the issue with PATCH endpoint.
This also fixes the issue for all other endpoints.

omkar-ethz:
Adding e2e tests written using Claude for the new behavior

minottic (approved):
thanks!

Junjiequan (dismissed):
This needs to be reviewed more

Junjiequan:
Sorry for the sudden request changes, but I have couple of questions before we merge this PR

omkar-ethz:
> Sorry for the sudden request changes, but I have couple of questions before we merge this PR

Of course! Thank you for reviewing

Junjiequan:
My concern is about the missing authorization(`ownableClass`) on the publishedData schema.

`getMutationAccessFilters` allows only the createdBy user to update, register, or resync published data. But the pattern used across other SciCat modules is that ownerGroup users can update the entity. Filtering by createdBy means only that specific user can update it, which seems inconsistent to me.

IMO (after a short discussion with Max) a more consistent solution is to extend OwnableClass in the PublishedData schema, then add a migration script to set ownerGroup and accessGroups on existing published data. That needs a bit more discussion though, `PublishedData` didn't have those 2 fields before, so we'd need to agree on what to assign

Current approach fixes missing authorization issues temporarily, but I don't think this is fix for everyone, especially considering scicat auth pattern is group based onwership

omkar-ethz:
Good points! Carlo and I were also wondering why ownerGroup is absent from the published data schema.

>getMutationAccessFilters allows only the createdBy user to update, register, or resync published data. But the pattern used across other SciCat modules is that ownerGroup users can update the entity. Filtering by createdBy means only that specific user can update it, which seems inconsistent to me.

Agree, i just borrowed this logic from the v4 endpoint. Carlo also mentioned that this breaks the "four eyes" review of a private published data record that used to be possible before - but then, the frontend already uses v4 publisheddata endpoints, so this is already not supported currently. 

IMO, as the v3 endpoints are deprecated, and not used in frontend, but only used in landing-page (and https://github.com/SciCatProject/oai-provider-service ?), I think it's better to have this more restrictive behavior as a temporary fix. As currently, ~~even unauthenticated user~~ any authenticated user can read READ private PublishedData, and any authenticated user can PATCH someone else's private PublishedData via the v3 endpoints

But also happy to hold off the PR until the long term solution is discussed. We might just deploy this as a hotfix at PSI in the meantime

Junjiequan:
Since this change only affects v3 endpoints and addresses a serious security issue, I'd suggest we proceed @nitrosx ?

nitrosx:
If it is a security issue, we should merge. 
I would like you to add somewhere in documentation a reference to this hot fix and mention that it is a temporary fix, it deviates from the supported pattern, it will be supported until we redesign the new ownership of published data.

omkar-ethz:
@nitrosx i added the comment to a (newly created) authorization section for published data. It is with respect to CASL _before_ the refactor, so if we merge this PR first, we should update the docs in https://github.com/SciCatProject/backend/pull/2812

Junjiequan (approved):

PR #2888, opened 2026-08-12
fix: authorize target dataset when creating/updating datablocks

knQzx:
Creating a datablock only checked the caller supplied ownerGroup, so a user could attach a datablock to another owner's dataset and inflate its size and file count. This looks up the real dataset, enforces the datablock create permission on it, and takes ownerGroup and access groups from the dataset.

## Summary by Sourcery

Secure datablock creation and update by validating access to the target dataset and using its authoritative ownership metadata.

Bug Fixes:
- Authorize datablock creation and update against the target dataset and derive datablock ownership and access groups from that dataset.

Enhancements:
- Return a not-found error when the target dataset does not exist before creating a datablock.


omkar-ethz:
Thanks @HayenNico for cleaning up the PR!
I also implemented the fix for PATCH, as the same issue was present there. And added an e2e test each for the new behavior.

omkar-ethz (approved):

Junjiequan (approved):

src/datablocks/datablocks.controller.ts
```diff
@@ -59,6 +60,20 @@ export class DatablocksController {
     return instance;
   }
 
+  private generateDatasetInstanceForPermissions(
```
omkar-ethz, line 63: Seems to duplicate method of the same name from DatasetsController. To be refactored: https://github.com/SciCatProject/backend/issues/2901

PR #2920, opened 2026-08-31
fix: link proposals to ds by ownergroup

minottic:
## Description
This feature is ported from the old backend from [here](https://github.com/SciCatProject/backend-v3/blob/master/common/models/raw-dataset.js#L40)

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Automatically associate raw datasets with matching proposals by owner group while safely backfilling existing records and maintaining proposal dataset counts.

New Features:
- Automatically link raw datasets to proposals sharing their owner group when proposal IDs are missing, both during creation and subsequent updates.
- Add a migration to backfill proposal links for existing raw datasets and update proposal dataset counts.

Bug Fixes:
- Prevent automatic linking from overwriting explicitly supplied proposal IDs or concurrent updates.
- Limit owner-group-based links to 100 proposals to avoid unbounded associations.

Tests:
- Add unit and integration coverage for proposal linking on creation, patch backfills, capped matches, dataset types, existing links, and concurrent updates.

### DIFF
--- a/migrations/20260831150000-backfill-dataset-proposalids-from-ownergroup.js
+++ b/migrations/20260831150000-backfill-dataset-proposalids-from-ownergroup.js
@@ -6,80 +6,106 @@
  * more than one proposal.
  *
  * A broad/shared ownerGroup could otherwise match an unreasonable number of
- * unrelated proposals, so groups matching more than
- * MAX_PROPOSALS_PER_OWNER_GROUP proposals are skipped and logged instead of
- * writing a huge proposalIds array.
+ * unrelated proposals, so a dataset whose ownerGroup matches more than
+ * MAX_PROPOSALS_PER_OWNER_GROUP proposals only gets linked to the first
+ * MAX_PROPOSALS_PER_OWNER_GROUP of them, and the truncation is logged,
+ * instead of writing an unbounded proposalIds array.
  */
 
 const MAX_PROPOSALS_PER_OWNER_GROUP = 100;
+const CHUNK_SIZE = 1000;
+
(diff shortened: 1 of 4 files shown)


HayenNico (changes requested):
The PR is technically sound and the `MAX_PROPOSALS_PER_OWNER_GROUP` guard should catch most common unwanted behaviors.

I still think this introduces an assumption that a raw dataset is _supposed_ to have a proposal, which will not necessarily generalize to the business logic of all SciCat instances. Legacy raw datasets without proposals could be backfilled by a one-time script, service-level mutation of new create/update requests to enforce this added assumption is too restrictive in my opinion.

For now, I'd propose to add a configuration flag that defaults to opt out of this backfill feature. The `MAX_PROPOSALS_PER_OWNER_GROUP` should probably be configurable too.

migrations/20260831150000-backfill-dataset-proposalids-from-ownergroup.js
nitrosx, line 1: This code is sound, but it applies only to PSI use case

nitrosx (commented):
This PR does not apply to the general project, although the implementation is spotless.
It implements the specific use case for PSI.
The link between datasets and proposal through ownergroup cannot be generalized to every instances deployed.

I do not have a good grasp of the data curation workflow in PSI, so take my suggestion as a grain of salt. I suggest to transfer the introduction of such constraints/rules during ingestion or post ingestion.

I will be happy to discuss further and help finding a solution, but I firmly believe that this PR should not be merged.

PR #2925, opened 2026-09-02
feat: remove types hardcoding from policy and add v4

minottic:
## Description
Introduces policy V4 endpoints which accept a more generic payload that removes job types hardcoding. The v3 endpoints are fully backward compatible.

## Motivation
The v3 policy endpoints accept/return archiveEmail, retrieveEmail, but now that the jobs are configurable there is no concept of archive/retrieve anymore in the code base. To maintain backward compatibility the v3 endpoints keep these terminology.

The policy schema is migrated to a job agnostic schema with

```
manager
jobPolicies: {jobType: {emaiEnable: bool, emailList: string[], anyOtherField}
```

This allows policies to hold job specific rules, all archive specific steps are migrated to an archive policy in the migration (run with updateMany for atomicity)

The v4 endpoints now accept this schema

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Introduce job-agnostic policy storage and v4 endpoints while preserving backward compatibility for v3 consumers.

New Features:
- Add version 4 policy endpoints with generic, job-type keyed policy settings and CRUD support.
- Support field projection and nested filter limits for policy retrieval.

Bug Fixes:
- Preserve sibling nested policy settings during partial and bulk updates.
- Default unspecified policy sort directions to ascending and correctly translate legacy policy filters.

Enhancements:
- Restructure policy storage around configurable job policies while retaining the existing v3 API contract through legacy field mapping.
- Centralize policy access filtering and migrate existing notification fields into job-specific policy paths.

Tests:
- Add coverage for v3 backward compatibility, legacy-to-v4 mapping, filtering, ordering, partial updates, v4 CRUD behavior, and access control.

### DIFF
--- a/src/policies/dto/update-policy-v4.dto.ts
+++ b/src/policies/dto/update-policy-v4.dto.ts
@@ -1,0 +1,4 @@
+import { PartialType } from "@nestjs/swagger";
+import { CreatePolicyV4Dto } from "./create-policy-v4.dto";
+
+export class UpdatePolicyV4Dto extends PartialType(CreatePolicyV4Dto) {}
--- a/src/policies/pipes/legacy-notification.pipe.ts
+++ b/src/policies/pipes/legacy-notification.pipe.ts
@@ -1,0 +1,20 @@
+import { Injectable, PipeTransform } from "@nestjs/common";
+import { createDeepSetter } from "src/common/utils/deep-mapper.util";
+import { PartialUpdatePolicyDto } from "../dto/update-policy.dto";
+import { policyV3toV4FieldMap } from "../dto/policy.obsolete.dto";
+import { Policy } from "../schemas/policy.schema";
+
+const dtoV3toV4 = createDeepSetter<PartialUpdatePolicyDto, Partial<Policy>>(
+  policyV3toV4FieldMap,
+);
+
+@Injectable()
+export class V3ToV4MigrationPipe<S, T> implements PipeTransform {
+  constructor(private readonly mapper: (source: S) => T) {}
+
+  transform(value: S): T {
+    return this.mapper(value);
+  }
+}
+
+export const LEGACY_NOTIFICATION_PIPE = new V3ToV4MigrationPipe(dtoV3toV4);
--- a/src/policies/utils/policy-access-filter.util.ts
+++ b/src/policies/utils/policy-access-filter.util.ts
@@ -2,17 +2,16 @@ import { Request } from "express";
 import { JWTUser } from "src/auth/interfaces/jwt-user.interface";
 import { Action } from "src/casl/action.enum";
 import { CaslAbilityFactory } from "src/casl/casl-ability.factory";
-import { IFilters } from "src/common/interfaces/common.interface";
-import { IPolicyFilter } from "../interfaces/policy-filters.interface";
-import { Policy, PolicyDocument } from "../schemas/policy.schema";
+import { IPolicyFilterV4 } from "../interfaces/policy-filters.interface";
+import { Policy } from "../schemas/policy.schema";
 
 // Users who can list their own policies but not all of them are restricted
 // to policies for a group they belong to, or ones already public.
 export function restrictToOwnPolicies(
   caslAbilityFactory: CaslAbilityFactory,
   request: Request,
-  mergedFilters: IFilters<PolicyDocument, IPolicyFilter>,
-): IFilters<PolicyDocument, IPolicyFilter> {
+  mergedFilters: IPolicyFilterV4,
+): IPolicyFilterV4 {
   const user = request.user as JWTUser;
   if (!user) return mergedFilters;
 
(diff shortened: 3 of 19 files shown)


src/policies/utils/policy-access-filter.util.ts
```diff
@@ -0,0 +1,33 @@
...
+  caslAbilityFactory: CaslAbilityFactory,
+  request: Request,
+  mergedFilters: IPolicyFilterV4,
+): IPolicyFilterV4 {
+  const user = request.user as JWTUser;
+  if (!user) return mergedFilters;
+
+  const ability = caslAbilityFactory.policyAccess(user);
+  const canViewAll = ability.can(Action.ListAll, Policy);
+  const canViewTheirOwn = ability.can(Action.ListOwn, Policy);
```
HayenNico, line 20: These actions (ListAll and ListOwn) aren't defined on casl policyAccess (and are planned to be removed entirely). The correct actions would be `Action.AccessAny` for unrestricted visibility and `Action.Read` otherwise.

There is no group under the current model that has ownership-constrained permissions here; `ADMIN_GROUPS` and `POLICY_GROUPS` may read all policies, all other users have no access anyway. So the canViewTheirOwn condition can be omitted

src/policies/dto/update-policy-v4.dto.ts
```diff
@@ -0,0 +1,4 @@
+import { PartialType } from "@nestjs/swagger";
+import { CreatePolicyV4Dto } from "./create-policy-v4.dto";
+
+export class UpdatePolicyV4Dto extends PartialType(CreatePolicyV4Dto) {}
```
HayenNico, line 4: For consistency this might be renamed to `PartialUpdatePoliyV4Dto`. The pure Update DTOs are normally used for PUT endpoints and are superclasses of the Create DTOs. If there are no PUT endpoints it of course doesn't make sense to define that separately.

src/policies/pipes/legacy-notification.pipe.ts
```diff
@@ -0,0 +1,20 @@
...
+@Injectable()
+export class V3ToV4MigrationPipe<S, T> implements PipeTransform {
+  constructor(private readonly mapper: (source: S) => T) {}
+
+  transform(value: S): T {
+    return this.mapper(value);
+  }
+}
+
+export const LEGACY_NOTIFICATION_PIPE = new V3ToV4MigrationPipe(dtoV3toV4);
```
HayenNico, line 20: Could you explain the naming? I don't see a connection to notifications here

HayenNico (changes requested):
A couple small changes needed related to authorization, looking good otherwise.

For my understanding: Is there any other intended use of this collection outside of archive/retrieve jobs in the current codebase?

src/policies/utils/policy-access-filter.util.ts
minottic: I see, thanks. This part of the code, I have not changed I have just changed its location , it used to live [here](https://github.com/SciCatProject/backend/pull/2925/changes#diff-4e24a3259b327c091de6714f6528a6b2ddf7596b0da37b2b25ab4363691147baL43-L67).

I am not super familiar with the new CASL, what do you recommend to use there?

src/policies/pipes/legacy-notification.pipe.ts
minottic: right, my bad, it was an old name that slipped through

src/policies/utils/policy-access-filter.util.ts
HayenNico: The new casl ability lives in `src/casl/abilities/policies.ability.ts` - it appears I've overlooked the old code during the controller refactor (though I think these actions were never defined on policy casl).

Realistically you can skip this check entirely, since there is no instance-level permission checks - if a user passes endpoint authorization, `canViewAll` should always be true.

If you want to keep the permission check here anyway, you'd need the following changes:
Here:
```ts
const canViewAll = ability.can(Action.AccessAny, Policy);
const canView = ability.can(Action.Read, Policy);

if (!canViewAll && canView) {
    ...
}
```

And in `src/casl/abilities/policies.ability.ts` L.45 f.:
```ts
      /**
       * User belonging to ADMIN_GROUPS or POLICY_GROUPS
       */
      can(Action.AccessAny, Policy);

      can(Action.Create, Policy);
      can(Action.Read, Policy);
      can(Action.Update, Policy);
```

minottic: I see, thanks. Would it be fine if I do later in a dedicated PR? This PR only adds the new contorller without touching on existing casl logic, so I'd prefer separating the change

HayenNico: There's no immediate auth problem with the code as is, but I'd prefer if you switched both actions to `Action.Read`. We can consider the casl change later, or put it on the docket for the auth rework project.

Improperly used casl checks like this (either undefined action or wrong subject class) that always evaluate to false are a landmine that can lead to permissions bypasses like in the recent security advisory when combined with a second bug.

minottic: ok, my point is a bit that this requires a dedicated PR imo rather than merging into this one that did not touch that part of the code. But I can create one new PR that we merge before this (or maybe if you could do, I think you would be much faster). Then I can rebase this one on top and the moved code block will mirror the change

PR #2929, opened 2026-09-02
fix: fix occ concurrency comparison

minottic:
## Description
OCC comparisons had 2 problems:

1. 2 clients submitting requests at the same exact time, both of them would succeed since they would override each other since they had the same updatedAt as long as the mongo update takes sub milliseconds
2. when a client sends an http date in the header, this gets truncated to seconds resulting in a false 412. This ceils to 99 ms more

It includes a refactor to avoid duplication

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Fix optimistic concurrency comparisons and centralize conditional update handling across resource services.

Bug Fixes:
- Prevent optimistic-concurrency updates from allowing simultaneous requests with the same timestamp to overwrite each other.
- Treat second-precision If-Unmodified-Since headers as covering the full second, avoiding false 412 responses while preserving explicit millisecond precision.

Enhancements:
- Centralize OCC update handling across supported services, including timestamp management and consistent not-found/precondition errors.
- Read and normalize If-Unmodified-Since values consistently across update controllers.

Tests:
- Add coverage for timestamp race prevention, header precision handling, shared OCC update behavior, and related controller updates.


src/datasets/utils/occ-util.ts
HayenNico, line 1: This file (and the .spec file) should be placed under common, they are not tied to datasets specifically

HayenNico (approved):
The new logic looks good to me, I'd appreciate if someone else had a second look over the main changes in `occ-util.ts` before merging though.

All controller/service changes are solid and significantly reduce code duplication :+1:

omkar-ethz (changes requested):
I'm not sure I understand the need for this change. Is there some use-case where this is becoming a problem?

> 2 clients submitting requests at the same exact time, both of them would succeed since they would override each other since they had the same updatedAt as long as the mongo update takes sub milliseconds

The original feature supports millisecond precision for this reason. The intended behavior is that client GET's the dataset, reads the updatedAt e.g. 10:00:00.437, and sends it in the if-unmodified-since header.

If the client decides to truncate it down and send only a second precision value e.g. 10:00:00, both clients will fail with 412 - which i believe is the expected behavior, as 10:00:00 == 10:00:00.000.

Alternatively, the clients can truncate it upward to 10:00:01, and let's say the clients are making this request at 10:00:05: as mongodb will serialize the writes, exactly one client will succeed, as the new updatedAt will become 10:00:05. (the second client will fail with 412)

The only case when the issue you describe can occur is:
1. the dataset's updatedAt is 10:00:00.437
2. two client's have a copy of the dataset and cannot use millisecond precision for some reason
3. within the same second, say at 10:00:00.542, both of them send a patch request with if-unmodified-since containing 10:00:01
4. In this case, both clients will succeed - but again, this is in line with what the original feature promises. The solution for fast clients is to use millisecond precision

In conclusion, i think maybe the problem this PR is trying to address lies on the client side rather than the server.

No comments on the refactor though - looks like Nicolas reviewed it already! I only reviewed the logic change

minottic:
1. imagine 2 clients sending a patch with the same `unmodified-since`. Without this change the updatedAt (new Date) could be = to unmodifiedSince (as long as the whole patch is faster than millisecond precision) and so they would override each other. With this change the client succeeding first will set updatedAt = unmodifiedSince + 1 and the second one will fail since unmodifiedSince < updatedAt (=unmodifiedSince +1)
2. this is a check that seemed needed by the specs for [unmodifiedSince](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-Unmodified-Since) that sets precision up to seconds. The second change here assumes that every client passing seconds precision (as per the doc) wants to allow mods up to +99 milliseconds. But this one is some sort of assumption, so I am happy to discard it

minottic:
better reading the docs, I think point 2 is actually too strong of an assumption (probably even wrong), thanks for challenging that, I will remove it

omkar-ethz:
> imagine 2 clients sending a patch with the same unmodified-since.

There is no problem if two clients send the same unmodified-since (https://github.com/SciCatProject/backend/blob/master/test/DatasetV4.js#L2131 and https://github.com/SciCatProject/backend/blob/master/test/DatasetV4.js#L2136 - the e2e test sets the same value and fires two requests simultaneously) 

(Just to make sure we're on the same page, the OCC feature is the client saying: "I have this version of the dataset, which was last updated at 10:00:00.123. Now it's 10:00:05. Please update it if nobody else has updated it since 10:00:00.123. Otherwise let me know with a 412, so i can GET it again, and retry the patch")

So even if two clients act at exactly 10:00:05.123, they pass the same unmodified-since i.e. 10:00:00.123 (as in the e2e test). As soon as one of them succeeds, MongoDB sets updatedAt to new Date(), say 10:00:05.456, and the second client gets a 412 (10:00:05.456 > 10:00:00.123).

>  Without this change the updatedAt (new Date) could be = to unmodifiedSince

If the client can use millisecond precision, it is impossible that new Date() == unmodified-since. unmodified-since is supposed to be the dataset's last updatedAt, which by definition is in the past. Assuming a HTTP request takes at least 1ms, new Date will always be > unmodifiedSince

If the client cannot use millisecond precision i.e. we want to be compatible with the spec, there are two ways to derive the if-unmodified-since header from updatedAt = 10:00:00.123:

Truncating down (if-unmodified-since: 10:00:00): Here, in the original impl, both clients get a 412. I think that's correct behaviour, as the dataset was modified at 10:00:00.123.

Rounding up (if-unmodified-since: 10:00:01): here both can succeed, if they both PATCH *before* 10:00:01. I think this is also acceptable. Because, if the clients don't have millisecond precision, they should just wait until it's at least 10:00:01 to send a PATCH request with unmodified since value of 10:00:01.
Note that how unlikely it is though that a dataset that was last modified at 10:00:00.123 is being updated before it's 10:00:01. If both clients PATCH after 10:00:01, again there is no problem: one of them will win, set the updatedAt in mongo to e.g. 10:00:01.123, and the second client will get 412. So in order to be compatible with the spec i.e. only using second precision, the clients simply need to avoid PATCHing the dataset in the same second that it was last updatedAt.

If waiting a whole second is unacceptable, millisecond precision works without problems as shown in the e2e test.

> The second change here assumes that every client passing seconds precision (as per the doc) wants to allow mods up to +999 milliseconds. But this one is some sort of assumption, so I am happy to discard it

I like the idea, but it seems we're changing the meaning of the feature rather than fixing anything. Basically we are saying 10:00:00 == 10:00:00.999 rather than the more intuitive (imo)  10:00:00 == 10:00:00.000

I think the fix you describe in 1 is only necessary if we implement the assumption in 2. In the end, I don't see the need to add complexity to the logic when the original implementation handles all cases.

src/datasets/utils/occ-util.ts
```diff
@@ -13,3 +19,56 @@ export function withOCCFilter<T extends QueryableClass>(
...
+): Promise<T> {
+  const queryFilter = withOCCFilter(filter, unmodifiedSince);
+  const updated = await model
+    .findOneAndUpdate(
+      queryFilter,
+      { ...update, updatedAt: nextOCCTimestamp(unmodifiedSince) },
+      // timestamps: false so Mongoose's own updatedAt auto-management
+      // doesn't clobber nextOCCTimestamp's value and reopen the
+      // same-millisecond OCC race.
+      { new: true, timestamps: false, ...queryOptions },
```
omkar-ethz, line 62: i think this part is also dangerous, as we now let the client override document's `updatedAt` timestamp. If one passes `if-unmodified-since: 2030-01-01`, with this change, the update would succeed _and_ mark the document as updated in 2030.
In the original impl, the update would succeed, but updatedAt timestamp would still be now().

minottic:
ah I see, makes sense. Closing it

PR #2955, opened 2026-09-10
fix(proposal): don't overwrite `MeasurementPeriodList` on updates

fpotier:
## Description

Updating a proposal with no `MeasurementPeriodList` should not reset the existing one.

## Motivation

see https://github.com/SciCatProject/backend/pull/2904/changes#r3977231988

## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Preserve measurement periods during partial proposal updates instead of replacing omitted lists with empty values.

Bug Fixes:
- Preserve an existing proposal's MeasurementPeriodList when updates omit that field.

Tests:
- Add interceptor and end-to-end coverage for UTC conversion and retaining measurement periods during partial proposal updates.

### DIFF
--- a/src/common/interceptors/multi-utc-time.interceptor.ts
+++ b/src/common/interceptors/multi-utc-time.interceptor.ts
@@ -23,7 +23,12 @@ export class MultiUTCTimeInterceptor<T, U> implements NestInterceptor {
   ): Observable<unknown> | Promise<Observable<unknown>> {
     const req = context.switchToHttp().getRequest();
     const instances = req.body[this.subGroup] as U[];
-    req.body[this.subGroup] = updateAllTimesToUTC<U>(this.dateKeys, instances);
+    if (instances) {
+      req.body[this.subGroup] = updateAllTimesToUTC<U>(
+        this.dateKeys,
+        instances,
+      );
+    }
     return next.handle();
   }
 }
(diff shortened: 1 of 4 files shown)


src/common/interceptors/multi-utc-time.interceptor.ts
```diff
@@ -23,7 +23,12 @@ export class MultiUTCTimeInterceptor<T, U> implements NestInterceptor {
   ): Observable<unknown> | Promise<Observable<unknown>> {
     const req = context.switchToHttp().getRequest();
     const instances = req.body[this.subGroup] as U[];
-    req.body[this.subGroup] = updateAllTimesToUTC<U>(this.dateKeys, instances);
+    if (instances) {
+      req.body[this.subGroup] = updateAllTimesToUTC<U>(
+        this.dateKeys,
+        instances,
+      );
+    }
```
fpotier, line 31: Maybe the this should be handled in `updateAllTimesToUTC`, no-op instead of returning an empty array

minottic (approved):

Junjiequan (approved):
The root cause/ side effect seems to be from here: https://github.com/SciCatProject/backend/blob/master/src/main.ts#L82. 
The global validation pipe inherits `MeasurementPeriodList=[]` from updateProposalDto when it's missing.

If the `MeasurementPeriodList=[]` is intended, it should be sitting in the createDatasetDto I suppose. But this fix looks good to me as well

src/common/interceptors/multi-utc-time.interceptor.ts
Junjiequan: This seems like the right place for the guard, rather than in `updateAllTimesToUTC`. Otherwise we'd need two guards, one here and one there, since `req.body[this.subGroup]` can be null or undefined if no value is return from updateAlTimesToUTC.

PR #2967, opened 2026-09-15
chore(attachmentDTO): update description for targetId api property

Junjiequan:
## Fixes

* Bug fixed (#X)

## Changes:

* changes made

## Tests included

- [ ] Passing? 

## Documentation

## Summary by Sourcery

Enhancements:
- Clarify the API descriptions for attachment relationship target IDs to state that each value identifies the entity linked to the attachment.


HayenNico (approved):
Side note following today's meeting discussion: One attachment can still be linked to multiple resources at the same time, since `attachment.relationships` is an array of `AttachmentRelationshipClass`. Making the relation 1:1 would require a schema change after all.

Junjiequan:
> Side note following today's meeting discussion: One attachment can still be linked to multiple resources at the same time, since `attachment.relationships` is an array of `AttachmentRelationshipClass`. Making the relation 1:1 would require a schema change after all.

Yes, you are right. I rushed making concolusion based on the targetId, probably would still not noticed untill you commented,  thanks for pointing that out.

Now, I rememberd that, we made relationships as array is to have an attachment page, where user can find attachments they have access to without duplication. Looking back now, I'm not sure it's a feature that is worth to implement

HayenNico:
Depends on what types of attachments people produce - if it were only thumbnail images I'd tend to agree, but I believe some facilities use attachments for other file types as well (PDF files at DESY are one recent example). For those, an overview might be worth the effort?

Junjiequan:
> Depends on what types of attachments people produce - if it were only thumbnail images I'd tend to agree, but I believe some facilities use attachments for other file types as well (PDF files at DESY are one recent example). For those, an overview might be worth the effort?

I'm not really sure. I feel that user feedbacks are important in this case. Would user be interested spending time to read PDFs first and then search for datasets associated with them, or vice versa.

fpotier (approved):

